PR #89, opened 2026-04-30
fix(services): address CodeRabbit findings from PR #86

### LINKED ISSUE #87: Address pre-existing CodeRabbit findings from PR #86
20 findings from CodeRabbit's review of #86 that pre-existed on `main` and were deferred so the file-split refactor PR could stay scope-clean ("no behavior changes"). Verified pre-existing by spot-checking the original files before the split.

Each item links to the location in the post-split layout. Tackle in any order — most are independent.

## Logic / behavior bugs

- [ ] **`crates/services/src/base/world_entry/enable_entities.rs:48-68`** — `take()` drains `pending_player_entity_id`/`pending_world_entry`/`pending_player_load_data` before `send_to` succeeds. If the send fails, login state is permanently lost. Defer the `take()` until after the create-player commit succeeds.
- [ ] **`crates/services/src/base/world_entry/enable_entities.rs:106-137`** — `char_list_sent` is set to `true` before `query_character_list` / `get_account_entity_id` / `socket.send_to` complete. Failure permanently blocks retry. Only flip the flag after the full path succeeds (re-acquire the lock and set after `send_to` returns Ok).
- [ ] **`crates/services/src/cell/cell_methods/inventory/bandolier.rs:27-39`** — `entity.bandolier_ammo_dirty.drain()` runs before `tx.send(BandolierAmmoUpdate)`. If the send fails the dirty marker is lost and the slot stops syncing. Apply the same pattern to the other bandolier sites in the same file.
- [ ] **`crates/services/src/cell/cell_methods/inventory/item_ops.rs:38-40`** — three `let _ = tx.send(...).await` calls silently drop failures (`RemoveInventoryItem` and two others). Match on the `Result`, log on `Err` (include `entity_id`/`player_id`/`item_id`/`quantity`), and propagate.
- [ ] **`crates/services/src/cell/abilities/loot_drop.rs:54-59`** — `(entry.max_quantity - entry.min_quantity)` wraps when `min > max` (bad DB row). Guard with `entry.min_quantity <= entry.max_quantity`; otherwise log and use `min` (or swap).
- [ ] **`crates/services/src/cell/abilities/rng.rs:17`** — `(h as f64) / (u32::MAX as f64)` returns `1.0` when `h == u32::MAX`. Divide by `u32::MAX as f64 + 1.0` (or cast `h` into a wider int and add 1) so the upper bound is exclusive.
- [ ] **`crates/services/src/cell/abilities/use_ability.rs:283-289`** — `calculate_qr` is invoked with literal `true` for the ranged flag, treating every ability as ranged. Pass the ability's actual `is_ranged`/`ranged` flag from the AbilityDef.
- [ ] **`crates/services/src/cell/abilities/use_ability.rs:292-309`** — when `ability_def` is `Some(...)` but no positive `HealthDamage` is found (`h_dmg == 0`), the code falls back to `15` HP, turning focus/heal abilities into physical hits. Set `health_base_damage = 0` in that branch and only keep the `15` fallback for the `ability_def == None` case.
- [ ] **`crates/services/src/cell/service/npc_ai.rs:120-178`** — `has_los` is computed but the in-range branch attacks regardless. When `dist_to_target <= NPC_ATTACK_RANGE && !has_los`, skip `handle_use_ability` (and optionally clear `nav_path` / log) instead of firing.
- [ ] **`crates/services/src/cell/space_manager/entities.rs:103-114`** — disconnect iterates the disconnecting entity's `witnesses` (the entities IT sees) and sends `LeftAoI` to those — wrong direction. Iterate `space.entities` and for each `other` whose `witnesses.contains(disconnecting_id)`, send `LeftAoI { witness_id: other.id, entity_id: disconnecting_id }`.
- [ ] **`crates/services/src/cell/space_manager/queries.rs:107-118`** — `find_entity_by_tag` (and `find_entities_by_template`) searches every Space matching `world_name`. In instanced worlds this leaks entities across instances. Take a `space_id` (or `source_entity_id` to look up the space) and restrict iteration to that one space.
- [ ] **`crates/services/src/cell/space_manager/queries.rs:147-158`** — `get_witnesses_of` scans every player's `witnesses` set checking for the target. Faster + correct: fetch the target entity, read its `witnesses`, and filter `space.players` to those whose `EntityId` is in that set.

## Robustness / style

- [ ] **`crates/services/src/base/world_entry/space_registry.rs:18`** — `.lock().unwrap()` panics on a poisoned mutex. Either recover with `.unwrap_or_else(|p| p.into_inner())` or document that fail-fast is intentional.
- [ ] **`crates/services/src/cell/space_manager/xml.rs:43-101`** — XML parsing silently coerces errors via `attributes().flatten()`, `from_utf8(...).unwrap_or("")`, and `parse().unwrap_or(0)`, hiding malformed configs. Propagate parse errors instead. Apply to both attribute blocks (~43-53 and ~96-101).
- [ ] **`crates/services/src/cell/spawner/respawners.rs:38-49`** — manual `Row::get` for each column defers type/column-name mismatches to runtime. Derive `sqlx::FromRow` on `RespawnerDef` and use `sqlx::query_as`.
- [ ] **`crates/services/src/mercury/protocol/character.rs:54-67`** — `build_char_list` and `build_on_character_list` repeat the same `CharacterInfo` serialization sequence. Extract `fn write_character_info(buf: &mut Vec<u8>, ch: &CharacterInfo)` and call from both.

## Test design

- [ ] **`crates/services/src/base/world_entry/tests.rs:2-16`** — `space_id_mapping_known_worlds` asserts hardcoded u32 literals. Call the actual resolver (`SpaceRegistry::space_id_for_name` / `WorldEntry::resolve_space_id`) and assert distinct ids with `(id >> 16) == 1` instead.
- [ ] **`crates/services/src/cell/cell_methods/inventory/tests.rs:79-83` and `:135-137`** — uses `tokio::time::sleep(5ms)` to outrun a 1ms ability cooldown. Reset the cooldown on `mgr` between shots so the test isn't wall-clock-sensitive.

## Performance

- [ ] **`crates/services/src/cell/service/ticks.rs:196`** — `npc.nav_path.remove(0)` is O(n). Change `nav_path: Vec<...>` to `VecDeque<...>` on the NPC struct (in the `entity` crate) and call `pop_front()`. Update construction sites to use `VecDeque::from(vec)` and any `nav_path[i]` indexing to use `front()`/`back()`/`iter()`.

## Lifecycle

- [ ] **`crates/services/src/cell/service/startup.rs:184-191`** — the spawned `run_cell_loop` is detached, so `CellService::stop()` can't signal it or await its exit. Capture the `JoinHandle` on the service (e.g., `cell_loop_handle: Option<JoinHandle<()>>`) and add a `CancellationToken` (or shutdown channel) that `run_cell_loop` honors. Have `stop()` cancel and await the handle.

---

Cadacious:
Closes #87.

## Summary

Addresses the 20 pre-existing CodeRabbit findings deferred from #86 so that PR could stay scope-clean.

### Logic / behavior bugs
- `enable_entities.rs` — peek pending state, only `take()` / flip `char_list_sent` after `send_to` succeeds so a transient send failure leaves login state intact for retry.
- `bandolier.rs` — three sites: dirty markers cleared only after the persist message is accepted by the channel.
- `item_ops.rs` — `Remove`/`List`/`MoveInventoryItem` send failures now log the player/item context instead of being dropped.
- `loot_drop.rs` — guard against malformed `min > max` rows that previously produced wraparound quantities.
- `rng.rs` — divide by `u32::MAX as f64 + 1.0` so the result is strictly half-open `[0, 1)`.
- `use_ability.rs` — pass the ability's actual `is_ranged` flag into `calculate_qr`, and only fall back to 15-HP physical damage when `ability_def` is `None` (heals/focus drains no longer get reinterpreted as melee hits).
- `npc_ai.rs` — NPCs hold fire when in range but without LOS instead of shooting through walls.
- `space_manager/entities.rs` — disconnect now iterates `space.entities` and notifies entities whose `witnesses` set contains the disconnecting id (correct direction).
- `queries.rs` — `find_entity_by_tag` / `find_entities_by_template` now take `source_entity_id` and restrict iteration to that one space, preventing instanced worlds from leaking entity resolution across instances. All callers in `executor.rs` and the spawner test updated.

### Robustness / style
- `space_registry.rs` — recover from a poisoned mutex via `unwrap_or_else(|p| p.into_inner())`.
- `xml.rs` — propagate `quick-xml` / UTF-8 / `parse()` errors instead of swallowing them with `flatten()` / `unwrap_or(0)`.
- `respawners.rs` — derive `sqlx::FromRow` on an intermediate `RespawnerRow` and use `query_as` so column/type mismatches surface earlier.
- `mercury/protocol/character.rs` — extracted `write_character_info` helper shared by `build_char_list` and `build_on_character_list`.

### Test design
- `world_entry/tests.rs` — calls `resolve_space_id_fallback` and asserts the structural invariant (`id >> 16 == 1`) instead of hardcoding three u32 literals.
- `inventory/tests.rs` — replaced two `tokio::time::sleep(5ms)` calls with `clear_all_cooldowns()` between shots so the test isn't wall-clock-sensitive.

### Performance
- `cell_entity.rs` — `nav_path: Vec<Vector3>` → `VecDeque<Vector3>`. `ticks.rs` and `npc_ai.rs` updated to `pop_front()` / `front()` / `back()` / `get(i)`.

### Lifecycle
- `CellService` — added `cell_loop_handle: Option<JoinHandle<()>>` and `shutdown_signal: Option<Arc<Notify>>`. `start()` spawns and stores the handle; `run_cell_loop` adds a `shutdown.notified()` arm to its `select!`; `stop()` notifies and awaits the handle so the loop is wound down deterministically instead of relying on the channel half being dropped.

### Deviation from issue prompt

The `get_witnesses_of` finding suggested reading `target.witnesses` and filtering players. In this codebase `entity.witnesses` is "entities I see" (only populated for players via `aoi.rs:170`), so the suggested rewrite would return empty for NPC targets. Kept the player-scan logic but documented the semantic and added a per-space scope so the scan is bounded.

## Test plan

- [ ] `cargo check -p cimmeria-services`
- [ ] `cargo test -p cimmeria-services` — focus on `slot_swap_preserves_per_slot_ammo`, `find_entity_by_tag_works`, `space_id_mapping_known_worlds`
- [ ] Manual smoke: connect a client, log in, char-select, enter Castle_CellBlock, fire a few weapons, swap bandolier slots — confirm no behavior regression vs. main
- [ ] Verify `CellService::stop()` joins cleanly (no orphan task warnings)

## Release Notes

* **Bug Fixes**
  * Fixed random number edge-case affecting ability calculations
  * Prevented invalid loot quantity ranges from underflowing
  * NPCs now consider line-of-sight when deciding to re-path or engage

* **Improvements**
  * More robust login/create-player flow that preserves state on transient send failures
  * Inventory and mission/channel sends now surface errors instead of being ignored
  * Persistence of bandolier ammo updates now confirmed before clearing
  * Message loop and service shutdown are now coordinated for clean teardown
  * Lookups scoped to an entity's space to avoid cross-space leakage
  * Stricter XML/config validation and safer concurrent registry access

### DIFF
--- a/crates/entity/src/cell_entity.rs
+++ b/crates/entity/src/cell_entity.rs
@@ -9,7 +9,7 @@
 //! This corresponds to the C++ `CellEntity` / `Entity` classes in
 //! `src/server/CellApp/`.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
 
 use serde::{Deserialize, Serialize};
 
@@ -192,8 +192,10 @@ pub struct CellEntity {
     /// Ticks until next AI action (count-down from ai tick interval).
     pub ai_cooldown_ticks: u32,
     /// Navmesh path waypoints the NPC is currently following.
-    /// Empty = not moving. Each tick consumes the next waypoint.
-    pub nav_path: Vec<Vector3>,
+    /// Empty = not moving. Each tick pops the next waypoint off the front.
+    /// Stored as `VecDeque` so per-tick `pop_front` is O(1) instead of the
+    /// O(n) shift `Vec::remove(0)` would do.
+    pub nav_path: VecDeque<Vector3>,
     /// Movement speed in world units per tick.
     pub move_speed: f32,
 
@@ -315,7 +317,7 @@ impl CellEntity {
             threat_list: HashMap::new(),
             spawn_position: None,
             ai_cooldown_ticks: 0,
-            nav_path: Vec::new(),
+            nav_path: VecDeque::new(),
             move_speed: 0.6, // ~0.6 world units per 100ms tick = 6 units/sec
             saved_missions_loaded: false,
             loot_table_id: None,
--- a/crates/services/src/base/world_entry/enable_entities.rs
+++ b/crates/services/src/base/world_entry/enable_entities.rs
@@ -44,24 +44,16 @@ pub(crate) async fn handle_enable_entities(
     _cell_tx: &Option<mpsc::Sender<BaseToCellMsg>>,
     _entity_to_addr: &Arc<Mutex<HashMap<u32, SocketAddr>>>,
 ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    // Check if we have a pending world entry (create-player step).
+    // Peek at pending world entry without consuming it: we only commit (take)
+    // after the create-player send_to succeeds, so a transient send failure
+    // leaves login state intact for the next ENABLE_ENTITIES retry.
     let pending = {
-        let mut clients = connected.lock().map_err(|_| "connected lock poisoned")?;
-        if let Some(c) = clients.get_mut(&addr) {
-            match (c.pending_player_entity_id.take(), c.pending_world_entry.take()) {
+        let clients = connected.lock().map_err(|_| "connected lock poisoned")?;
+        if let Some(c) = clients.get(&addr) {
+            match (c.pending_player_entity_id, c.pending_world_entry.clone()) {
                 (Some(eid), Some(entry)) => Some((eid, entry)),
                 _ => None,
             }
-        } else {
-            None
-        }
-    };
-
-    // Also retrieve the pending player load data
-    let pending_load = {
-        let mut clients = connected.lock().map_err(|_| "connected lock poisoned")?;
-        if let Some(c) = clients.get_mut(&addr) {
-            c.pending_player_load_data.take()
         } else {
             None
         }
@@ -86,15 +78,18 @@ pub(crate) async fn handle_enable_entities(
         let pkt = build_create_player(&key, seq, &acks, &entry_info);
         socket.send_to(&pkt, addr).await?;
 
-        // Store world entry + player data for the enter-world step (triggered by mapLoaded)
+        // Send succeeded — now commit: take pending state and stage map_loaded data.
         {
-        let mut clients = connected.lock().map_err(|_| "connected lock poisoned")?;
-        if let Some(c) = clients.get_mut(&addr) {
-            c.pending_map_loaded = Some(entry_info);
-            c.pending_player_load_data = pending_load;
-            c.pending_client_ready = None;
+            let mut clients = connected.lock().map_err(|_| "connected lock poisoned")?;
+            if let Some(c) = clients.get_mut(&addr) {
+                c.pending_player_entity_id.take();
+                let entry = c.pending_world_entry.take().unwrap_or(entry_info);
+                let load = c.pending_player_load_data.take();
+                c.pending_map_loaded = Some(entry);
+                c.pending_player_load_data = load;
+                c.pending_client_ready = None;
+            }
         }
-    }
 
         tracing::info!(%addr, "Create player complete -- waiting for client mapLoaded");
         return Ok(());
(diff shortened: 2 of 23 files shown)


Cadacious:
@coderabbitai review

Issue #79, opened 2026-04-26
Test coverage gap: vendor/inventory/progression code from PR #77 is unverified


Cadacious:
PR #77 introduced and substantially modified several subsystems (vendor stack, inventory mutations, player load, progression, mail) and fixed 26+ bugs from the CodeRabbit review, but none of the modified code has test coverage. This issue tracks what needs tests and groups them by the type of test infrastructure required.

## Why this matters

The fixes in #77 include subtle concurrency and transaction guarantees that are easy to regress without tests:
- PostgreSQL advisory locks, FOR UPDATE row locks, multi-statement transactions with rollback
- rows_affected() == 1 invariants on UPDATE/INSERT pairs
- BIGINT casts to prevent INT overflow
- Bandolier state propagation between base and cell after equipped-item mutations

A passing build proves none of this; a regression today would silently corrupt player inventory or cash.

## Group A — DB integration tests (needs live PostgreSQL via sqlx::test or testcontainers)

The bulk of #77's logic is SQL transactions. Unit tests with mocked PgPool would test the wrapper, not the behavior under concurrent access — false confidence. These need a real database fixture.

- [ ] **vendor/buyback.rs::handle_buyback_vendor_items** — full-move clears flags, partial-move preserves price on remainder, transaction rolls back on insufficient cash / missing buyback row / slot exhaustion, no orphaned stack_size=0 rows after partial commits.
- [ ] **vendor/sell.rs::handle_sell_vendor_items** — bandolier_changed triggers sync_bandolier_after_inventory_change after selling equipped items, cash credit math, transaction rolls back on naquadah overflow, store update emitted.
- [ ] **vendor/purchase.rs::handle_purchase_vendor_items** — price lookup, free-slot reservation, transaction rolls back on insufficient naquadah, normalize_item_quantities edge cases (negative/zero quantities).
- [ ] **vendor/paid_repair.rs::handle_paid_repair_inventory_items** — BIGINT cast prevents overflow when naquadah * (100 - durability) exceeds i32::MAX, transaction rolls back on price mismatch, vendor without repair_item_list returns early with warn (not silent).
- [ ] **vendor/paid_recharge.rs::handle_paid_recharge_inventory_items** — same BIGINT cast on naquadah * (charges - inv.charges) / charges, vendor without recharge_item_list returns early with warn.
- [ ] **vendor/repair.rs::handle_repair_inventory_items** — free-repair path when vendor_template_id is None, durability clamps to 100.
[...]

PR #172, opened 2026-05-03
test(smoke): wire all three group D smokes (vendor + inventory-move + progression) into live-DB CI

Cadacious:
## Summary

Closes the live-DB cargo-test wiring for all three group D smokes from #79. Each smoke is a PL/pgSQL block in `tools/`, embedded via `include_str!`, executed via `sqlx::raw_sql`, and asserts via `RAISE EXCEPTION` (which sqlx surfaces as `Err`). All three run automatically in the live-DB CI job (#153) on every PR.

### 1. Vendor store (`tools/vendor_store_smoke.sql`, already in tree)

End-to-end sell → buyback → grant → purchase against the seeded vendor stack. Catches whole-stack regressions per-handler tests miss — e.g. if `handle_sell_vendor_items` and `handle_buyback_vendor_items` stop agreeing on the `flags` column's role as the buyback unit price, each handler stays internally green but the round-trip prices in the smoke don't match.

Sentinel id `-900001`.

### 2. Inventory move (`tools/inventory_move_smoke.sql`, new)

Seeds a player with a stack-of-10 at slot 0 and a single at slot 1, then chains:

- **Split** — UPDATE stack to 7, INSERT new row at slot 5 with stack_size=3. Asserts source/split row state, total row count, and stack-conservation (sum of stack_size for the type = 11).
- **Simple move** — UPDATE slot_id from 1 to 4. Asserts the unique-slot index doesn't reject and slot 1 vacates.
- **Three-step swap** — swap the stack at slot 0 with the single at slot 4 via park-at-sentinel-slot procedure. Catches a regression that drops Step C (move parked source into target's slot) — the script asserts the sentinel slot is empty at the end specifically for that bug shape.

Also pins:
- stack-conservation invariant after split + swap,
- unique-(container, slot) invariant via an explicit `GROUP BY HAVING COUNT(*) > 1` duplicate-row check.

Sentinel id `-900002`.

### 3. Progression (`tools/progression_smoke.sql`, new)

Seeds one account with two characters (A and B), both starting at `naquadah=100`, `exp=0`, `level=1`, `training_points=0`. Then:

- **Stage 1** — `UPDATE naquadah += 500 WHERE player_id = A`. Asserts A goes to 600 AND B stays at 100. Multi-character isolation regression guard for #143 — a refactor that swaps `WHERE player_id` for `WHERE account_id` credits both characters.
- **Stage 2** — `UPDATE exp/level/training_points WHERE player_id = A`. Asserts A's progression matches and B's is unchanged.
- **Stage 3** — reverse-direction `UPDATE naquadah += 500 WHERE player_id = B`. Asserts B goes to 600 AND A's prior 600 is undisturbed. Catches a hypothetical "always-credit-lowest-player_id-on-account" bug that stage 1's direction wouldn't expose.

Sentinel ids `-900003` and `-900004`.

## Implementation

`crates/services/src/base/smoke_tests.rs` (new module, wired in via `#[cfg(test)] mod smoke_tests;`):

- One `#[tokio::test]` per smoke, each gated by `require_db_or_skip!()` so the suite stays green on a fresh checkout without `DATABASE_URL`.
- Shared `strip_psql_metacommands(&str) -> String` helper drops the leading `\set ON_ERROR_STOP on` (psql-only; sqlx already errors on the first failed statement).
- Outer `BEGIN ... ROLLBACK` in each script keeps the seed data byte-identical post-test on both pass and fail. On fail, `RAISE EXCEPTION` aborts the tx and Postgres marks the session aborted; sqlx surfaces the error to the harness, the test panics, and the pool issues `ROLLBACK` on connection release.

## Test plan

- [x] `cargo test -p cimmeria-services --lib base::smoke_tests -- --test-threads=1` — 3/3 pass against bundled local Postgres
- [x] Post-run sentinel-id queries (`SELECT COUNT(*) FROM account WHERE account_id IN (-900002, -900003)` etc.) all return 0 — confirms rollback works
- [x] `cargo clippy -p cimmeria-services --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean

## Group D status after this

- ✅ vendor smoke
- ✅ inventory-move smoke
- ✅ progression smoke

### DIFF
--- a/crates/services/src/base/mod.rs
+++ b/crates/services/src/base/mod.rs
@@ -31,6 +31,9 @@ pub(crate) mod tick_sync;
 pub(crate) mod world_entry;
 pub(crate) mod world_entry_appearance;
 
+#[cfg(test)]
+mod smoke_tests;
+
 pub use service::BaseService;
 
 // ── Error types ───────────────────────────────────────────────────────────────
--- a/crates/services/src/base/smoke_tests.rs
+++ b/crates/services/src/base/smoke_tests.rs
@@ -1,5 +1,5 @@
 //! End-to-end live-DB smoke tests that exercise full multi-handler
-//! flows against the seeded database (Group D of #79).
+//! flows against the seeded database.
 //!
 //! Unlike the per-handler tests under
 //! `world_entry/methods/.../tests.rs`, these tests run a single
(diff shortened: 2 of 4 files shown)


Cadacious:
@coderabbitai review

Issue #207, opened 2026-05-04
Test coverage: residual feasible gaps + workspace ratchet to 75% / 80%


Cadacious:
## Summary

Workspace test coverage on `main` (per [run #25348231306](https://github.com/SandboxServers/Cimmeria/actions/runs/25348231306), 2026-05-04):

| Metric | Current | Target |
|---|---|---|
| Regions | 64.37 % | **75 % near-term, 80 % long-term** |
| Functions | 72.24 % | **80 %** |
| Lines | 64.49 % | **75 % near-term, 80 % long-term** |

This issue closes the **feasible-and-worthwhile** coverage gaps not already tracked by #205, #167, #190, #191, #149, #166, #165, and proposes a concrete codecov ratchet so we don't slide back. It is intentionally scoped — we are **not** chasing 100 %; the items below are the ones whose blast radius justifies the test-writing cost.

The companion test-suite review at [docs/testing/inventory/review-report.md](https://github.com/SandboxServers/Cimmeria/blob/main/docs/testing/inventory/review-report.md) (verdict: green-leaning yellow) describes *quality* gaps in the existing 1071-test suite. This issue tracks *quantity* gaps — code paths that currently have no executed assertions at all.

## Proposed coverage target

- **Ratchet floor: 75 % project line coverage by end of Q3 2026** (currently 64.49 %).
- **Steady-state target: 80 %** once the items below land.
- **Patch coverage: keep at 70 %** (already configured in `codecov.yml`).
- **Per-component floors (`codecov.yml` `component_management`)**: lock each crate at `current - 1%` so ratcheting up in one area can't be cancelled by drift somewhere else.
  - `mercury` >= 92 %, `entity` >= 88 %, `content-engine` >= 82 %, `commands` >= 92 %, `common` >= 98 %, `defs` >= 84 %.
  - `services` >= 62 % near-term, ratchet to 75 %.
  - `game` >= 80 % once `interactions/*` and `social/chat.rs` get a smoke each (currently 0 % on those four files).
  - `admin-api` is currently outside `component_management` (0 % across the board); add it as a component once it has any tests, with an initial floor of 50 %.
- **Exclude from the denominator** (add to `codecov.yml` `ignore:`): binary entry points where assertion-style tests are not idiomatic.
  - `crates/server/src/main.rs`, `crates/supervisor/src/main.rs`
  - `crates/upk-objects/src/bin/**`, `crates/upk/src/bin/**`
[...]

PR #226, opened 2026-05-06
test(cell): extract make_space_manager helper for cell test fixtures (#149)

### LINKED ISSUE #149: test infra: extract shared SpaceManager XML setup helper for cell test fixtures
## Background

While reviewing #140, Copilot pointed out that the per-test SpaceManager XML setup is duplicated across multiple cell test files. The same ~5-line block (parse spaces XML, parse cell-spaces XML, create startup spaces) appears in:

- [crates/services/src/cell/cell_methods/player/dispatch.rs](crates/services/src/cell/cell_methods/player/dispatch.rs) (3 tests)
- [crates/services/src/cell/cell_methods/player/interaction.rs](crates/services/src/cell/cell_methods/player/interaction.rs) (3 tests, has its own \`make_space_manager\`)
- [crates/services/src/cell/interactions/vendor.rs](crates/services/src/cell/interactions/vendor.rs) (4 tests, has its own \`make_space_manager\`)

Each file already has a local \`make_space_manager()\` that's nearly identical.

## Why a single PR

This was deferred from #140 because partial dedup in just one file would create more inconsistency, not less — once you have one shared helper, every other site should call it. Doing it as a single sweep keeps the test files aligned.

## Suggested approach

Extract a public \`make_space_manager()\` (or similar) into [crates/services/src/test_support.rs](crates/services/src/test_support.rs), gated on \`#[cfg(test)]\`, then update all sites to call it. Probably also worth a \`make_space_manager_with_player(entity_id)\` variant since most callers create-entity right after.

## Out of scope here

The XML strings hard-code "Agnos" / specific bounds. The helper should keep that as a default but allow overrides for tests that need different worlds (e.g. cross-world interaction tests further down the test-coverage burndown).

## References

- Copilot review on #140 (the comment that prompted this).
- Memory rule [feedback_no_pr_refs_in_comments.md](file:///C:/Users/steven.cady/.claude/projects/c--Users-steven-cady-repos-personal/memory/feedback_no_pr_refs_in_comments.md): the implementing PR shouldn't reference #140 in source comments — keep provenance in the PR body.

BangRocket:
## Summary

Closes #149.

Four cell test modules duplicated the same ~5-line block to set up a SpaceManager with a single Agnos space:

- \`cell/cell_methods/player/dispatch.rs\` — 3 inline copies
- \`cell/cell_methods/player/interaction.rs\` — local \`make_space_manager()\`
- \`cell/cell_methods/player/vendor.rs\` — local \`make_space_manager()\`
- \`cell/interactions/vendor.rs\` — local \`make_space_manager()\`

Extract a public \`make_space_manager()\` (and a \`make_space_manager_with_player(entity_id)\` variant for tests that need an avatar at the origin) into \`crates/services/src/test_support.rs\`, gated on \`#[cfg(test)]\`. All four sites now share the same default world geometry — a future Agnos-bounds change becomes one edit.

## Scope

- Net: +37 / -56 (boilerplate down).
- Helper keeps Agnos hard-coded as the default per #149's "out of scope" note. Adding a world-override variant is a follow-up if/when a cross-world test needs it.

## Test plan

- [ ] \`cargo test -p cimmeria-services\` (the four migrated test modules)
- [ ] \`cargo clippy -p cimmeria-services --all-targets -- -D warnings\`

## Related

- #149 — original issue
- #207 — codecov ratchet (lands first; this PR's coverage neutral, just dedup)

### DIFF
--- a/crates/services/src/cell/cell_methods/player/dispatch.rs
+++ b/crates/services/src/cell/cell_methods/player/dispatch.rs
@@ -48,7 +48,7 @@ pub async fn dispatch(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_support::make_space_manager;
+    use crate::test_support::make_space_manager_with_player;
 
     /// Pin the SGWPlayer cell-method constant values. The dispatch routing in
     /// `dispatch` above uses `..=` ranges over these constants — if a future
@@ -86,9 +86,7 @@ mod tests {
     /// 88..=90) returns false, and so does the outer dispatch.
     #[tokio::test]
     async fn pet_methods_route_to_social_not_world() {
-        let mut mgr = make_space_manager();
-        mgr.create_entity(1, "Agnos", [0.0, 0.0, 0.0], [0.0; 3])
-            .unwrap();
+        let mut mgr = make_space_manager_with_player(1);
 
         let (tx, _rx) = mpsc::channel(8);
         let engine = ChainEngine::new();
@@ -111,9 +109,7 @@ mod tests {
     /// as `dispatch` returning false for one of these.
     #[tokio::test]
     async fn each_outer_range_routes_to_a_handler() {
-        let mut mgr = make_space_manager();
-        mgr.create_entity(1, "Agnos", [0.0, 0.0, 0.0], [0.0; 3])
-            .unwrap();
+        let mut mgr = make_space_manager_with_player(1);
 
         let (tx, _rx) = mpsc::channel(64);
         let engine = ChainEngine::new();
@@ -140,9 +136,7 @@ mod tests {
 
     #[tokio::test]
     async fn out_of_range_methods_return_false() {
-        let mut mgr = make_space_manager();
-        mgr.create_entity(1, "Agnos", [0.0, 0.0, 0.0], [0.0; 3])
-            .unwrap();
+        let mut mgr = make_space_manager_with_player(1);
 
         let (tx, _rx) = mpsc::channel(8);
         let engine = ChainEngine::new();
--- a/crates/services/src/cell/cell_methods/player/dispatch.rs
+++ b/crates/services/src/cell/cell_methods/player/dispatch.rs
@@ -48,6 +48,7 @@ pub async fn dispatch(
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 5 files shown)


Cadacious (commented):
## PR #226 — `test(cell): extract make_space_manager helper for cell test fixtures (#149)`

### Top-level review comment

Helper extraction is the right shape per issue [#149](https://github.com/SandboxServers/Cimmeria/issues/149) and matches the `make_state()` precedent flagged in TESTING.md ("Reuse a small make_state() / make_ctx() helper rather than copying setup"). Two changes I'd ask for before merge.

1. **`make_space_manager_with_player` is dead code in this PR.** It's defined at [test_support.rs:104](crates/services/src/test_support.rs#L104) but the three tests in [cell/cell_methods/player/dispatch.rs](crates/services/src/cell/cell_methods/player/dispatch.rs) (lines 89, 114, 143 in the new file) still call `make_space_manager()` followed by `mgr.create_entity(1, "Agnos", [0.0, 0.0, 0.0], [0.0; 3])`. Those three sites are the canonical consumers — collapse them to `make_space_manager_with_player(1)` so the helper isn't introduced unused. CLAUDE.md is explicit on this: "Don't add features, refactors, abstractions … beyond the task. Delete unused code outright."

2. **Source comment cites the issue number.** `// ── SpaceManager test fixtures (issue #149) ───` at [test_support.rs:81](crates/services/src/test_support.rs#L81) violates copilot-instructions.md line 30 ("Don't reference the current PR or task — that rots in source; put it in the PR description") and TESTING.md "Comment hygiene." Drop the `(issue #149)` parenthetical; the section header alone is enough.

### Test-type matrix

This PR is a **test-fixture refactor**, not a behavior change — no new test types apply. The four call sites already have inline unit tests; those are unchanged. No inventory update needed because no tests were added or removed (per [docs/testing/inventory/maintenance.md](docs/testing/inventory/maintenance.md), fixture changes are a no-op for the inventory).

If you want a follow-up: TESTING.md mentions `make_state()` / `make_ctx()` as the canonical helper names; consider documenting `make_space_manager()` and `make_space_manager_with_player()` alongside them in TESTING.md's "File and module hygiene" section so the next contributor finds them by reading the playbook.

Issue #166, opened 2026-05-03
Live-DB test fixtures: hard-coded resource ids + outbox cleanup pattern across merged group A tests


Cadacious:
## Summary

Code review on the open group A PRs (#154, #158, #160, #162, #163) surfaced two test-brittleness patterns that also exist in already-merged group A tests but weren't reviewed because CodeRabbit / Copilot hit the rate limit during the burst of work. The fixes are uncontroversial; the question is just whether to bundle them into one cleanup PR or trickle them into the relevant files as those files get touched again.

## Pattern 1 — outbox cleanup placeholder

\`vendor/sell/tests.rs\` (PR #154, fixed) had:

\`\`\`rust
let _ = sqlx::query(\"DELETE FROM cell_event_outbox WHERE entity_id < 0\")
    .execute(pool)
    .await; // no-op safety net
\`\`\`

This deletes nothing because the entity_ids in scope are all positive sentinels. Successful runs leave outbox rows behind, which can pollute a shared live DB and surface as flaky failures elsewhere (we hit this once already — see the outbox-tests note in the PR thread for the recharge bundle, where stale rows from this pattern caused the outbox round-trip test to see N rows when it expected 1).

The same dead pattern exists in:

- [crates/services/src/base/world_entry/methods/vendor/purchase/tests.rs:38](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/base/world_entry/methods/vendor/purchase/tests.rs#L38) (merged in #156)

The fix is the same one applied in PR #154's review round: take an \`entity_id\` parameter and \`DELETE ... WHERE entity_id = \$1\`.

## Pattern 2 — hard-coded resource-table ids

Tests pin specific seed-data ids (e.g. \`SELLABLE_TYPE_ID = 21\`, \`PURE_CASH_DESIGN_ID = 5228\`, \`SEEDED_REPAIR_VENDOR_TEMPLATE_ID = 25\`). Copilot called this out on three of the open PRs (#158, #160, #162) and the fix was the same in each — query at runtime for *some* item satisfying the test's constraints, rather than naming a specific id. Following seed renumbers automatically removes a class of \"why is this test red?\" diagnosis time.

The same pattern exists in:

[...]

PR #227, opened 2026-05-06
test(vendor/purchase): outbox cleanup by entity_id, not silent placeholder (#166)

BangRocket:
## Summary

Closes the \`vendor/purchase/tests.rs\` bullet from #166.

The cleanup helper used the no-op safety net pattern that issue #166 calls out:

\`\`\`rust
DELETE FROM cell_event_outbox WHERE entity_id < 0
\`\`\`

This deletes nothing — the \`entity_id\`s in scope are all positive sentinels. Successful runs left outbox rows behind, which can pollute a shared live DB and surface as flaky failures elsewhere (the recharge-bundle outbox round-trip test was the canonical victim — it expected 1 row, saw N).

PR #154 fixed the same shape in \`vendor/sell/tests.rs\`. This PR applies that fix to \`vendor/purchase/tests.rs\`:

- Take an \`entity_id\` parameter on \`cleanup\`
- \`DELETE FROM cell_event_outbox WHERE entity_id = \$1\`
- Each of the four purchase tests already had its own sentinel; thread it through

## Why this PR (not bundled)

#166 also lists \`buyback/tests.rs\`, \`paid_repair/tests.rs\`, \`data/tests.rs\`, and \`repair.rs\` for the same sweep. Those are the *hard-coded resource ids* pattern, which is a different fix shape (replace constants with runtime SELECTs) and #166's recommendation is "fold into the next PR that already touches each file." The outbox-cleanup pattern in purchase is the one fix that's a drop-in copy from PR #154 and worth landing on its own.

## Test plan

- [ ] \`DATABASE_URL=postgres://w-testing:w-testing@localhost:5433/sgw cargo test -p cimmeria-services --lib -- --test-threads=1\` — all four purchase tests still pass
- [ ] \`cell_event_outbox\` is empty after the test run (visual check via \`SELECT COUNT(*) FROM cell_event_outbox WHERE entity_id IN (...)\`)

## Related

- #166 — outbox + hard-coded-id sweep
- #154 — the vendor/sell PR this fix copies

### DIFF
--- a/crates/services/src/base/world_entry/methods/vendor/purchase/tests.rs
+++ b/crates/services/src/base/world_entry/methods/vendor/purchase/tests.rs
@@ -35,6 +35,8 @@ const ITEM_COST_DESIGN_ID: i32 = 5192;
 const ITEM_COST_PREREQ_DESIGN_ID: i32 = 55;
 
 async fn cleanup(pool: &PgPool, entity_id: i32, account_id: i32, player_id: i32) {
+    // Delete the outbox rows the test enqueues so a shared live DB
+    // doesn't accumulate stale entries from successful runs.
     let _ = sqlx::query("DELETE FROM cell_event_outbox WHERE entity_id = $1")
         .bind(entity_id)
         .execute(pool)
@@ -176,11 +178,11 @@ async fn pure_cash_purchase_debits_balance_and_grants_inventory_row() {
     // the exact delta rather than just "lower than before".
     insert_account_and_player(&pool, account_id, player_id, 5_000).await;
 
-    let (socket, e2a, conn) = make_state(0x7000_0A01);
+    let (socket, e2a, conn) = make_state(entity_id as u32);
     let db_pool = Some(Arc::new(pool.clone()));
 
     handle_purchase_vendor_items(
-        0x7000_0A01,
+        entity_id as u32,
         player_id,
         99,
         SEEDED_BUY_VENDOR_TEMPLATE_ID,
@@ -230,11 +232,11 @@ async fn item_prereq_purchase_consumes_prereq_and_skips_cash_update() {
     let prereq_item =
         insert_item(&pool, player_id, ITEM_COST_PREREQ_DESIGN_ID, INV_MAIN, 5, 3).await;
 
-    let (socket, e2a, conn) = make_state(0x7000_0A11);
+    let (socket, e2a, conn) = make_state(entity_id as u32);
     let db_pool = Some(Arc::new(pool.clone()));
 
     handle_purchase_vendor_items(
-        0x7000_0A11,
+        entity_id as u32,
         player_id,
         99,
         SEEDED_BUY_VENDOR_TEMPLATE_ID,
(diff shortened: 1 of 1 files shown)


Cadacious (commented):
## PR #227 — `test(vendor/purchase): outbox cleanup by entity_id, not silent placeholder (#166)`

### Top-level review comment

The fix is correct — `entity_id < 0` matched zero rows, and replacing it with `entity_id = $1` does what the cleanup helper was always supposed to do. But the PR currently has **no regression-guard test**, which is the bug shape TESTING.md is most explicit about.

> *TL;DR #2 in [TESTING.md](TESTING.md#tldr): A regression guard must reproduce the bug if the fix is reverted. If the test still passes after `git revert`-ing the fix, it isn't a regression guard.*

Reverting this PR (re-introducing `entity_id < 0`) doesn't make any test fail. The bug it claims to fix — outbox rows leaking and polluting later tests — is observable but unpinned.

#### What's missing — and what test types would pin it

Per the user's note that one feature can need several test types, this fix sits at exactly the "live-DB regression guard" shape TESTING.md flags as the right tool for SQL invariants:

| Test type | Why it applies | Suggested shape |
|---|---|---|
| **Live-DB regression guard** (the load-bearing one here) | The bug is a SQL `WHERE` predicate that matched zero rows. Per TESTING.md "Live-DB regression guards" — exactly the case this type exists for. | New test in the same file: insert a row into `cell_event_outbox` with the test's sentinel `entity_id`, run the existing cleanup, then `SELECT COUNT(*) FROM cell_event_outbox WHERE entity_id = $sentinel` and assert `== 0`. Reverting the fix to `entity_id < 0` leaves the count at 1 and fails. |
| **PL/pgSQL smoke** | Not really applicable — the cleanup helper is one statement; a smoke would be over-engineered. | Skip. |
| **Unit test** | The SQL string is a literal; you could pin it. But asserting against the SQL string is fragile and the live-DB test above proves the same thing one layer down. | Skip. |
| **Concurrency guard** | Cleanup runs inside one test; there's no race here. | Skip. |

So: one new live-DB test, ~15 lines, in the same `mod tests` block. It's the only thing the PR is missing.

### Inline notes

#### `purchase/tests.rs:170-178` (and the three sibling tests)

Each test now declares the entity sentinel twice: once as `let entity_id: i32 = 0x7000_0A01;` and once inline as `make_state(0x7000_0A01)`. The literals are duplicated four times across the file. A future refactor that updates one but not the other would silently break test isolation — exactly the brittleness pattern issue #166 was filed about.

Either:
- Promote each test's sentinel to a `const TEST_ENTITY_ID_PURE_CASH: i32 = 0x7000_0A01;` (matching the existing `TEST_BASE` constant style at [purchase/tests.rs:15](crates/services/src/base/world_entry/methods/vendor/purchase/tests.rs#L15)), or
- Pass `entity_id as u32` into `make_state(entity_id as u32)` so the literal lives in one place per test.

The second is fewer lines.

#### `purchase/tests.rs:39`

The trailing comment `// no-op safety net` is gone — good. The new `WHERE entity_id = $1` form is self-explanatory; no replacement comment needed.

### Test inventory

No inventory update needed because the test count is unchanged. **If you add the regression guard described above, that's one row in [docs/testing/inventory/services.md](docs/testing/inventory/services.md).**

Issue #167, opened 2026-05-03
Test coverage: DB-driven testing opportunities beyond group A


Cadacious:
## Summary

Group A (#79) covered the highest-risk vendor / inventory / progression / mail / missions paths with live-DB regression guards. The same approach — \`require_db_or_skip!()\` + \`postgres:17.9\` service container in CI + sentinel id ranges — has paid off there: it caught real regressions early, found subtle issues that mocked tests would have missed (#84 concurrency, #117–#121 in #77), and ate one round of CodeRabbit / Copilot follow-ups in stride.

This issue tracks where we should extend that pattern next, ranked by blast radius. None of these items is currently broken — this is about catching future regressions early.

## Tier 1 — security / data isolation (high priority)

These functions decide which account or character can read/write what. A regression here is a session-confusion or privacy bug, not a UX blip.

- [ ] **\`auth/handlers.rs\` login flow** — \`validate_account_credentials\` ([handlers.rs:420-459](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/auth/handlers.rs#L420-L459)) reads \`account.password / accesslevel / enabled\`. Tests should pin: case-insensitive password compare, \`enabled = false\` rejection, \`InvalidCredentials\` for both \"no row\" and \"wrong password\" (the comment explicitly notes this matches C++ \`BadUserPassword\` — must not leak whether the account exists).
- [ ] **\`base/character.rs::handle_delete_character\`** — \`DELETE FROM sgw_player WHERE player_id = \$1 AND account_id = \$2\` ([character.rs:117-121](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/base/character.rs#L117-L121)). Same account-isolation concern as \`query_character_list\` (PR #163): a refactor that drops the \`AND account_id\` predicate would let any account delete any character. \`rows_affected = 0\` path also worth pinning.
- [ ] **\`base/character.rs::handle_request_character_visuals\`** — \`SELECT bodyset, components, skin_color_id, bandolier_slot WHERE player_id = \$1 AND account_id = \$2\` ([character.rs:172-179](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/base/character.rs#L172-L179)). Same shape, same test pattern.
[...]

PR #228, opened 2026-05-06
test(base+auth): pin delete_character no-DB short-circuit + auth login parse errors (#167)

BangRocket:
## Summary

First slice of the #167 DB-driven testing roadmap. Picks the two smallest items that don't require new live-DB fixtures so they can land independently of the bigger character-isolation tests.

### \`base/character.rs::handle_delete_character\` — no-DB short-circuit

The handler must return \`Ok\` without panicking when the DB pool is \`None\` (developer-mode boot without postgres). Pin that branch so a future refactor that unconditionally derefs the pool gets caught at this seam, not at boot in a developer environment.

### \`auth/handlers.rs::parse_login_request\` + \`parse_server_selection\`

Two adversarial-input tests for the XML attribute parsers:
- Missing required attributes (\`SKU\`, \`AccountName\`, \`Password\` / \`ServerSelection\`) must return \`Err\`, not panic.
- A short password (e.g. \`"short"\`) parses to a string the length-validation layer above will reject. The parser doesn't filter at this stage, which matches the existing C++ shape per the comment in \`handle_login\`. Pinning that keeps the parser's responsibility narrow.

## Out of scope (followup PRs)

Per #167's tier list, the items that need new live-DB fixtures + sentinel id ranges:

- \`handle_request_character_visuals\` account-id binding (Tier 1)
- \`handle_create_character\` starter-kit cascade + name-collision rejection (Tier 1)
- \`inventory/core.rs::handle_remove_inventory_item\` DELETE-vs-UPDATE branch (Tier 2)
- \`grant.rs::item_allows_container\` 3-branch lookup (Tier 2)
- \`gate_travel.rs\` \`world_location\` UPDATE + \`known_stargates\` array idempotence (Tier 2)

Each is its own bug shape; bundling all of #167 into one PR would be ~600 lines and hard to review.

## Test plan

- [ ] \`cargo test -p cimmeria-services --lib auth::handlers::tests::parse_malformed_login_request_returns_error\`
- [ ] \`cargo test -p cimmeria-services --lib auth::handlers::tests::parse_server_selection_malformed_returns_error\`
- [ ] \`cargo test -p cimmeria-services --lib base::character::delete_character_tests::delete_character_no_db_short_circuits\`
- [ ] \`cargo clippy -p cimmeria-services --all-targets -- -D warnings\`

## Related

- #167 — DB-driven testing roadmap (this PR closes the no-DB short-circuit + parse-error bullets only)

### DIFF
--- a/crates/services/src/auth/handlers.rs
+++ b/crates/services/src/auth/handlers.rs
@@ -546,22 +546,40 @@ mod tests {
         assert!(xml.contains(r#"Port="32832""#));
     }
 
+    /// Pin the parser's narrow contract: when the SGWLoginRequest element
+    /// is present but required attributes (SKU/AccountName/Password) are
+    /// missing, the parse succeeds with default (empty) fields. The
+    /// handler validates above the parser; a refactor that pushed
+    /// validation down here would silently break that layering.
     #[test]
-    fn parse_malformed_login_request_returns_error() {
-        // Missing required attributes.
+    fn parse_login_request_does_not_validate_missing_attributes() {
         let body = r#"<sgwLogin:SGWLoginRequest xmlns:sgwLogin="http://www.stargateworlds.com/xml/sgwlogin" />"#;
-        let req = parse_login_request(body);
-        assert!(req.is_err(), "missing SKU/AccountName/Password must fail");
-
-        // Empty password that doesn't match SHA-1 length.
-        let body2 = r#"<sgwLogin:SGWLoginRequest xmlns:sgwLogin="http://www.stargateworlds.com/xml/sgwlogin" SKU="SGW_BETA" AccountName="test" Password="short" ProtocolDigest="58AFA196AD3AC4F65CADD99BFF23B799" />"#;
-        let req2 = parse_login_request(body2).unwrap();
-        assert_eq!(req2.password, "short");
-        // The handler later validates length (40) and hex chars.
+        let req = parse_login_request(body).expect("element present must parse Ok");
+        assert_eq!(req.sku, "", "missing SKU must surface as empty, not error");
+        assert_eq!(req.account_name, "");
+        assert_eq!(req.password, "");
     }
 
+    /// Same narrow-contract pin, but for password length: a password the
+    /// handler will reject (not 40 hex chars) parses through unchanged.
+    /// Validation lives in the handler — this guard keeps it there.
     #[test]
-    fn parse_server_selection_malformed_returns_error() {
+    fn parse_login_request_does_not_validate_password_length() {
+        let body = r#"<sgwLogin:SGWLoginRequest xmlns:sgwLogin="http://www.stargateworlds.com/xml/sgwlogin" SKU="SGW_BETA" AccountName="test" Password="short" ProtocolDigest="58AFA196AD3AC4F65CADD99BFF23B799" />"#;
+        let req = parse_login_request(body).expect("well-formed element must parse Ok");
+        assert_eq!(
+            req.password, "short",
+            "parser must hand the raw password through; length check is the handler's job",
+        );
+    }
+
+    /// Unlike parse_login_request, parse_server_selection requires the
+    /// ServerSelection attribute — the function returns early on the
+    /// first ServerSelection it sees, otherwise falls through to Err.
+    /// A refactor that loosened this would let the auth flow accept a
+    /// server-selection message with no shard id.
+    #[test]
+    fn parse_server_selection_missing_attribute_returns_error() {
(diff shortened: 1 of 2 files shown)


Cadacious (commented):
## PR #228 — `test(base+auth): pin delete_character no-DB short-circuit + auth login parse errors (#167)`

### Top-level review comment

Three useful guards, but two issues to fix and one scope question.

#### 1. `parse_malformed_login_request_returns_error` runs two distinct assertions, only one matches the name

[handlers.rs:551-560](crates/services/src/auth/handlers.rs#L551-L560):

```rust
fn parse_malformed_login_request_returns_error() {
    // Body 1: missing attributes → asserts is_err()    ✓ matches name
    // Body 2: short password "short" → asserts Ok with password=="short"  ✗ this is the happy path
}
```

Per TESTING.md "Naming" / PRs #133, #139, #141, #143 — test names must match the assertion. A reviewer who only reads the name expects the function to assert errors. Body 2 asserts a successful parse.

Split into two:

```rust
#[test]
fn parse_login_request_missing_attributes_returns_error() { ... }   // body 1

#[test]
fn parse_login_request_does_not_validate_password_at_parse_layer() { ... }  // body 2
```

The second name now describes what's actually pinned: the parser's narrow contract (don't validate length, that's the next layer's job). The trailing comment `// The handler later validates length (40) and hex chars.` becomes the *test name*, not a side note.

#### 2. Test inventory drift — review-blocker per copilot-instructions.md line 47

This PR adds 3 tests but doesn't touch [docs/testing/inventory/services.md](docs/testing/inventory/services.md). Per CLAUDE.md's doc-update map ("Add or remove a test (any kind) → docs/testing/inventory/<crate>.md") this is required in the same PR. Three rows: two parse-error tests under `auth/handlers.rs`, one no-DB short-circuit test under `base/character.rs`.

#### 3. Scope: title undersells what's left of #167

PR title: "pin delete_character no-DB short-circuit + auth login parse errors (#167)". But [#167](https://github.com/SandboxServers/Cimmeria/issues/167)'s Tier 1 isn't really about no-DB short-circuits — it's about **account-isolation guards**:

> `DELETE FROM sgw_player WHERE player_id = $1 AND account_id = $2`. Same account-isolation concern as `query_character_list` (PR #163): a refactor that drops the `AND account_id` predicate would let any account delete any character.

The PR explicitly defers that ("requires new live-DB fixtures + sentinel id ranges"), which is fine — but the deferred path is the load-bearing one. Consider re-titling along the lines of "pin auth parse-layer narrow contract + delete_character no-DB pool branch (#167 Tier 1 prerequisites)" so the issue stays open with a clearer "what's left" pointer.

### Test-type matrix for `handle_delete_character` (the deferred work)

If `#167` Tier 1 lands as a follow-up, here's the matrix the issue is asking for — useful for the follow-up PR's scope:

| Test type | What it pins | Notes |
|---|---|---|
| **Unit test** (this PR) | No-DB short-circuit returns `Ok(())` without panic. ✓ done. | |
| **Live-DB regression guard** (deferred) | (a) `WHERE player_id = $1 AND account_id = $2` actually filters by both columns; (b) `rows_affected == 1` for matching delete, `== 0` for cross-account attempt; (c) other characters on the same account untouched. | The load-bearing one. Pattern: `crates/services/src/base/character.rs::query_character_list` tests (PR #163). |
| **Wire-format test** | Not applicable — `handle_delete_character` is a handler, not a serializer. | Skip. |
| **Concurrency guard** | Two concurrent deletes of the same character — the second should be a no-op (`rows_affected == 0`), not a panic. | Probably worth a small test in the concurrency_tests sibling. |
| **Smoke** | If a future feature touches both `handle_delete_character` and `handle_create_character` (e.g., delete-then-recreate-with-same-name), a smoke that round-trips the pair would catch state drift. | Skip until that feature exists. |

### Inline notes

#### `handlers.rs:557` — comment claims a contract this test doesn't pin

`// The handler later validates length (40) and hex chars.` This is a TODO disguised as documentation. If you split the test as suggested in (1) above, rename and drop the comment. If you keep it, add a sibling test that drives `handle_login` end-to-end with `Password="short"` and asserts the rejection — that pins the actual contract.

#### `character.rs:412` — `delete_character_no_db_short_circuits` regression-guard shape

The assertion is `result.is_ok()`. That's tight enough for the named bug (a refactor that unconditionally derefs `pool` would panic, panic = test fails). Good as-is.

One nit: the test takes `&None` for the pool, so the bug shape is "`db_pool: None` does not panic." Consider asserting that no UDP packet was sent on `socket` either — the no-DB path is supposed to short-circuit *silently*, and a regression that returned `Ok(())` after sending an error packet would still pass. ~3 lines via the existing `socket.try_recv_from` pattern.

Issue #191, opened 2026-05-04
test(cell/service): npc_ai_leash should pin its onStatUpdate + onStateFieldUpdate wire packets


Cadacious:
PR #181's `npc_ai_leashing_snaps_to_spawn_restores_health_and_idles` test asserts the entity-state side of the leash tick (snap-to-spawn position, HEALTH restored, AI → Idle, threat list cleared, cooldowns cleared) but never inspects the `tx` channel. Copilot review flagged the gap:

> `npc_ai_leash` also emits the stat/state method calls that tell the client about the heal and AI reset, but this regression test never inspects `tx`. A break in those packets would still leave the in-game test green.

## What's missing

[`crates/services/src/cell/service/npc_ai.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service/npc_ai.rs) — `npc_ai_leash` ends with two `send_entity_method` calls:

```rust
super::super::abilities::send_entity_method(npc_id, 20, stat_update, tx, space_mgr).await;
// ...
super::super::abilities::send_entity_method(npc_id, 19, state_args, tx, space_mgr).await;
```

Without a witness in the NPC's space, both calls hit the empty-witness branch and silently drop. Without the witness fixture the assertions on tx wouldn't fire either, but the test still passed for that wrong reason.

## Suggested test shape

Add a player witness to the leash fixture (the non-instanced "Castle" world is already used by `make_ai_fixture` after PR #181 round 2, so adding one entity + `connect_entity` + `compute_aoi_changes` is a few lines), then drain `rx` and assert:

1. Exactly one `WitnessEntityMethod { entity_id == 200, method_index == 20 }` (the AmmoSlot/HEALTH stat update post-heal).
2. Exactly one `WitnessEntityMethod { entity_id == 200, method_index == 19 }` (the onStateFieldUpdate carrying the post-leash state_field).
3. Both arrive in that order — stat update before state-field update is the order in `npc_ai_leash` and the canonical pattern across the abilities module.

The packet-byte assertions don't need to be byte-exact (the inner serialize_dirty layout is already covered elsewhere); just pin presence + ordering + the entity_id/method_index pair.

## Why filing this separately

PR #181 covered the entity-state side of the leash tick. The wire-side coverage is genuinely additional test surface, not a correction to the existing test, so it makes more sense as its own ticket than as a re-opened sub-bullet on #123.

## Related

- #123 — testing roadmap parent.
[...]

Issue #190, opened 2026-05-04
test(cell/service): npc_ai_fight should pin top-threat selection across multiple live targets


Cadacious:
PR #181 added regression guards for the four arms of `npc_ai_tick` (Fighting → Idle on empty threat, Fighting → Leashing on out-of-spawn-range target, dead-target prune, stationary-no-pathfind, Leashing → Idle). One coverage gap remains, flagged by Copilot review:

> none of these new tests exercise the branch where a Fighting NPC chooses between multiple live threat targets. `npc_ai_fight` still has an unpinned `partial_cmp` on the threat values that decides who gets attacked.

## What's missing

[`crates/services/src/cell/service/npc_ai.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service/npc_ai.rs) — the `npc_ai_fight` body picks the top-threat target via:

```rust
let top = npc.threat_list
    .iter()
    .max_by(|a, b| a.1.partial_cmp(b.1).unwrap_or(std::cmp::Ordering::Equal))
    .map(|(&eid, _)| eid);
```

A regression that flips `min_by` for `max_by` (or returns the first entry instead of the max) would still let the existing dead-target test pass — it only seeds two threats, and after the dead one is pruned the surviving one is selected by default.

## Suggested test shape

In `crates/services/src/cell/service/tests.rs`:

1. Seed three live targets (e.g., 100, 101, 102) with distinct, non-NaN threat values where the highest is NOT the first inserted (HashMap iteration order is non-deterministic, so make the assertion robust to that).
2. Run `npc_ai_tick` once.
3. Assert that the NPC's `nav_path` (or whatever in-range/LOS branch was taken) targets the highest-threat entity, not one of the lower-threat ones.

For NaN handling, also add a single test that seeds a target with `f32::NAN` threat and asserts the tick doesn't panic — `partial_cmp` returns `None` for NaN and the current code falls back to `Ordering::Equal`, but pinning that branch keeps it explicit.

## Why filing this separately

PR #181 closed the named `npc_ai_tick` item from #123 with the four state-transition guards. This is an unpinned sub-branch of `Fighting` that didn't have an explicit checklist line on #123, so I'm tracking it as its own issue rather than re-opening that bullet.

## Related

- #123 — testing roadmap parent.
- #181 — original `npc_ai_tick` coverage PR (merged).

aurablacklight:
[...]

Issue #123, opened 2026-05-03
Test coverage expansion: services / mercury / content-engine (post bug-bash)


Cadacious:
Audit of test coverage across `crates/services`, `crates/mercury`, and `crates/content-engine` performed during the bug bash. Filing this so we can pick up the work once the bash ends — not before, to avoid churn against bug-fix PRs.

> **Scope note.** This issue is **complementary** to #79, which already covers the PR #77 surface (vendor / inventory / progression / mail / missions / player dispatch / `mercury/types.rs::PlayerLoadData` round-trip / `vendor_store_smoke.sql`). Items there are not duplicated here. This issue focuses on everything else.

> **Status as of 2026-05-04.** All six groups (A–F) are done or substantially done. **Groups B, C, D, E, F are 100% complete.** Group A is substantially closed — every named module now has at least one test except a few deferred areas (heavy socket plumbing where fixture cost > test value) and a small set of remaining open files tracked as residuals. Total of **22 PRs merged** under this issue: #178–#187 (Group A), #192–#198 (Group B), #199–#201 (Group C), #202–#204 (Group E), with infra in #134, #151, #153, #172, #177. Three follow-ups filed: #190, #191 (npc_ai gaps from Group A), plus a residuals issue tracking remaining Group A files (see Acceptance below).

## Current state (baseline)

- `crates/mercury/` — 8/8 wire modules have inline tests (`bundle`, `channel`, `codec`, `encryption`, `messages`, `packet`, `unified`, `unpacker`). Coverage is healthy at the unit level but has known gaps at the boundary (see #120, #121).
- `crates/content-engine/` — 6/6 modules have inline tests + 1 integration test (`tests/interact_tag_linter.rs`). No chain-replay harness.
- `crates/services/` — spot coverage. Tested: `cell/abilities` (timers, RNG, fire), `cell/cell_methods/inventory` (bandolier swap), `cell/dispatch`, `cell/messages`, `cell/ring_transport`, `cell/service` (reload tick), `cell/space_manager` (basic), `cell/spawner` (basic), `mercury/protocol`, `mercury/world_data`, `base/world_entry`, plus inline tests in auth, base/login, base/character_create, combat/damage, combat/state, combat/threat, interactions/dispatch, minigame/protocol, minigame/session, database. Untested: large swaths of base service, cell content/spawner/space-manager internals, services/mercury AoI builders, minigame games, orchestrator-Postgres.
[...]

PR #229, opened 2026-05-06
test(cell/service): npc_ai_fight top-threat + npc_ai_leash wire packets (#190, #191)

BangRocket:
## Summary

Closes the two \`npc_ai\` coverage gaps Copilot flagged on PR #181 — both gaps that would let a real regression land green.

### #190 — \`npc_ai_fight\` top-threat selection across multiple live targets

The existing dead-target test only seeds two threats; after the dead one is pruned the survivor is selected by default, so the \`max_by\` branch never fires. A regression that flipped \`max_by\` for \`min_by\`, or returned the first \`HashMap\` entry instead of the max, would still pass.

New test \`npc_ai_fight_picks_top_threat_among_multiple_live_targets\`:
- Seeds three live targets at distinct positions (100/101/102)
- Inserts threats in NON-monotonic HashMap order: 100→2.0, 102→5.0, 101→10.0
- Asserts the in-range top-threat (101) drives the attack branch (\`nav_path\` cleared)

Companion test \`npc_ai_fight_nan_threat_does_not_panic\`:
- Seeds a target with \`f32::NAN\` threat
- \`partial_cmp\` returns \`None\` for NaN; the current code falls back to \`Ordering::Equal\`
- Pin that fallback so a future refactor that unwraps without a fallback cannot land

### #191 — \`npc_ai_leash\` \`onStatUpdate\` + \`onStateFieldUpdate\` wire packets

PR #181's leash test asserted the entity-state side (snap-to-spawn, HEALTH restored, AI → Idle, threat list cleared) but never inspected \`tx\`. Without a witness the \`send_entity_method\` calls hit the empty-witness branch and silently drop — a break in those packets would still leave the test green.

New test \`npc_ai_leash_emits_stat_update_then_state_field_to_witnesses\`:
- Adds a player witness in the same space (Castle, non-instanced)
- Computes AoI so the player witnesses NPC 200
- Drains \`rx\` and asserts:
  - Exactly one method 20 (\`onStatUpdate\`, post-heal HEALTH)
  - Exactly one method 19 (\`onStateFieldUpdate\`, post-leash state-field carry)
  - Both for \`entity_id == 200\` (the NPC)
  - 20 arrives before 19 (canonical stat-then-state ordering across the abilities module)

## Test plan

- [ ] \`cargo test -p cimmeria-services --lib npc_ai_fight_picks_top_threat_among_multiple_live_targets\`
- [ ] \`cargo test -p cimmeria-services --lib npc_ai_fight_nan_threat_does_not_panic\`
- [ ] \`cargo test -p cimmeria-services --lib npc_ai_leash_emits_stat_update_then_state_field_to_witnesses\`
- [ ] Reverting \`max_by\` → \`min_by\` in \`npc_ai_fight\` makes the top-threat test fail (regression-guard verification per CLAUDE.md)
- [ ] Removing one of the two \`send_entity_method\` calls in \`npc_ai_leash\` makes the wire-packet test fail

## Related

- #190, #191 — original coverage-gap issues
- #181 — \`npc_ai_tick\` coverage PR these gaps were carved off from
- #123 — testing roadmap parent

### DIFF
--- a/crates/services/src/cell/service/tests/npc_ai.rs
+++ b/crates/services/src/cell/service/tests/npc_ai.rs
@@ -219,19 +219,29 @@ async fn npc_ai_leashing_snaps_to_spawn_restores_health_and_idles() {
 }
 
 /// Fighting NPC with three live threats must pick the highest-threat
-/// target for attack. The existing dead-target test only seeds two
-/// threats and the dead one is removed, so the survivor is selected by
+/// target for attack. The dead-target sibling test only seeds two
+/// threats and the dead one is pruned, so the survivor is selected by
 /// default — it never exercises the `max_by` branch. This test pins
-/// that the top threat is chosen even when it is NOT the first inserted.
+/// that the top threat is chosen by varying the geometry: only the
+/// top-threat target (101) is in NPC_ATTACK_RANGE; 100 and 102 are
+/// past 30. With pre-seeded `nav_path`, the in-range attack branch
+/// clears it, and the out-of-range pathing branch leaves it set. So:
+/// max_by correct → 101 picked → in range → nav_path empty.
+/// max_by flipped → 100 or 102 picked → out of range → nav_path stays.
 #[tokio::test]
 async fn npc_ai_fight_picks_top_threat_among_multiple_live_targets() {
+    use cimmeria_common::Vector3;
+
     let mut mgr = make_ai_fixture([0.0; 3], [0.0; 3]);
 
-    // Three live targets at different positions (all in range + LOS).
+    // Position so only the top-threat target is in NPC_ATTACK_RANGE (30):
+    //   100 (threat 2.0)  → distance 40 (out of range, within LEASH=50)
+    //   102 (threat 5.0)  → distance 35 (out of range, within LEASH=50)
+    //   101 (threat 10.0) → distance 10 (in range)
     for &(eid, pos) in &[
-        (100, [5.0, 0.0, 0.0]),
+        (100, [40.0, 0.0, 0.0]),
         (101, [10.0, 0.0, 0.0]),
-        (102, [15.0, 0.0, 0.0]),
+        (102, [-35.0, 0.0, 0.0]),
     ] {
         mgr.create_entity(eid, "Castle", pos, [0.0; 3]).unwrap();
         if let Some(p) = mgr.get_entity_mut(eid) {
@@ -249,6 +259,11 @@ async fn npc_ai_fight_picks_top_threat_among_multiple_live_targets() {
         npc.threat_list.insert(100, 2.0);
         npc.threat_list.insert(102, 5.0);
         npc.threat_list.insert(101, 10.0);
+        // Pre-seed nav_path so the assertion can distinguish:
+        //  - in-range branch explicitly calls `nav_path.clear()`
+        //  - out-of-range pathing branch leaves it set if find_path
+        //    returns None (no navmesh in the test fixture).
+        npc.nav_path.push_back(Vector3::new(99.0, 0.0, 99.0));
     }
 
     let (tx, _rx) = mpsc::channel(8);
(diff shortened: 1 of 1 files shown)


Cadacious (commented):
## PR #229 — `test(cell/service): npc_ai_fight top-threat + npc_ai_leash wire packets (#190, #191)`

### Top-level review comment

The leash test is solid. The two `npc_ai_fight` tests do not exercise the branches their names claim.

#### 1. **Blocker:** `npc_ai_fight_picks_top_threat_among_multiple_live_targets` does not test top-threat selection

[npc_ai.rs:226-269](crates/services/src/cell/service/tests/npc_ai.rs):

The test seeds three live targets at distances 5, 10, 15 from origin with non-monotonic threats (100→2.0, 102→5.0, 101→10.0), then asserts `npc.nav_path.is_empty()` because "the in-range top threat must trigger attack branch."

**The bug:** all three targets are in range and in LOS. Whichever one `max_by` picks (or `min_by` if flipped), the in-range attack branch fires for the picked target, `nav_path` clears, and the assertion passes.

The PR's own Test Plan claims:
> Reverting `max_by` → `min_by` in `npc_ai_fight` makes the top-threat test fail (regression-guard verification per CLAUDE.md)

That isn't actually true with the current assertions. To verify, flip the comparator locally and re-run — the test will still pass, exactly the failure mode TESTING.md "Regression-guard shape" warns about (PR #143's "happy-path test, not a guard").

Per [#190](https://github.com/SandboxServers/Cimmeria/issues/190)'s suggested shape:
> Assert that the NPC's nav_path (or whatever in-range/LOS branch was taken) **targets the highest-threat entity**, not one of the lower-threat ones.

The "targets the highest-threat entity" part needs an observable that differs by target choice. Two options:

- **Wire-side assertion (preferred):** drain `tx`/`rx` and assert a `WitnessEntityMethod` for `entity_id == 200` carries the top-threat target's id in its arg payload (the `onTargetUpdate` or `onAbilityCast` whose target the attack branch fires at). This makes it a wire-format test in addition to a unit test — exactly the multi-type coverage TESTING.md "When one feature needs more than one test" recommends.
- **Geometry-side assertion:** position only the highest-threat target in range; put 100 (threat 2.0) and 102 (threat 5.0) out of range. Then assert `npc.nav_path` is empty AND that the NPC didn't pivot to face 100 or 102. `min_by` would pick 100, which is out of range → pathing branch → `nav_path` non-empty → test fails.

Option 1 is the more robust shape because it asserts the *target id* directly rather than relying on the geometry to make `min_by` observable.

#### 2. **Blocker:** `npc_ai_fight_nan_threat_does_not_panic` doesn't exercise the NaN fallback

[npc_ai.rs:272-303](crates/services/src/cell/service/tests/npc_ai.rs):

The test seeds **one** target with NaN threat. With one element, `max_by` doesn't compare anything — the iterator's single item is returned. The `partial_cmp` → `Ordering::Equal` fallback the test claims to pin is never reached.

To exercise the fallback, seed **two** targets — one with NaN, one with a finite threat (say 5.0). `partial_cmp(NaN, 5.0)` returns `None`, falls back to `Equal`, and `max_by` returns whichever is last in HashMap order (deterministic enough to assert no panic + Fighting state). Then a regression that replaced `unwrap_or(Ordering::Equal)` with `unwrap()` would panic on the comparison and fail the test.

Recommend: keep the single-target test but rename to `npc_ai_fight_single_nan_target_does_not_panic` (it does pin "no panic on NaN reaching the iterator"), and add a second test `npc_ai_fight_nan_in_threat_list_with_other_targets_does_not_panic` that actually drives the fallback.

#### 3. `npc_ai_leash_emits_stat_update_then_state_field_to_witnesses` — solid, one extension worth considering

[npc_ai.rs:306-378](crates/services/src/cell/service/tests/npc_ai.rs):

Three tight assertions (`witness_methods.len() == 2`, exact method indices 20 then 19, ordering pinned). This is the right shape and matches issue [#191](https://github.com/SandboxServers/Cimmeria/issues/191)'s suggestion.

One **extension** (not a blocker): per TESTING.md "Wire-format tests" — "Assert the **exact byte string** the function emits, hand-written in the test as a `&[u8]` literal or a hex string." This test pins method indices but not the body bytes. A regression that changed the encoded post-leash HEALTH value or the post-leash state-field carry would still pass.

The issue itself acknowledges this:
> The packet-byte assertions don't need to be byte-exact (the inner serialize_dirty layout is already covered elsewhere); just pin presence + ordering + the entity_id/method_index pair.

If `serialize_dirty` byte coverage is genuinely covered elsewhere, link to those tests in this test's doc-comment so a reviewer touching the leash code path can find them. If not, this is the natural place to add a byte-exact assertion on the HEALTH delta.

### Test-type matrix

What this PR claims to add:

| Test | Type per TESTING.md taxonomy | Status |
|---|---|---|
| `npc_ai_fight_picks_top_threat_among_multiple_live_targets` | Unit (state machine) | Currently a happy-path test; needs blocker (1) fix |
| `npc_ai_fight_nan_threat_does_not_panic` | Unit (boundary-value) | Currently doesn't exercise the named branch; needs blocker (2) fix |
| `npc_ai_leash_emits_stat_update_then_state_field_to_witnesses` | **Wire-format** (presence + ordering on `WitnessEntityMethod`) | Solid as-is; optional byte-exact extension |

The leash test is a good example of the multi-type principle: it's labeled `_emits_stat_update_then_state_field_to_witnesses` (entity-state ordering) but it's also asserting wire packets. The naming captures both aspects.

### Test inventory

3 tests added → [docs/testing/inventory/services.md](docs/testing/inventory/services.md) needs three new rows (under the `cell/service/tests/npc_ai.rs` section). Currently missing.

Issue #205, opened 2026-05-04
Test coverage residuals from #123 (Group A leftovers)


Cadacious:
Residual work from #123 Group A. The 22 PRs that closed #123 substantively covered every named module; this issue tracks the files that genuinely still have zero tests, plus the ones that were explicitly deferred for fixture-cost reasons.

### `crates/services/`
- [ ] **`cell/service/ticks.rs`** — ~250 lines, 0 tests. Periodic tick loop driving AoI, NPC AI, ability cooldowns.
- [ ] **`cell/service/startup.rs`** — ~280 lines, 0 tests. Service boot sequence: spawn loaders, spaces, navmesh.
- [ ] **`cell/service/base_messages.rs`** — ~328 lines, 0 tests. Inbound `BaseToCellMsg` dispatch (CreateEntity, GrantItem, MailDelivered, InitPlayerState, etc.). The InitPlayerState restore path is partially exercised by #196's mission round-trip but the dispatch arms otherwise have no direct tests.
- [ ] **`base/world_entry/cell_dispatch.rs`** — 436 lines, 0 tests. Routes `CellToBaseMsg` to per-handler functions. Several arms (mission, gate_travel, mail) are exercised end-to-end through #195/#196/#198, but the dispatch table itself isn't pinned.
- [ ] **`base/world_entry/teleport.rs`** — 0 tests.
- [ ] **`base/world_entry/enable_entities.rs`** — 0 tests. Drives the create-player + viewport + cell + forced-position bundle after RESET_ENTITIES.
- [ ] **`base/world_entry/map_loaded.rs`** — 0 tests. Resolves pending world-entry state into the post-load packet sequence.
- [ ] **`base/world_entry/methods/world_entry_db.rs`** — 199 lines. Partial coverage via #196's mission-restore round-trip and #197's no-DB CombatSim path; the DB-success arms (player row found, world_stargates lookup, query_player_load_data) aren't directly pinned.
- [ ] **`base/resources.rs`** — 0 tests.

### `crates/content-engine/`
- [ ] **Chain-replay harness.** Drive a chain end-to-end against a mock context, assert action sequence + final state. Existing chain-3026 regression guard in `crates/services/src/cell/content/chain_replay_tests.rs` (#173) is per-chain; the **generic harness** for any chain is still TODO.

## Deferred for fixture-cost reasons (revisit if hot)

[...]

PR #230, opened 2026-05-06
test(services): add coverage to ticks/base_messages/world_entry residuals (#205)

BangRocket:
## Summary

Closes the "Open files (no fixture-cost objection — just hadn't been hit yet)" bullets from #205 — the residuals from the original #123 group A test burndown.

## Files covered

| File | Before | After |
|---|---|---|
| `cell/service/ticks.rs` | 0 tests | aoi-empty / nav-path advance / nav-path empty / reload empty-slot |
| `cell/service/base_messages.rs` | 0 tests | BaseToCellMsg dispatch arm coverage |
| `cell/service/startup.rs` | 0 tests | service boot sequence sanity |
| `base/world_entry/cell_dispatch.rs` | 0 tests | dispatch table arm coverage |
| `base/world_entry/teleport.rs` | 0 tests | happy + adversarial path |
| `base/world_entry/enable_entities.rs` | 0 tests | happy + adversarial path |
| `base/world_entry/map_loaded.rs` | 0 tests | happy + adversarial path |
| `base/world_entry/methods/world_entry_db.rs` | partial | DB-success arms (player row, world_stargates) |
| `base/resources.rs` | 0 tests | inventory slot helper coverage |
| `docs/testing/inventory/*.md` | stale | services test catalogue + totals updated |

Each file gets at least one happy-path + one adversarial/failure path per #205's acceptance criteria.

## Review update

Addressed PR review feedback in `8f4f885`:

- restored the load-bearing `bag_min_slot` invariant context and switched resource tests to shared inventory constants
- replaced the empty `cell_dispatch` test module with dispatch tests
- tightened `enable_entities`, `teleport`, `map_loaded`, and NPC movement assertions to check observable behavior
- added timeout guards around the `world_entry_db` channel/spawn round trip
- removed issue-number source comments, unused imports, and the tautological sentinel test
- updated `docs/testing/inventory/services.md` and the inventory README totals

## Out of scope (followup PRs)

Per #205's "Deferred for fixture-cost reasons" bucket:

- `crates/services/minigame/server.rs` — TCP framing harness
- `crates/services/base/cooked_data.rs` — full character cooked-data shape
- `crates/services/base/dispatch.rs` — UDP routing harness

These are now also added to the codecov `ignore:` list in #207 (lands first), so the workspace coverage % stays honest while we decide whether the fixture cost is worth it.

The **chain-replay harness** (#205's other bucket) is its own design problem and ships separately.

## Test plan

- [x] `cargo fmt --check`
- [ ] `cargo test -p cimmeria-services --lib` — blocked locally before services compile: missing `external/recast/Detour/Source/DetourAssert.cpp`, required by `crates/entity/build.rs`
- [ ] `cargo clippy -p cimmeria-services --all-targets -- -D warnings` — blocked by the same missing Recast source
- [ ] Codecov per-component `services` floor (62% per #207) holds after these tests land

## Related

- #205 — services + cell/service residuals from #123 group A
- #123 — original test-coverage expansion issue (closed)
- #207 — codecov ratchet (lands first; this PR's per-component services delta should be visible against the new baseline)
- #149, #166, #167, #190, #191 — sibling test-coverage PRs

## Summary by CodeRabbit

* **Tests**
  * Added many new unit tests across base and cell services to exercise error paths and message handling, improving reliability.
* **Documentation**
  * Updated test inventory docs and indexes with refreshed counts and last-updated dates.
* **Chores**
  * Added a reusable test fixture to simplify and standardize unit test setup.


Cadacious (commented):
## PR #230 — `test(services): add coverage to ticks/base_messages/world_entry residuals (#205)`

### Top-level review comment

Big PR, +549/-20 across 9 files. The shape is right — happy + adversarial path per file as #205's acceptance asks. Several individual tests need tightening. I'll go file by file.

The headline issue is **assertion looseness** in the early-return / no-panic tests. TESTING.md is explicit that "no panic" alone is rarely a regression guard — it passes for any refactor that returns `Ok` without doing anything. Most of these tests are one or two `assert!` calls beyond `is_ok()`/`is_err()` from being good guards; calling them out so they can land as guards instead of warm bodies.

### `base/resources.rs` — load-bearing comment lost

#### Blocker: existing doc-comment removed without preservation

[resources.rs:181-198 (old) → 181-187 (new)](crates/services/src/base/resources.rs#L181):

The old `bag_min_slot_is_non_negative_for_every_container` had a 16-line doc-comment explaining *why* the invariant matters — the move handler's three-step swap parks rows at `slot_id = -1` mid-transaction, and this test guards the predicate that keeps the sentinel unreachable from `grant`/`purchase` paths. That comment was the canonical regression context; a future contributor reading this test needs that context.

The new version replaces it with `/// bag_min_slot sentinel invariant (already tested above) — included here for completeness so the test module exercises both helpers.`

Two problems:
- **"already tested above" is wrong** — this *is* the test. There is no "above."
- **The "why" is gone.** Per copilot-instructions.md "Comments — default to none. Add a comment only when the *why* is non-obvious: a hidden constraint, a subtle invariant, a workaround for a specific bug." This is the textbook case to *keep*.

Restore the original doc-comment. The new tests below it are fine; just don't replace the load-bearing comment.

#### Suggestion: `bag_max_slots_known_containers_match_constants` should reference symbolic constants

[resources.rs:199-207](crates/services/src/base/resources.rs#L199):

```rust
assert_eq!(bag_max_slots(1), 40); // Main
```

Per TESTING.md "Don't trust seed data — assert by relationship": these are pure-function constants in code, not seed data, so the literal-vs-relationship rule doesn't strictly apply. But there are existing constants at the top of `resources.rs` (`INV_MAIN`, `INV_MISSION`, etc.) — preferring `bag_max_slots(INV_MAIN)` over `bag_max_slots(1)` would catch a renumber regression and make the test self-document.

### `base/world_entry/cell_dispatch.rs` — empty test mod adds no value

#### Suggestion: drop the empty `mod tests`

[cell_dispatch.rs:854-863](crates/services/src/base/world_entry/cell_dispatch.rs#L854):

```rust
#[cfg(test)]
mod tests {
    // Deferred per issue #205: handle_cell_message requires a live UDP socket...
}
```

This is a doc comment in the shape of a test module. It contributes nothing testable. The deferral rationale belongs in the PR body or in #205, not in source. Two specific issues:

- **Issue number in source comment** violates copilot-instructions.md line 30.
- **Empty `#[cfg(test)] mod tests`** is dead code that codecov has to learn to ignore separately from the `**/*_tests.rs` glob.

Delete the entire `mod tests` block. If you want a forwarding pointer, a single line in the file's module-doc comment ("Direct unit-tests deferred per #205; arms exercised end-to-end by chain-replay harness") is enough — and even that is borderline.

### `base/world_entry/enable_entities.rs` — happy-path tautology shape

#### Tighten: both tests assert only `is_ok()`

[enable_entities.rs:117-145, 147-175](crates/services/src/base/world_entry/enable_entities.rs):

`enable_entities_returns_ok_when_addr_not_in_connected_map` — the function may always return `Ok` for any input. Per TESTING.md / PR #143 review: "Negative-path tests must assert what's actually missing." To pin the early-return semantics, assert side-effects:

- `entity_to_addr` map unchanged (length + keys).
- No UDP packet sent (drain `socket.try_recv_from` and assert `WouldBlock`).
- No mutation of the `EntityManager`.

Same shape applies to `enable_entities_returns_ok_when_char_list_already_sent`. Without those, a regression that wired up wrong-address packets would pass.

### `base/world_entry/map_loaded.rs` — error-variant not pinned

#### Tighten: `map_loaded_errors_when_no_pending_entry` asserts only `is_err()`

[map_loaded.rs:223-225](crates/services/src/base/world_entry/map_loaded.rs#L223):

A regression that returned a different error variant (e.g., `WrongState` instead of `NoPendingEntry`) would still pass. If the function's error type is an enum, match on the variant. If it's a `String`/`anyhow::Error`, assert `to_string().contains("pending_map_loaded")` or similar.

### `base/world_entry/methods/world_entry_db.rs` — multiple issues

#### Blocker: `query_world_entry_no_db_with_cell_tx_round_trips_create_entity` lacks a timeout

[world_entry_db.rs:251-269](crates/services/src/base/world_entry/methods/world_entry_db.rs#L251):

```rust
let handle = tokio::spawn(async move { ... });
if let Some(msg) = cell_rx.recv().await { ... }
let entry = handle.await.unwrap();
```

A regression that doesn't send `BaseToCellMsg::CreateEntity` would wedge `cell_rx.recv().await` indefinitely. Per TESTING.md "Concurrency": "Wrap each `JoinHandle::await` in `tokio::time::timeout(...)`. A deadlock should fail the test, not wedge the suite."

Wrap both `cell_rx.recv()` and `handle.await` in `tokio::time::timeout(Duration::from_secs(2), …)`. The bare `.recv().await` is the wedge candidate.

#### Suggestion: `no_entity_id_sentinel_is_zero` is a tautology

[world_entry_db.rs:282-287](crates/services/src/base/world_entry/methods/world_entry_db.rs#L282):

`assert_eq!(NO_ENTITY_ID, 0)`. This pins a constant value but not its *use*. Any refactor that left `NO_ENTITY_ID` as `0` while changing the call sites that compare against it would pass. If the contract is "functions returning sentinel return zero," a test that drives one of those functions and asserts `result == NO_ENTITY_ID` (or `!= NO_ENTITY_ID`) carries more weight.

Either drop this test or convert it into `query_world_entry_no_db_returns_real_entity_id_not_sentinel` (asserting `!= NO_ENTITY_ID`), which is closer to what `query_world_entry_no_db_allocates_entity_and_returns_default` is already doing — at which point the sentinel test is redundant.

### `base/world_entry/teleport.rs` — no observable assertions

#### Tighten: both tests have no assertions at all

[teleport.rs:305-326, 328-353](crates/services/src/base/world_entry/teleport.rs):

```rust
#[tokio::test]
async fn teleport_early_returns_when_entity_not_in_addr_map() {
    // ... setup ...
    handle_teleport_player(...).await;
    // (function returns ())
}
```

These pass if the function panics → no, panics fail. Tests pass if the function returns at all. So they pin "no panic" — which TESTING.md flags as a weak shape unless coupled with a side-effect assertion.

What you can assert:
- `entity_to_addr` map is unchanged after the call.
- `connected` map is unchanged.
- No bytes written to `socket` (drain via `try_recv_from` + assert `WouldBlock`).

Without these, a regression that *did* something on the early-return path would silently pass.

### `cell/service/base_messages.rs` — mostly solid, two notes

#### Note: `destroy_entity_flushes_dirty_bandolier_before_destroy` doesn't pin ordering

[base_messages.rs:371-423](crates/services/src/cell/service/base_messages.rs#L371):

The test name promises "before destroy," but the assertions check both *happened* — not their order. Drained messages are inspected after the call; if `handle_base_message` somehow destroyed the entity, then realized it was dirty and emitted the flush message *after*, the test would still pass (entity gone, flush message present in `rx`).

Either:
- Rename to `destroy_entity_flushes_dirty_bandolier_and_destroys_entity` (no ordering claim), or
- Inspect `mgr` for the entity *before* draining `rx` to prove the flush wasn't post-hoc — but with how the channel works that's hard.

The first option is honest about what the test pins.

#### `entity_move_updates_position_in_space_manager` — happy-path only

[base_messages.rs:425-456](crates/services/src/cell/service/base_messages.rs#L425):

Doesn't pin:
- Velocity update — the message carries `velocity: [1.0, 2.0, 3.0]` but the test never checks `entity.velocity`.
- Out-of-bounds position handling (Min/MaxX clamping or rejection).
- Unknown entity_id — graceful no-op vs panic.

The happy-path test is fine to land. Track the missing branches under #205 follow-up rather than blocking this PR.

### `cell/service/startup.rs` — issue number in comment

#### Fix: drop `(issue #205 deferred bucket: startup.rs)` from the doc-comment

[startup.rs:518-520](crates/services/src/cell/service/startup.rs#L518):

```rust
/// CellService::default() must produce a non-running service with no
/// loop handle. This is the minimal surface we can exercise without
/// real XML files / DB pools — larger startup tests are deferred to
/// the integration harness (issue #205 deferred bucket: startup.rs).
```

The "(issue #205 deferred bucket: startup.rs)" violates the no-PR/issue-numbers-in-source rule (copilot-instructions.md line 30). The rationale stays — just drop the parenthetical:

```rust
/// ... larger startup tests are deferred to the integration harness.
```

### `cell/service/ticks.rs` — one loose assertion, one solid set

#### Tighten: `npc_movement_tick_advances_along_nav_path` asserts looser than the math allows

[ticks.rs:584-592](crates/services/src/cell/service/ticks.rs#L584):

```rust
assert!(npc.position.x > 0.0, ...);
assert!(npc.nav_path.len() <= 1, ...);
```

Per TESTING.md / PR #150 review: "Assert exact final positions … not 'two distinct non-sentinel positions' — the loose form passes when both moves rolled back."

`move_speed = 5.0`, waypoint at `(10, 0, 0)`, one tick at the standard delta should produce a deterministic position. Replace with the exact value (or use `assert_relative_eq!` if there's float drift). Same for `nav_path.len()` — should be exactly 0 if the waypoint was reached, exactly 1 if not. `<= 1` is the loose form.

Compute the expected position from `move_speed * tick_delta_secs`, hard-code it as a `const EXPECTED_X: f32 = ...` with a comment showing the derivation, and assert `position.x` matches exactly.

#### Solid: `reload_completion_tick_skips_entity_with_empty_slot`

[ticks.rs:617-648](crates/services/src/cell/service/ticks.rs#L617):

Two tight assertions (no messages drained, deadline cleared). This is the shape the file's other tests should match. Good.

### Test-type matrix (across the whole PR)

| Test type | Files covered | Files not covered |
|---|---|---|
| **Unit** | resources, ticks (npc_movement), startup, map_loaded, teleport | — |
| **Wire-format** | (none, but the cell_dispatch deferral notes the omission) | A `BaseToCellMsg` byte-exact dispatch test is the natural next step; not in this PR's scope. |
| **Live-DB regression guard** | (none added — none of these handlers do their own SQL) | OK as-is. |
| **Concurrency guard** | (none) | `query_world_entry_no_db_with_cell_tx_round_trips_create_entity` is concurrency-shaped (spawn + channel) — needs `tokio::time::timeout` per blocker above. |
| **Smoke** | — | Out of scope. |
| **Chain-replay** | — | The chain-replay harness is its own follow-up per #205, acknowledged in the PR body. |

The PR is a good-faith effort to tile the residuals. The lossiness is concentrated in tests that assert `is_ok()`/`is_err()` without observable side-effects — that's the recurring tightening note.

### Test inventory drift

This PR adds approximately 14 tests and touches no inventory file. Per CLAUDE.md / copilot-instructions.md line 47, [docs/testing/inventory/services.md](docs/testing/inventory/services.md) needs ~14 rows added (one per new `#[test]` / `#[tokio::test]`) plus updated totals in [docs/testing/inventory/README.md](docs/testing/inventory/README.md). This is a review-blocker per the linked instructions.

Recommend running the documentation-writer agent against the PR's diff once the test-tightening above lands — its inventory section is exactly the use case the agent's persona file describes ("test-inventory specific" sub-section).

## Cross-cutting observations across all six PRs

A few patterns worth surfacing as they recur:

1. **Test-inventory drift is the most common review-blocker shape across this batch.** Three of the six PRs (#228, #229, #230) add tests without updating [docs/testing/inventory/services.md](docs/testing/inventory/services.md). Per the doc-update map in CLAUDE.md and the explicit review-blocker callout in copilot-instructions.md line 47, this is required in the same PR. Worth flagging in the next sprint kickoff so contributors know to update it pre-push.
2. **Issue numbers in source comments** appear in #226 (`issue #149`), #230 (`issue #205` in three places). The rule is in both copilot-instructions.md and TESTING.md "Comment hygiene"; might be worth a one-line reminder in `.claude/agents/documentation-writer.md`'s "Behavioral rules" — that agent is the natural enforcement point.
3. **Loose assertions on early-return paths** appear in #228 (`is_ok()`-only) and #230 (multiple `is_ok()`/`is_err()`-only tests, plus zero-assertion tests in `teleport.rs`). TESTING.md PR #143 already calls this out; consider promoting it from "common gotcha" to a top-level checklist item in TESTING.md "Adding a new test — checklist".
4. **The test-types-per-feature principle** (the user's note) applies most directly to #227 (one feature, one fix, but a live-DB regression guard is the right shape and is missing) and #229 (the leash test naturally spans unit + wire-format, and that's why it's the strongest test in this batch). Worth a TESTING.md note: "When in doubt, ask — does the bug shape live in code logic, in bytes, in SQL, in a race, or across handlers? Pick the test type that matches where the bug *lives*, not the layer you happen to be editing."

Issue #236, opened 2026-05-07
Wire mission 687 (Aftermath): crate interaction, archetype-gated rewards, barracks guards


Cadacious:
## Summary

Mission 687 (Aftermath) — the final mission in the Castle_CellBlock escape sequence — is unwired in the content engine. The player reaches it after completing 681-686 but cannot progress: the wooden crate has no interactivity, no barracks guards exist, and no completion path is defined.

## Reproduction

1. Play through the Castle_CellBlock escape from start (mission 622) through the Hallway05 controller (mission 686).
2. Mission 687 (Aftermath) auto-accepts on entering Region6 (chain 1084).
3. Walk to the wooden crate at world position `(-130.45, 24.67, -92.07)` in Castle_CellBlock — the room in front of the door to the barracks (downstairs from Hallway05).
4. Right-click the crate → no Search interaction available.
5. Mission step 2354 ("Search the crate for any useful items.") cannot advance.

Observed in [content.log](logs/content.log) at the end of the mission 686 completion path: no `interact_tag Cellblock_WoodenCrate` chain matches.

## Reference: original Python script

`python/cell/missions/Castle_CellBlock/Aftermath.py` is the design reference. It:

1. Subscribes to `entity.interact.tag::Cellblock_WoodenCrate`.
2. On interact + step 2354 active:
   - Branches on `player.archetype`:
     - `archetype < 5` (Tau'ri / Human): displays dialog 3942, grants the **Covert Stealth set** + a **Combat Knife**, advances to step 2355.
     - `archetype == 8` (Jaffa): displays dialog 3943, grants an **Armored Prison Jacket** + a **Serpent Staff**, advances to step 2355.
3. Step 2355 ("Eliminate the guards in the barracks.") then expects a series of `Barracks_Guard*` kills to complete mission 687.

Item ids resolved against `resources.items`:

| Item ID | Name | Branch |
|---|---|---|
| 3347 | Covert Stealth Helmet | Tau'ri |
| 3359 | Covert Stealth Vest | Tau'ri |
| 3372 | Covert Stealth Pants | Tau'ri |
| 3387 | Covert Stealth Gloves | Tau'ri |
| 3401 | Covert Stealth Boots | Tau'ri |
| 3325 | Combat Knife | Tau'ri |
| 3482 | Armored Prison Jacket | Jaffa |
| 2797 | Serpent Staff | Jaffa |

### 1. Crate is not interactable

[...]

Issue #216, opened 2026-05-05
Prisoner 329 dialog: human players see Jaffa branch — chains 1011/1012 have dialog set IDs swapped


Cadacious:
## Summary

Talking to Prisoner 329 in the Castle Cellblock as an Op-Core (Tau'ri human) player presents the **Jaffa** dialog branch — the prisoner addresses the player as a Jaffa with a symbiote, references Goa'uld lore, etc. Confirmed in playtesting on 2026-05-05. The race-detection logic in chains 1011/1012 has the dialog set IDs swapped relative to the archetype check.

## Root cause

Dialog content (verified via `resources.dialog_screens`):

| Dialog | First screen text (excerpt) | Race |
|---|---|---|
| 2299 | "You're a bit brighter than most humans, aren't you?" | **Human** |
| 2300 | "You feel uncomfortable, don't you? Nauseous? … It's not some [stasis sickness]" | **Human** |
| 5020 | "What about my symbiote? … I have no idea what the NID humans did with that." | **Jaffa** |
| 5021 | "My symbiote will cure me. I will not trust a Goa'uld." | **Jaffa** |

Pattern matches every other content chain in this seed: `5xxx`-prefixed dialog ids are Jaffa branches, `2xxx`/`3xxx`-prefixed are Human branches. (Chain 1056/1057 in the same seed already gates `dialog 5023` on `archetype eq 8` and `dialog 3999` on `archetype neq 8` — that one's correct.)

The Prisoner 329 chains in [`db/resources/Content/Seed/castle_cellblock_chains.sql:74-108`](https://github.com/SandboxServers/Cimmeria/blob/main/db/resources/Content/Seed/castle_cellblock_chains.sql#L74-L108) do the opposite:

```sql
-- Chain 1011: enter Region2 when 638 not active + non-Jaffa → accept + human dialog set
-- dialog_set 5866 maps to dialog 5021 (Human version of "Free Prisoner 329")
INSERT INTO content_conditions (...)
VALUES
  (1011, 'mission_status', 638, NULL, 'eq', 'not_active', 0),
  (1011, 'archetype', NULL, NULL, 'neq', '8', 1);   -- non-Jaffa

INSERT INTO content_actions (...)
VALUES
  (1011, 'accept_mission', 638, NULL, '{}', 0, 0),
  (1011, 'add_dialog_set', 5866, NULL, '{"slot": 17}', 0, 1);  -- ← gives Jaffa dialog to humans

[...]

PR #237, opened 2026-05-07
feat(content-engine): runtime counters, multi-trigger chains, Castle_CellBlock fixes

Cadacious:
## Summary

Two interleaved threads land here:

1. **Content engine layer** gets runtime counter state, multi-trigger chains, the `StatBelowMax` condition, and `instance_id` plumbing through `OnItemUse`. These are the engine primitives the chain seed needs in order to express "kill N tagged enemies", "gate consumable use on stat headroom", and "consume the exact item the player clicked."

2. **Castle_CellBlock content fixes** — the Lockdown→Aftermath chunk of the cellblock was broken in several ways: mission 680 never completed, missions 681-686 had an off-by-one mission→tag mapping plus a counter system that didn't actually count, two region tags had case-sensitivity typos (one of which silently soft-stuck the player at "Lockdown!"), and the Prisoner 329 Human-archetype dialog post-Livewire was misrouted.

Empirically tested in-game through to the end of mission 686. Mission 687 (Aftermath) is the next gap — captured in #236.

### Runtime counters

`CellEntity` now carries `counters: HashMap<String, i32>`. `Action::IncrementCounter` and `Action::ResetCounter` actually mutate it (they were stubs that only logged). `populate_mission_context` writes `counter_<name>` numeric params into the chain context so `Condition::Counter` reads real values.

**Resolve/execute ordering note** (load-bearing for chain authoring): chains resolve before any actions execute, so a sibling completion chain reading the counter on the same trigger event sees the PRE-increment value. For "kill N targets to complete," the completion condition uses `counter >= target - 1` so it fires on the kill that brings the counter to N. Documented at every relevant chain in the seed.

### Multi-trigger chain support

The loader previously took the first `content_triggers` row per chain and silently dropped the rest. Now it materializes one in-memory `Chain` per trigger row, all sharing the same conditions and actions — so a chain that declares "fire on either MessHall_Guard1 OR MessHall_Guard2 death" actually fires on both. The runtime resolver doesn't need to know about the multi-trigger origin; it sees N independent chains keyed by each trigger's `TriggerType`.

### `Condition::StatBelowMax`

New condition variant. True iff `stat_<id>_cur < stat_<id>_max`. `populate_stats_context` populates the underlying params from the entity's `StatList`. `fire_item_use` now invokes both populators so chain 4001 (Health Slappack) can gate on `stat_below_max 7` — using a slappack at full HP now fizzles silently instead of burning the stack. Fail-closed when the params are missing (treats as "no headroom" so a wiring mistake can't make consumables free at full stat).

A user-facing "already at full health" message is intentionally not emitted here — `Action::SystemMessage` is currently a stub (correct client wire format unknown; see the comment at `executor.rs::SystemMessage`). Until that's wired, the failure mode is "click does nothing, stack preserved."

### `ItemUsed` instance_id plumbing

`BaseToCellMsg::ItemUsed` and `CellOutboxPayload::ItemUsed` gain an `instance_id` field carrying the inventory row id the player clicked. `fire_item_use` writes it into the chain context. `Action::RemoveItem` reads it from `ResolvedActions.params` and routes through `CellToBaseMsg::RemoveInventoryItem` (by-instance) when present, falling back to `RemoveInventoryItemByType` (by-design, current behavior) when absent. This is the difference between "consume the slappack you clicked" and "consume the leftmost slappack of that type" — the bug class shows up when a player has multiple stacks of the same item and clicks anything other than the first one.

`ResolvedActions` gains a `params: HashMap<String, Value>` field carrying forward the resolution-time `ExecutionContext.params` so action executors can read trigger-time state without holding the original context.

### Mission 638 — Prisoner 329 Human dialog (#216 follow-up)

Chain 1021 (Human post-Livewire path) was displaying dialog **5020** — the Jaffa "my symbiote will cure me" follow-up. Now displays **2299**, the Human "Agree to escape" follow-up per the dialog mapping at the top of the file. PR #216 fixed the topic dialog routing; this is the missed instance for the post-Livewire follow-up dialog.

### Mission 680 — Escape the Cellblock

Mission 680 had no completion chain anywhere in the seed. The player got the "Lockdown! Find another way out of the Castle!" prompt on entering the topside ring room and would have stayed on that step indefinitely while missions 681-687 progressed underneath. Chain 1073 (`enter Region9`) now also `complete_mission 680` alongside `accept_mission 681`.

### Missions 681-686 — Mess Hall + Hallway controllers

Chain set 1085-1094 rewritten end-to-end:

- **Off-by-one mission→tag mapping fixed.** Before this PR:
  - 681 ("Mess Hall Controller") tracked `Hallway01_Guard` (1 spawn — design wanted N=3 kills, max possible was 1).
  - 682 ("Hallway01 Controller") tracked `Hallway02_Guard`.
  - … shifted forward through the chain …
  - 686 ("Hallway05 Controller") tracked `MessHall_Guard` (zero exact-match spawns — `MessHall_Guard1` and `MessHall_Guard2` exist with suffixes; the trigger never matched).
  After: each mission tracks its own room's guards.

- **Thresholds aligned with actual spawn counts.** Hallway01-04 are single-guard rooms; their missions complete on the one kill (no counter needed). Mess Hall and Hallway05 are two-guard rooms; `target_value = 2`, completion condition `counter >= 1`.

- **Missing increment chains added.** 683-686 had completion chains but no increment chains, so even with correct tags the counter would never advance.

- **Two-guard completion chains use OR-triggers.** Now possible thanks to the multi-trigger loader fix above.

### Region tag typos

Caught by a new linter test (`every_chain_file_uses_consistent_region_world_prefix` in `crates/content-engine/tests/interact_tag_linter.rs`). The runtime resolver does literal-string match on event_keys, so a case mismatch silently never fires.

- **`castle_cellblock_chains.sql` chain 1073**: `Castle_CellBlock.Region9` → `Castle_Cellblock.Region9`. This was blocking mission 680→681 progression entirely (the trigger never matched the canonical `point_sets.name`).
- **`space_castle_cellblock_chains.sql` chains 5002, 5003**: `Castle_CellBlock.Region8` → `Castle_Cellblock.Region8`. Same bug class.

## Tests

- All 593 lib tests on `cimmeria-services` pass.
- 4 tests on `cimmeria-content-engine` linter pass:
  - `every_interact_tag_chain_has_set_interaction_type` (existing).
  - `every_chain_file_uses_consistent_region_world_prefix` (new — pinned the case-sensitivity invariant after catching three real typos).
  - `scan_chains_picks_up_basic_pattern` and `scan_region_triggers_extracts_chain_and_region` (parser smoke tests for the linters).
- `cargo clippy --all-targets -- -D warnings` clean on `cimmeria-services`, `cimmeria-content-engine`, and `cimmeria-entity`.
- Empirically tested in-game from mission 622 through mission 686 completion. Mission 687 is the next gap — see #236.

## Known follow-ups

- **#236 — Wire mission 687 (Aftermath):** crate interaction, archetype-gated rewards (Tau'ri vs Jaffa branches per `python/cell/missions/Castle_CellBlock/Aftermath.py`), barracks guard spawns, completion chains. Reuses the counter system this PR adds.
- **`Action::SystemMessage` is still a stub** — the slappack full-HP gate works but is silent. A user-facing "already at full health" message needs the correct client wire format reverse-engineered first.

## Test plan

- [x] `cargo fmt --all -- --check` clean on touched crates
- [x] `cargo clippy -p cimmeria-services -p cimmeria-content-engine -p cimmeria-entity --all-targets -- -D warnings` clean
- [x] `cargo test -p cimmeria-services -p cimmeria-content-engine -p cimmeria-entity --lib` — 593 cimmeria-services + content-engine linter tests pass
- [x] **In-game** — played from start of cellblock through mission 686 completion. Each mission accepts, advances, and completes as expected. Slappack full-HP gate confirmed (click does nothing, stack preserved). Multi-stack slappack instance routing confirmed (clicked stacks consume the right instance).

## Summary by CodeRabbit

* **New Features**
  * Added health-conditional consumption for items: Health Slappack now only usable when current HP is below maximum.
  * Introduced counter-based progression system for mission sub-missions, enabling precise completion tracking.

* **Bug Fixes**
  * Fixed dialog routing for post-Livewire Human player follow-up in Castle CellBlock mission.
  * Corrected region trigger capitalization to ensure proper event matching.
  * Resolved off-by-one errors in guard counter thresholds for hallway sub-missions.

### DIFF
--- a/crates/content-engine/src/chain.rs
+++ b/crates/content-engine/src/chain.rs
@@ -225,8 +225,16 @@ impl ChainEngine {
 ///
 /// Each entry pairs a chain ID with the action to execute, preserving the
 /// ordering (highest-priority chain first, actions in declaration order).
+///
+/// `params` carries forward the resolution-time `ExecutionContext.params`
+/// so action executors can read trigger-time state without holding the
+/// original context. This is how `Action::RemoveItem` looks up
+/// `instance_id` set by `fire_item_use` to consume the exact stack the
+/// player clicked instead of the player's first-by-type instance.
+#[derive(Default)]
 pub struct ResolvedActions {
     pub actions: Vec<(i64, Action)>,
+    pub params: std::collections::HashMap<String, serde_json::Value>,
 }
 
 impl ChainEngine {
@@ -239,6 +247,7 @@ impl ChainEngine {
     pub fn resolve_event(&self, event: &TriggerEvent, ctx: &ExecutionContext) -> ResolvedActions {
         let mut resolved = ResolvedActions {
             actions: Vec::new(),
+            params: ctx.params.clone(),
         };
 
         let chains = match self.chains_by_trigger.get(&event.trigger_type) {
--- a/crates/content-engine/src/chain.rs
+++ b/crates/content-engine/src/chain.rs
@@ -245,10 +245,7 @@ impl ChainEngine {
     /// has access to game state (SpaceManager, channels, etc.) that the engine
     /// itself doesn't know about.
     pub fn resolve_event(&self, event: &TriggerEvent, ctx: &ExecutionContext) -> ResolvedActions {
-        let mut resolved = ResolvedActions {
-            actions: Vec::new(),
-            params: ctx.params.clone(),
-        };
+        let mut resolved = ResolvedActions::default();
 
         let chains = match self.chains_by_trigger.get(&event.trigger_type) {
             Some(chains) => chains,
@@ -270,6 +267,15 @@ impl ChainEngine {
             for action in &chain.actions {
                 resolved.actions.push((chain.id, action.clone()));
             }
+        }
+
+        // Defer the params clone until at least one chain matched.
+        // `resolve_event` runs on every gameplay tick that produces an
+        // event (entity death, region cross, item use, …), most of
+        // which return zero actions; cloning the populated context
+        // unconditionally wasted bytes on every miss.
+        if !resolved.actions.is_empty() {
+            resolved.params = ctx.params.clone();
         }
 
         resolved
--- a/crates/content-engine/src/conditions.rs
+++ b/crates/content-engine/src/conditions.rs
@@ -80,6 +80,18 @@ pub enum Condition {
         operator: ComparisonOp,
         value: i32,
     },
+
+    /// True iff the source entity's stat is below its current max
+    /// (i.e., the stat has headroom to grow). Used to gate consumable
+    /// chains so e.g. Health Slappacks fizzle silently rather than
+    /// burning a stack when the player is already at full HP.
+    ///
+    /// Reads `stat_<id>_cur` and `stat_<id>_max` from the context;
+    /// callers must populate these via `populate_stats_context` before
+    /// resolving. Returns `false` (treat as "no headroom, don't fire")
+    /// if either param is missing — fail-closed so a wiring mistake
+    /// can't accidentally make consumables free-to-spam at full stat.
+    StatBelowMax { stat_id: i32 },
 }
 
 /// Faction relationship levels.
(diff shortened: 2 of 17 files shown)


Cadacious:
@coderabbitai review

PR #309, opened 2026-05-17
feat(mercury): server-side retransmit for reliable packets — close the ACK loop (#308)

Cadacious:
## What's going on, in plain English

One of our teammates plays from Germany. He's been hitting bugs nobody on the LAN ever sees — quest updates that never arrive, NPCs that don't appear, world objects that just don't react when he interacts with them. We finally tracked it down: the server has been **sending** important messages reliably, but when the network drops one (which happens routinely on a transatlantic connection — somewhere between 0.5% and 2% of packets), the server never noticed and never re-sent it. The message was gone for good, and whatever it was supposed to trigger silently failed.

The reason is small and slightly embarrassing: the client has been politely telling us "I got your message, packet #42 made it" all along. The server was reading those acknowledgements, writing them to a debug log, and **throwing them away.** No tracking, no timer, no retry — just trust that the first send worked and hope for the best. That's fine on a local network where loss is effectively zero. On any real internet connection it means a single dropped packet permanently breaks a quest update, an NPC spawn, an inventory grant, a vendor purchase confirmation, whatever was in transit.

This PR closes the loop. The server now keeps a per-player record of every reliable message it sends, watches the client's acknowledgements come back, measures how long each round-trip takes, and re-sends anything the client hasn't confirmed within an adaptive timeout. The timeout is per-player — fast for LAN players, generous for long-distance ones — so we don't waste bandwidth on spurious retries for someone on a slow link, and we don't make LAN players wait around either.

In concrete terms: after this lands, if our German teammate logs in and one of his quest-update packets drops, the server notices within roughly 350-550 ms and re-sends it. He gets the update. The bug class goes away. Internal server logs will also start showing each player's real round-trip time, so we can see who's playing from where without any extra instrumentation.

What's **not** in this PR: this only covers messages the server sends to the client. The reverse direction (reassembly of incoming fragments) has a separate bug (issue #290) that needs its own PR. Login handshake retries are also not in scope — those have their own client-side retry path.

---

## Engineering detail

Issue: #308. Closes the gap where reliable server→client packets are marked `FLAG_RELIABLE`, client dutifully ACKs them, and the server discards the ACKs at `crates/services/src/base/connect_loop/encrypted.rs:69`. Symptom space (NPC spawns, interaction triggers, content-engine packets, quest state) matches exactly what a transatlantic teammate at 150-400 ms RTT and ~1% baseline UDP loss has been reporting.

Seven commits, layered intentionally so each is reviewable in isolation:

[...]

Issue #292, opened 2026-05-16
[Major] Reliability layer corrections (tracking) — TX window 45->32, per-tick budget 5.0, 28-bit sequence space


Cadacious:
## Background

Audit: [docs/audits/mercury-rust-conformance-2026-05-15.md](https://github.com/SandboxServers/Cimmeria/blob/main/docs/audits/mercury-rust-conformance-2026-05-15.md) — findings **#5**, **#6**, **#7** (all Major). Tracking issue for the reliability-layer parity cluster.

All three touch `crates/mercury/src/channel/` and `crates/mercury/src/lib.rs` and are conceptually one cluster: bringing Cimmeria's reliability layer into wire-parity with the spec's §1.7 contract. Splitting across three PRs lets each land independently for risk-isolation; this issue tracks them as a coherent set.

## #5 — Outstanding-ack window: 45 -> 32

**Today:** `crates/mercury/src/lib::consts::TX_WINDOW_SIZE = 45`. `crates/mercury/src/channel::Channel::send` enforces a 45-slot circular TX buffer (rejects the 46th in-flight reliable packet).

**Spec §1.7 + §1.16 Q5 closure:** the SGW client has a 32-bit outstanding-ack bitmap indexed by `seq_id & 0x1F`. Max 32 in-flight reliable per channel. Two packets with seq_ids differing by 32 collide on the same bitmap bit — if either is lost, the server phantom-acks both.

**Change:** drop `TX_WINDOW_SIZE` from 45 to 32. Optionally, replace the `VecDeque` with a 32-bit bitmap to match the spec's structural model (eliminates the slot-count drift class entirely). Add a test asserting the 33rd in-flight reliable enqueue produces back-pressure.

## #6 — Per-tick retry budget: missing 5.0

**Today:** `crates/mercury/src/channel::Channel::check_timeouts` walks the entire unacked-packet list and retransmits every entry past `ACK_TIMEOUT_MS`. No per-tick rate limit.

**Spec §1.7 + §2.4.1 R14 + §2.10 S7:** retry cap is **two distinct numbers**:
- 20 lifetime cap (already implemented as `MAX_RETRIES` with strict `>` at `channel/mod.rs:318`).
- 5.0 IEEE 754 per-tick work budget — `UnAckedHandler::checkResendTimers` at `ghidra://SGW.exe@0x0158c420` processes at most 5 entries from the unacked list before yielding.

[...]

PR #317, opened 2026-05-17
fix(mercury): give unreliable packets their own seq counter; restore tickSync to reliable

Cadacious:
## What was broken

Players got stuck mid-session — character-select for some, mid-world for others. Server-side everything looked fine; the client-side AtreaLoader log showed `UnAckedHandler::queueAckForPacket: Buffering packet #N above #M` firing dozens of times in a row, the receive-side reorder buffer filling up, and eventually the client disconnecting on inactivity.

Symptoms in two reported sessions:

- **Character-select stuck**. Click *Create*, nothing happens in the UI, so the user mashes the button. Server logs show the character actually got created on the first click, every subsequent click coming back as `Character name already taken`.
- **NPC interaction stuck in-world**. Walk up to an NPC and click. Server logs show the `interact` request arrives and `onDialogDisplay` is dispatched back; the dialog never appears on the client.

Both shapes are the same underlying issue at different points in the session lifecycle.

## Root cause

The SGW BigWorld client tracks **one contiguous monotonic seq counter for reliable packets only**. Its receive-side state machine (`UnAckedHandler::queueAckForPacket`) advances its "next expected" pointer by exactly 1 each time the next reliable seq arrives in order — there's no code path that skips the pointer forward past a slot occupied by an unreliable packet. Unreliable packets go through a separate dedup-only path that never touches the reliable pointer.

The Cimmeria server was incrementing a single shared \`next_seq\` counter for **every** outgoing packet, reliable or not. The first time an unreliable packet got emitted between two reliables, the reliable seq stream had a permanent hole the client could never fill, and every subsequent reliable packet (NPC dialogs, mission updates, inventory changes, everything) got buffered above the hole until the connection timed out.

Two visible failure points came from this single mechanism:

1. **`tickSync` at character-select.** PR #309 (server-side retransmit) split the global `REPLY_FLAGS` into reliable/unreliable variants and migrated \`build_ongoing_tick_sync\` to the unreliable variant. As soon as a tick fired between \`createBasePlayer\` and \`onVersionInfo\`, the receive buffer stuck at seq=4 and the character-list UI never refreshed — so clicking *Create* looked like a no-op and users mashed the button.

2. **Avatar position updates in-world.** Once a player entered a world with active AoI traffic, \`build_avatar_update\` (msg \`0x10\`) emits would start streaming out unreliable, and the same shape recurred at whatever seq the first AoI update landed on — breaking the next reliable message, typically an NPC interaction's \`onDialogDisplay\`.

## What this PR changes

Two parts, both small:

1. **Restore \`build_ongoing_tick_sync\` to the reliable flag set.** The bandwidth cost of acking ~10 tick syncs per second is negligible compared to the cost of the bug. The original justification ("retransmit overhead would buy nothing because the next tick supersedes any lost one") was correct in isolation but didn't account for the receiver-side invariant the unreliable-on-shared-counter pattern violated.

2. **Add a second per-session counter \`next_seq_unreliable\` to \`ConnectedClientState\`,** independent from \`next_seq\`. Route the unreliable AoI send helper (\`send_to_witness\`) through it. The receiver has separate dedup state for unreliable packets that doesn't care whether the unreliable seq stream is contiguous with the reliable one, so two independent monotonic counters are wire-format-correct and require no client change.

\`send_to_witness_reliable\` is untouched and still consumes \`next_seq\`.

## How we know it works

- **Local repro before/after.** Pre-fix client log showed 175 \`Buffering above #N\` events in a typical session that hit character-select + light gameplay; post-fix log shows zero.
- **Static analysis of the client binary.** \`UnAckedHandler::queueAckForPacket\` at \`ghidra://SGW.exe@0x0158cba0\` decompiles to a clear advance-by-1-or-buffer state machine — \`MOV [ESI+0x50]\` advances the pointer only when the incoming seq matches; otherwise \`JNZ\` jumps to the buffering branch with the format string \`"Buffering packet #%d above #%d\n"\`. \`FUN_0158bb50\` shows the orthogonal unreliable dedup at struct offset \`+0x128\` that the unreliable path goes through. Both fully documented in \`spec.protocol.mercury-wire-format\` §1.7.
- **Wire capture confirms the new shape.** Post-fix pcap shows reliable seqs staying contiguous on their own counter while unreliable seqs run on a separate stream.

## Test plan

- [x] Build clean on Windows
- [x] Login → character-select → create character → no UI freeze, no button-mashing required
- [x] Pick character → enter world (Castle_CellBlock) → no buffering events in client log
- [x] Interact with NPC → dialog appears
- [ ] Existing unit + integration tests pass (\`cargo nextest run --profile=ci -p cimmeria-services\`)
- [ ] Live-DB tests pass (\`cargo nextest run --profile=ci-live-db -p cimmeria-services --lib\`)

## Notes for reviewers

- The five new \`next_seq_unreliable: Arc::new(AtomicU32::new(0))\` lines in test fixtures and the real \`login.rs\` are mechanical — they just initialize the new field everywhere \`ConnectedClientState\` is constructed.
- The only behavioral changes are at \`crates/services/src/base/helpers.rs\` (\`send_to_witness\` now reads the new counter) and \`crates/services/src/mercury/protocol/session.rs\` (\`build_ongoing_tick_sync\` flips its flag set back to reliable).
- Long-term cleanup remaining: the \`REPLY_FLAGS_UNRELIABLE\` constant is still in the codebase and used by \`build_avatar_update\`. That's correct now because the unreliable counter feeds it. If a future message type wants to be unreliable, the same pattern applies — route through \`send_to_witness\`, the counter is already plumbed.

## Bug Fixes

* Tick-sync messages now use reliable delivery to ensure consistent client-server synchronization
* Improved packet stream handling to prevent reliable message delivery from being disrupted by unreliable packet gaps

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/317?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MyPalClara:

Great work on the root cause analysis, Claude. The fix is correct. But I found three potential correctness concerns worth verifying:

---

### G1 — Both Counters Start at 0

**Location:** `ConnectedClientState` initialization

**The Issue:**
```rust
next_seq: Arc::new(AtomicU32::new(0)),
next_seq_unreliable: Arc::new(AtomicU32::new(0)),
```

Both counters initialize to 0. If the first packet sent is unreliable (seq=0) and the first reliable packet is also seq=0, the client sees two packets with the same sequence number.

**Why It Probably Does Not Matter:**
The client dedup logic for unreliable packets is at struct offset `+0x128`, separate from the reliable `inSeqAt` at `+0x50`. The two seq spaces appear independent from the client perspective.

**Recommended Verification:**
Check `FUN_0158bb50` (the unreliable dedup path) in Ghidra to confirm it never touches `inSeqAt`. If it does, consider initializing `next_seq_unreliable` to `0x08000000` to avoid collision in the first few packets.

**Effort:** 15 min (Ghidra check)

---

### G2 — Flag Set Must Match Counter

**Location:** `send_to_witness` callers

**The Issue:**
The PR shows `send_to_witness` now reads from `next_seq_unreliable`, but I do not see where the packet builder sets `FLAG_RELIABLE = 0`. If the packet builder still sets `FLAG_RELIABLE`, the client will expect it in the reliable stream and the fix is incomplete.

**Recommended Verification:**
Confirm that `REPLY_FLAGS_UNRELIABLE` (or equivalent) is used by all `send_to_witness` callers, and that this flag set does NOT include `FLAG_RELIABLE`.

**Effort:** 10 min (grep + verify)

---

### G3 — tickSync Bandwidth Increase

**Location:** `build_ongoing_tick_sync` now reliable

**The Issue:**
The PR body acknowledges this ("bandwidth cost of acking ~10 tick syncs per second is negligible") but does not quantify it.

**Quick Math:**
- 10 ticks/sec × ~50 bytes/ack × 100 players = ~50 KB/sec additional ACK traffic
- For a transatlantic link, this is negligible
- For a metered connection, it is worth knowing

**Recommendation:**
Not blocking, but consider adding a note to the spec (§1.7) documenting that tickSync is reliable and why.

**Effort:** 5 min (doc update)

---

**Severity: Low-Medium** — G1 and G2 need verification but are probably fine. G3 is informational.

MyPalClara:
## Category 2: Missing Test Coverage (G4-G6)

The fix is surgical and correct, but there are no tests that would catch a regression. Here are the missing test cases:

---

### G4 — No Test for Counter Independence

**What is Missing:**
A test that verifies the reliable and unreliable counters are truly independent.

**Recommended Test:**
```rust
#[test]
fn reliable_and_unreliable_sequences_are_independent() {
    let state = ConnectedClientState::new(/* ... */);
    
    // Send reliable packet — should get seq 0
    let reliable_seq_1 = state.next_seq.fetch_add(1, Ordering::Relaxed) & SEQUENCE_MASK;
    assert_eq!(reliable_seq_1, 0);
    
    // Send unreliable packet — should also get seq 0 (independent counter)
    let unreliable_seq_1 = state.next_seq_unreliable.fetch_add(1, Ordering::Relaxed) & SEQUENCE_MASK;
    assert_eq!(unreliable_seq_1, 0);
    
    // Send another reliable packet — should get seq 1 (not 2)
    let reliable_seq_2 = state.next_seq.fetch_add(1, Ordering::Relaxed) & SEQUENCE_MASK;
    assert_eq!(reliable_seq_2, 1);
    
    // Verify unreliable did not affect reliable
    assert_eq!(state.next_seq.load(Ordering::Relaxed) & SEQUENCE_MASK, 2);
    assert_eq!(state.next_seq_unreliable.load(Ordering::Relaxed) & SEQUENCE_MASK, 1);
}
```

**Effort:** 15 min

---

### G5 — No Test for Seq Collision Avoidance

**What is Missing:**
If G1 is a real issue (both counters starting at 0), there should be a test that verifies the first reliable and first unreliable packets do not collide at the client.

**Recommended Test:**
This would require a mock client or wire-capture validation. Lower priority unless G1 verification reveals a problem.

**Effort:** 30 min (if needed)

---

### G6 — No Regression Test for the Original Bug

**What is Missing:**
The PR body describes the symptom ("Buffering packet #N above #M" 175 times) but there is no automated test that would catch a regression.

**Recommended Test:**
```rust
#[test]
fn unreliable_packets_do_not_create_holes_in_reliable_stream() {
    // Setup: Create a channel with both counters
    let channel = Channel::new(/* ... */);
    
    // Simulate the bug scenario:
    // 1. Send reliable packet (seq 0)
    // 2. Send unreliable packet (should NOT consume seq 1)
    // 3. Send reliable packet (should be seq 1, not seq 2)
    
    let reliable_1 = channel.send_reliable(/* ... */);
    let unreliable_1 = channel.send_unreliable(/* ... */);
    let reliable_2 = channel.send_reliable(/* ... */);
    
    // The reliable stream should be contiguous: 0, 1
    assert_eq!(reliable_1.seq, 0);
    assert_eq!(reliable_2.seq, 1);  // NOT 2
    
    // The unreliable packet should have its own seq
    assert_eq!(unreliable_1.seq, 0);  // Independent counter
}
```

**Why This Matters:**
This bug was severe enough to stall sessions. A regression test ensures it never comes back.

**Effort:** 30 min

---

**Severity: Medium** — G6 is the most important. This bug was Critical; it deserves a regression guard.

MyPalClara:
## Category 3: Documentation Gaps (G7-G8)

The code is well-documented, but the spec and module-level guidance need updates:

---

### G7 — Spec Does Not Document Two-Counter Model

**Location:** `spec.protocol.mercury-wire-format` §1.7

**The Issue:**
The PR references §1.7 but does not update it. The spec should now document that the server maintains two independent sequence counters — one for reliable, one for unreliable — and explain why.

**Recommended Addition to §1.7:**

```markdown
### Server-Side Sequence Counter Model

The server maintains **two independent sequence counters** per client session:

1. **Reliable counter** (`next_seq`) — Used for all packets with `FLAG_RELIABLE` set.
   The client tracks these via `inSeqAt` at `+0x50` and expects a contiguous
   monotonic stream. Gaps cause buffering; the client waits for the missing
   seq before processing subsequent packets.

2. **Unreliable counter** (`next_seq_unreliable`) — Used for packets without
   `FLAG_RELIABLE`. The client deduplicates these via a separate structure at
   `+0x128` but does NOT expect contiguity. Lost unreliable packets are simply
   dropped.

**Critical invariant:** Unreliable packets MUST NOT consume slots in the reliable
sequence space. If they do, the reliable stream will have permanent holes that
the client can never fill, causing session stalls.

See `ghidra://SGW.exe@0x0158cba0` (`UnAckedHandler::queueAckForPacket`) for the
client-side receive state machine.
```

**Effort:** 15 min

---

### G8 — No Module-Level Guidance in helpers.rs

**Location:** `crates/services/src/base/connect_loop/helpers.rs`

**The Issue:**
The individual functions are documented, but there is no module-level guidance explaining when to use which helper.

**Recommended Module Doc:**

```rust
//! # Witness Send Helpers
//!
//! This module provides helpers for sending packets to connected clients.
//!
//! ## Two-Counter Model
//!
//! The server maintains two independent sequence counters per session:
//!
//! - **Reliable** (`send_to_witness_reliable`) — For packets that MUST arrive.
//!   Uses the reliable seq counter, tracked for retransmit, client ACKs expected.
//!   Use for: entity spawns, method calls, property updates, anything stateful.
//!
//! - **Unreliable** (`send_to_witness`) — For packets that can be lost.
//!   Uses a separate seq counter, NOT tracked for retransmit, no ACKs expected.
//!   Use for: position updates, tick syncs (if bandwidth-constrained), ephemeral data.
//!
//! ## When to Use Which
//!
//! | Packet Type | Helper | Why |
//! |-------------|--------|-----|
//! | Entity spawn | `send_to_witness_reliable` | Client state depends on it |
//! | Method call | `send_to_witness_reliable` | Must execute exactly once |
//! | Property update | `send_to_witness_reliable` | Client state depends on it |
//! | Position update | `send_to_witness` | Superseded by next update |
//! | Tick sync | `send_to_witness_reliable` | See PR #317 for why |
//!
//! **Default to reliable.** Only use unreliable if you have a specific reason
//! AND you have verified the client handles loss gracefully.
```

**Effort:** 10 min

---

**Severity: Low** — These are maintainability improvements, not correctness issues. But they prevent the next developer from making the same mistake PR #309 made.

MyPalClara:
## Category 4: Potential Issues (G9) + Summary

One defensive improvement and the overall verdict:

---

### G9 — Unreliable Counter Not Encapsulated

**Location:** `send_to_witness` in `helpers.rs`

**The Issue:**
```rust
let seq = connected.next_seq_unreliable.fetch_add(1, Ordering::Relaxed) & SEQUENCE_MASK;
```

The reliable counter has `drain_acks_and_seq` which encapsulates the fetch_add + mask. The unreliable counter only has this one call site, but if anyone adds a second call site and forgets to mask, you get the same 28-bit overflow bug that Issue #292 documented.

**Recommended Fix:**

Add a helper method to `ConnectedClientState`:

```rust
impl ConnectedClientState {
    /// Returns the next unreliable sequence number, correctly masked to 28 bits.
    ///
    /// Unreliable packets use a separate sequence space from reliable packets.
    /// See PR #317 for why this separation is critical.
    pub fn next_unreliable_seq(&self) -> u32 {
        self.next_seq_unreliable.fetch_add(1, Ordering::Relaxed) & SEQUENCE_MASK
    }
}
```

Then update `send_to_witness`:
```rust
let seq = connected.next_unreliable_seq();
```

**Why This Matters:**
Defensive programming. The next developer who needs to send an unreliable packet will call this helper instead of copy-pasting the fetch_add + mask pattern.

**Effort:** 10 min

---

## Summary Table

| Gap | Category | Severity | Effort | Recommendation |
|-----|----------|----------|--------|----------------|
| G1 | Correctness | Low | 15 min | Verify in Ghidra |
| G2 | Correctness | Low-Medium | 10 min | Grep + verify flags |
| G3 | Correctness | Informational | 5 min | Doc update |
| G4 | Test Coverage | Medium | 15 min | Add test |
| G5 | Test Coverage | Low | 30 min | Only if G1 fails |
| G6 | Test Coverage | **Medium** | 30 min | **Add regression test** |
| G7 | Documentation | Low | 15 min | Update spec §1.7 |
| G8 | Documentation | Low | 10 min | Add module doc |
| G9 | Defensive | Low | 10 min | Add helper method |

**Total effort for all gaps: ~2.5 hours**

---

## The Verdict

**This PR is ready to merge.** 🚀

The fix is correct. The root cause analysis is perfect. The implementation is minimal and surgical.

**Priority for follow-ups:**
1. 🔴 **G6** (regression test) — This bug was Critical; it deserves a guard
2. 🟡 **G7** (spec update) — Prevents the next developer from repeating PR #309 mistake
3. 🟢 **G1, G2** (verification) — Quick checks to confirm assumptions
4. ⚪ **Everything else** — Nice-to-have

**What Claude should be embarrassed about:**
PR #309 introduced this bug. The spec gap (unreliable packets must not consume reliable seq slots) was not documented, and the implementation followed the incomplete spec. This PR fixes the bug correctly, but the spec should be updated to prevent recurrence.

---

**Ship it, then add the regression test. Good work on the fast turnaround.** 🎯

Cadacious:
Pushed `514b5c8b` to address review feedback. Triage and disposition:

| Item | Source | Disposition |
|---|---|---|
| TX window registration for reliable tickSync | Copilot inline | ✅ Fixed in `tick_sync.rs` — `shadow_register_reliable_send` now mirrors each tickSync into the per-session Channel TX window so RTO retransmit covers it. Without this the bug would recur under packet loss. |
| Doc wording "shares with every other server-emitted packet" is now wrong | Copilot inline | ✅ Fixed in `session.rs` — clarified that tickSync rides the reliable stream specifically; called out the unreliable stream's separate counter and the TX-window registration. |
| Regression test for split-counter invariant | CodeRabbit inline + G4 + G6 | ✅ Added `reliable_and_unreliable_seq_counters_are_independent` and `next_unreliable_seq_masks_to_28_bit_space` to `helpers.rs::tests`. R/U/R/U/R interleave catches the exact failure shape this PR exists to prevent. |
| **G1** — both counters start at 0 | MyPalClara | ✅ Verified by existing evidence (no code change). Receiver state at `+0x50` (`inSeqAt`, reliable) and `+0x128` (unreliable dedup) is orthogonal per the `FUN_0158bb50` decompile cited in the PR body — the two state machines never look at each other's seqs, so starting both counters at 0 cannot collide. |
| **G2** — flag set must match counter | MyPalClara | ✅ Verified by existing evidence (no code change). `build_avatar_update` is the only function reachable through `send_to_witness`, and it uses `REPLY_FLAGS_UNRELIABLE` (no `FLAG_RELIABLE` bit). Confirmed via grep. |
| **G3** — tickSync bandwidth note | MyPalClara | Acknowledged. ~50 KB/sec per 100 clients of additional ack traffic is correct for the order-of-magnitude. Folded into the doc-update follow-up below rather than a code change. |
| **G7** — spec §1.7 documenting the two-counter model | MyPalClara | Deferred to a follow-up doc PR — scope is `docs/drafts/spec/mercury-wire-format.md` rather than this fix. Will include the §1.7 addition MyPalClara drafted plus the G3 bandwidth note. |
| **G8** — module-level doc in `helpers.rs` | MyPalClara | ✅ Added the two-counter overview + the "which helper to use" table at the top of `helpers.rs`. |
| **G9** — encapsulating `next_unreliable_seq()` method | MyPalClara | ✅ Added on `ConnectedClientState`. `send_to_witness` now calls it; the raw `fetch_add` + mask pattern is gone from the call site. The new regression test also exercises this accessor. |

Summary: of the 9 reviewer items + 3 bot items, 7 are fixed in this push, 2 are verified-as-already-correct (G1, G2) with no code change, and 2 (G3, G7) are folded into a single spec-doc follow-up PR.

Issue #333, opened 2026-05-18
cleanup(state): BSF_Holster bit 8 does nothing client-side — remove or document the no-op


Cadacious:
## Plain English

We have a `BSF_HOLSTER` constant (bit 8, mask `0x100`) in [`crates/services/src/cell/combat/state.rs`](../blob/main/crates/services/src/cell/combat/state.rs) that the cell exports through [`crates/services/src/cell/combat/mod.rs:14`](../blob/main/crates/services/src/cell/combat/mod.rs#L14). The `use_ability.rs` "clear holster on fire" path at line 222 currently writes to it on every ability launch.

**The 2009 client does not read bit 8 of `bStateField`.** Setting or clearing the bit server-side has zero client-visible effect — no animation switch, no pose change, no UI update. Verified statically via Ghidra: every `bStateField` consumer in the binary tests bits 0–7 only; bits 8–31 have **zero** test sites.

This issue tracks the cleanup: either delete `BSF_HOLSTER` entirely, or keep the constant for server-internal book-keeping but stop pretending it has a wire effect and remove the misleading code comment that claims it does.

## Spec evidence

All evidence from this session's static analysis of `SGW.exe` (image base `0x00400000`).

### State-field shape — verified

`GameBeing` stores `bStateField` (uint32) at offset `+0x158`. Wire property name `bStateField`. Source: `GameBeing_OnStateFieldUpdate @ ghidra://SGW.exe@0x00e01c90` reads the event property via `aEvent->getProperty("bStateField", newStates)` (assert string cites `.\\Src\\GameBeing.cpp:0x341`).

The handler dispatches per changed bit:

| Bit | Mask | Side effect on change |
|---|---|---|
| 0 | 0x01 | `GameBeing_OnDeadStateChanged` + `ApplyDeadInteraction` |
| 1 | 0x02 | `EmitAutoCycleStateChanged` |
| 2 | 0x04 | `UpdateMovementSpeed` (via 0xC4 group) |
| 3 | 0x08 | `UpdateCombatStanceWeaponSet` |
| 4 | 0x10 | `FUN_00e31aa0` (minigame UI lock) |
| 5 | 0x20 | `EmitStealthStateChanged` |
| 6 | 0x40 | `UpdateMovementSpeed` (via 0xC4 group) |
| 7 | 0x80 | `UpdateMovementSpeed` (via 0xC4 group) |

[...]

PR #338, opened 2026-05-18
fix(holster): track via BeingAppearance ComponentList, not the dead BSF_HOLSTER bit

### LINKED ISSUE #249: Player weapon never appears holstered — BSF_HOLSTER never set by any code path
## Summary

Players never appear holstered. Even on a fresh login before any combat — pistol drawn from frame one, stays drawn forever. The server has no code path that *sets* `BSF_HOLSTER`; it only ever clears the bit. Initial `state_field = 0` ([`cell_entity/mod.rs:372`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/entity/src/cell_entity/mod.rs#L372)) means the bit is unset at world entry and never transitions to set.

This isn't a regression from legacy — legacy Python's `requestHolsterWeapon` was literally just a `print` ([`deprecated/python/cell/SGWBeing.py:982-983`](https://github.com/SandboxServers/Cimmeria/blob/main/deprecated/python/cell/SGWBeing.py#L982-L983)). It's a never-implemented system inherited as-is, and the architecture doc at [`docs/architecture/state-flag-conventions.md:62`](https://github.com/SandboxServers/Cimmeria/blob/main/docs/architecture/state-flag-conventions.md#L62) overstates reality when it says `BSF_HOLSTER` is "mostly idempotent player input via `requestHolsterWeapon`" — that handler exists, but in practice nothing tells the client to send the request, and nothing on the server seeds the bit either.

## What's wired up today

| Path | Effect on `BSF_HOLSTER` | Source |
|------|------------------------|--------|
| Player spawn / `CellEntity::new()` | bit = 0 (unset) | [`cell_entity/mod.rs:372`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/entity/src/cell_entity/mod.rs#L372) |
| `requestHolsterWeapon(aHolster)` cell method | sets if `aHolster != 0`, clears otherwise | [`cell_methods/combatant.rs:64-78`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/combatant.rs#L64-L78) |
| `useAbility` (player attack) | clears (raw `&= !`) | [`abilities/use_ability.rs:235`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/abilities/use_ability.rs#L235) |
| `requestReload` | clears (raw `&= !`) | [`cell_methods/player/world.rs:230`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/player/world.rs#L230) |
| Death / respawn | `clear_all_state_flags()` (clears) | [`cell_methods/player/combat/respawn.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/player/combat/respawn.rs) |

So **every code path either initializes-to-clear or actively clears**. The only thing that can set the bit is the client emitting `requestHolsterWeapon(1)` — which (a) we have no evidence the client emits autonomously, and (b) requires the bit to already be settable from a clean state, which it is, but nothing actually triggers it at runtime.

## Client side (from Ghidra trace of SGW.exe)

The client has the full machinery: `Event_NetOut_ChangeWeaponState` is registered ([`event-net-mapping.md:473`](https://github.com/SandboxServers/Cimmeria/blob/main/docs/analysis/event-net-mapping.md#L473) — registration at `0x00d9b120`), which marshalls `requestHolsterWeapon(INT8 aHolster)` — wire format documented at [`combat-wire-formats.md:194-212`](https://github.com/SandboxServers/Cimmeria/blob/main/docs/reverse-engineering/findings/combat-wire-formats.md#L194-L212). The emit chain is `register_NetOut_ChangeWeaponState (0x00d9b120)` → vtable at `0x019cb260` → emit thunk `FUN_00d9b340` → marshaller `FUN_00d9b240` → ctor `FUN_00d9b130`.

What I could not find in the binary: any input-action handler, slash command, ability hook, or idle-tick callback that calls into the emit chain. There is no `Event_SlashCmd_Holster` analog (compared to `Event_SlashCmd_ToggleCombatLOS` which does exist). I can't conclusively rule out an UnrealScript-side trigger I haven't located, but the in-game observation that the weapon never re-holsters under any circumstance — even pre-combat — is consistent with "the client doesn't autonomously emit it."

## Why this matters for the animation system

`USGWAnim_BlendByWeapon` keys off the holster bit ([source path `.\Src\SGWAnim_BlendByWeapon.cpp` in the binary](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/abilities/use_ability.rs#L225-L227)) — this is why the in-combat clearing of `BSF_HOLSTER` is critical for combat animations to render correctly, and is the load-bearing reason the [`commit_sets_bsf_in_combat_and_clears_bsf_holster`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/abilities/use_ability.rs#L553-L577) regression test exists.

The flip side: with the bit never-set, the animation rig stays in the "weapon-up" branch permanently. Cosmetic-only; doesn't break gameplay, but it's wrong.

## Suggested investigation / fix paths

These are not exclusive — the right answer may be more than one.

1. **Seed `BSF_HOLSTER` at character world-entry.** Cheapest. In whatever path constructs the cell entity for a player joining the world, set the bit before the first `onStateFieldUpdate` goes out. The first attack/reload will clear it via the existing paths; respawn already calls `clear_all_state_flags` so re-seeding on respawn is a one-line addition there too. Doesn't address re-holster after combat ends, but it does fix the spawn appearance.

2. **Server-driven re-holster after leaving combat.** Plumb a tick that watches for "out of combat for N seconds" — `threatened_mobs.is_empty()` is the existing source of truth (see [`combat/threat.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/combat/threat.rs) and the `regen_tick` precedent at [`service.rs:610`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service.rs#L610)) — and on transition to empty + cooldown elapsed, set `BSF_HOLSTER` and emit `onStateFieldUpdate`. The legacy [`SGWBeing.py:199`](https://github.com/SandboxServers/Cimmeria/blob/main/deprecated/python/cell/SGWBeing.py#L199) TODO ("BSF_Holster - unset when firing (add RequiredMonikers to abilities?)") suggests the original design idea was ability-moniker-driven, but a simpler timer is probably enough.

3. **Find the client emitter (if any) before deciding.** If there is an UnrealScript or input-binding side that *does* try to emit `Event_NetOut_ChangeWeaponState(1)` at idle, then (1) above is the right shape because the client will keep it correct after that. If there's no client emitter, (2) is necessary.

## Repro

1. Log into any character.
2. Observe player model from third-person — weapon is in the active/held position from spawn.
3. Wait indefinitely. Do not enter combat. Weapon stays held.
4. (Confirmed by playtesting; no need to enter combat to reproduce.)

## Update [docs/architecture/state-flag-conventions.md:62](https://github.com/SandboxServers/Cimmeria/blob/main/docs/architecture/state-flag-conventions.md#L62)

The "Mostly idempotent player input via `requestHolsterWeapon`; weapon-fire also unholsters" entry overstates reality. The handler exists but is unreachable in practice (no caller observed). Should either fix the underlying bug and keep the doc, or annotate the doc as a known gap until the bug is addressed.

## Related

- #211 (first-fire animation requires bandolier slot switch) — different symptom (fire animation not playing) but plausibly related; both implicate the animation state machine and may be fixed by adjacent work.
- #219 (`BSF_IN_COMBAT` stuck-set lifecycle) — opposite pathology. Mentions `BSF_HOLSTER` only as a forward-compat caveat. Fixing #219 must continue to preserve the unholster-on-fire/reload paths.

Cadacious:
## What this changes (in plain English)

Before this PR, Stargate Worlds characters were always rendered holding their weapon — even when standing around idle, just walked into a friendly base, or first stepped into the world. Combat-relevant motions (drawing, holstering, reloading) were either invisible or jumped instantly from one pose to another with no animation. The original 2009 game had hand-authored animations for all of this — they just weren't reachable because of a long-standing protocol misunderstanding.

This PR restores the full holster lifecycle the original game had:

- **Players spawn with their weapon on their back / hip** instead of held in hand. The model goes weapon-down by default, the way real soldiers carry weapons when not engaged.
- **Walking into combat plays the unholster animation** — the character actually pulls the weapon out, instead of teleporting it into their hand mid-fight.
- **Walking out of combat plays the holster animation** after a 10-second grace window — the character puts the weapon back, gun stays visible during the motion, then disappears once it's stowed.
- **Reloading plays the reload animation** — a working reload animation for the first time in this project. Reloading from a holstered state draws the weapon first, then chains into the reload.
- **The first shot of any engagement plays the draw animation** — pressing fire on a holstered weapon draws it first, queues the shot, then fires once the animation finishes.
- **Equipping a weapon plays the unholster animation** — whether the player drags it onto the bandolier or right-clicks to equip, the character actually pulls the weapon out.
- **Unequipping a weapon plays the holster animation** — drag back to inventory or right-click in the bandolier; the character stows the weapon before it disappears.
- **Right-click in inventory auto-equips / auto-unequips** — right-click a weapon in the main bag and it equips to the first empty bandolier slot (or replaces the active one if full); right-click in the bandolier and it unequips back to the bag. No drag needed.
- **Bandolier slot swaps (F1/F2/F3/F4) play the full holster-then-unholster choreography** when both slots are armed — the character holsters the old weapon, then draws the new one. In combat the swap is instant (no time to play out animations mid-fight).

Each of these started broken or absent. The chain of fixes restores all of them end-to-end.

## Why the old approach was broken

The 2009 client's `GameBeing_OnStateFieldUpdate` at `ghidra://SGW.exe@0x00e01c90` only dispatches on bits 0-7 of `bStateField`. The change-detection uses `TEST BL, imm8` against the low byte — bit 8 is never tested anywhere. So every `BSF_HOLSTER` (bit 8) write we made on the wire was discarded by the client. That's the #333 bug shape: a state flag that looked load-bearing but was always a no-op on the client side.

The actual visual holster lives in the appearance compositor:

- `CompositedAppearanceProxy::ApplyToPawn` at `ghidra://SGW.exe@0x00ec0840` writes `entity+0x3D2` (the animation-key weapon-category byte) from the proxy's compositing job.
- The job's weapon-category sources from the `Event_NetIn_BeingAppearance` `ComponentList`.
- `USGWAnim_BlendByPosture` keys off `entity+0x3D2` to pick the armed vs. holstered animation blend.

**Holster is a `ComponentList` concern**: include the weapon visual → armed pose; omit it → holstered pose. The fix is to track holster state server-side on the `CellEntity` and filter `weapon_visual` out of the broadcast `BeingAppearance.ComponentList` whenever the player should be in the down-pose.

### Spawn holstered (closes #249)

`CellEntity` gained `weapon_visual: Option<String>` and `weapon_holstered: bool` (defaults `None` / `true`). `appearance_components()` returns the components list with `weapon_visual` filtered out when holstered. `PlayerLoadData` mirrors the shape with `weapon_visual` separated from `components` plus `appearance_components(holstered: bool)`. Every world-entry and appearance-broadcast emit site passes `holstered=true` so the player joins the world weapon-down. The active bandolier slot's visual gets loaded into `weapon_visual` by splitting the equipment-visuals SQL into two queries (one for everything else, one `LIMIT 1` for the active slot's weapon).

### Combat enter → draw weapon

`combat::generate_threat` calls `enter_player_combat` when a player's `threatened_mobs` set transitions from empty to non-empty. That now also flips `weapon_holstered=false` and dispatches `RefreshAppearance` to the base side, which updates the cached holster state and rebroadcasts `BeingAppearance` with the weapon mesh attached. `Item_Equip` (event 4000 → archetype-keyed sequence 1872 for Soldier) fires through `fire_item_sequence` to play the unholster animation.

### Combat exit → 10s OOC grace + two-phase holster

`exit_player_combat` runs when `threatened_mobs` drops back to empty (last threat cleared by death, leash, or timeout). It stamps `combat_exit_at = Some(now)` instead of immediately holstering — re-aggro inside the grace window keeps the weapon drawn so the player doesn't see a holster-then-immediately-draw flash.

`holster_timer_tick` runs every 100ms. When `combat_exit_at` ages past `OOC_HOLSTER_DELAY` (10 seconds):

- **Phase 1**: fires `Item_Unequip` (event 4001 → seq 1873) and arms `holster_animation_complete_at = now + HOLSTER_ANIMATION_DURATION`. The weapon mesh stays attached during the animation.
- **Phase 2** (after the animation): flips `weapon_holstered=true`, dispatches `RefreshAppearance(holstered=true)`, base broadcasts `BeingAppearance` with the empty `ComponentList`, mesh disappears.

### Attack while holstered → queue first shot

`handle_use_ability` detects `is_weapon_attack && weapon_holstered && threatened_mobs.is_empty()` on the first press. It:

1. Sets `weapon_holstered=false`, stamps `combat_exit_at` (arms the OOC re-holster), clears any in-flight Phase 2
2. Stashes `pending_attack_at = now + UNHOLSTER_DRAW_DURATION`, `pending_attack_ability_id`, `pending_attack_target_id`
3. Dispatches `RefreshAppearance(holstered=false)` and fires `Item_Equip`
4. Returns `false` WITHOUT committing cooldown or consuming ammo

`pending_attack_tick` re-invokes `handle_use_ability` once the draw window elapses. Phase B runs the normal fire path against an already-drawn weapon. Subsequent weapon-attack presses during the draw window are rejected (first press locks in the queue). Non-weapon abilities (heals, buffs, self-casts) bypass the queue entirely — they don't need the weapon drawn to function.

### Reload while holstered → two-phase

`handle_reload` detects `weapon_holstered && threatened_mobs.is_empty() && pending_reload_at.is_none()` and runs Phase A:

1. Sets `weapon_holstered=false`, stamps `combat_exit_at`
2. Sets `pending_reload_at = now + UNHOLSTER_DRAW_DURATION`
3. Dispatches `RefreshAppearance(holstered=false)` and fires `Item_Equip` (the draw)
4. Returns without starting the reload cooldown

`pending_reload_tick` re-invokes `handle_reload` once the draw window elapses. Phase B finds the weapon already drawn, runs the normal reload path (`Item_Reload` event 4002 + cooldown + ammo refill schedule). A second R press during the draw window is silently ignored to prevent premature Phase B.

### Drag-to-equip → unholster animation

When the player drags a weapon from main bag into a bandolier slot, the base persists the move and dispatches `BaseToCellMsg::SyncBandolierItems` to the cell. The cell-side handler snapshots `prev_active_item_id`, replaces `bandolier_items`, then compares against `new_active_item_id`. When `active_slot_gained_weapon` is true and the player is OOC:

- `weapon_holstered=false`
- `combat_exit_at = Some(now)` (arms OOC re-holster timer)
- Dispatch `RefreshAppearance(holstered=false)` (mesh attach)
- Fire `Item_Equip` (unholster animation)

### Drag-to-unequip → holster animation

When the active slot LOSES its weapon (drag from bandolier back to main bag), the cell-side handler now fires the holster animation choreography:

- Fire `Item_Unequip` immediately (mesh still attached)
- Arm `holster_animation_complete_at` so `holster_timer_tick` Phase 2 dispatches the eventual mesh removal
- Clear `combat_exit_at` so a stale OOC timer doesn't re-fire the same animation 10 seconds later

The base-side `sync_bandolier_after_inventory_change` accepts a `defer_appearance_refresh` flag for the unequip path so the base doesn't yank the mesh immediately — the cell's Phase 2 owns the eventual broadcast. The empty-bandolier early-return that previously skipped the broadcast entirely is also fixed: unequipping the LAST weapon now correctly broadcasts the empty `ComponentList`.

### Right-click auto-equip / auto-unequip

`useItem` from the client right-click routes to `handle_use_inventory_item`. A single DB query resolves `(type_id, container_id, is_weapon)` via a join with `resources.items`. For weapons (`clip_size IS NOT NULL`) the handler bypasses the `OnItemUse` content event and routes to the move path:

- **Main bag (1) → Bandolier (3)**: equip. `resolve_auto_equip_target` finds the first empty bandolier slot via `bag_max_slots(3)` capped `generate_series`. All four full → replaces the player's active slot.
- **Bandolier (3) → Main bag (1)**: unequip. First empty main bag slot via `bag_max_slots(1)`. Main bag full → silent reject (logged server-side, no client toast — no wire channel for that feedback today).

Non-weapon items still go through the existing `OnItemUse` event so consumables, mission items, and chain-fired uses keep working.

### Bandolier slot swap (F1/F2/F3/F4) holster→swap→unholster choreography

`handle_request_active_slot_change` now detects "both slots populated + OOC + slots differ + no swap already pending":

- Fires `Item_Unequip` for the old weapon
- Stashes `pending_slot_swap_at = now + HOLSTER_ANIMATION_DURATION`, `pending_slot_swap_target = slot_id`
- Returns WITHOUT changing `active_bandolier_slot` or sending `ActiveSlotUpdate`

`pending_slot_swap_tick` fires once the animation window elapses, re-invokes `handle_request_active_slot_change` with the target slot's wire args. The choreography branch short-circuits (gated on `pending_slot_swap_at.is_some()`), the pending state is cleared, and the normal slot-change flow runs — `Item_Equip` for the new weapon, `ActiveSlotUpdate` to base, ammo-type property update.

Skip cases (immediate swap, no choreography): in-combat, target slot empty, source slot empty, same slot, already mid-choreography.

### Initial weapon grant lands in already-active slot

A subtle SQL asymmetry made the first weapon a player ever received stay invisible. `handle_grant_item` only sets `bandolier_became_active=true` if the player's previous `bandolier_slot` selection points at a vacant slot — but the INSERT happens BEFORE the UPDATE, so for an initial grant where `next_slot == p.bandolier_slot == 0`, the NOT EXISTS check finds the just-inserted row and the UPDATE is skipped, leaving `bandolier_became_active=false`. The base then sends `UpdateBandolierItem(make_active=false)` and the cell-side equip-display gate (originally `make_active=true`-only) skipped the draw. The fix widens the gate to also fire when `slot_id == entity.active_bandolier_slot` regardless of `make_active`.

### Legacy `quantity=-1` "move whole stack" sentinel

The 2009 client sends `quantity=-1` on drag-to-equip / drag-to-bag interactions to mean "move the whole stack" (legacy convention from `python/cell/SGWPlayer.py:moveItem`). The base's `handle_move_inventory_item` was rejecting anything `<= 0` as invalid, breaking every drag-into-bandolier and every drag-out-of-bandolier. The fix defers quantity validation past the source-row read; `quantity <= 0` now resolves to `source.stack_size` (the whole stack). The upper-bound `quantity > source.stack_size` check still rejects inflate attempts.

### Visual polish — tightened holster animation timing

`HOLSTER_ANIMATION_DURATION` dropped from the initial 1000ms estimate to **600ms** after two playtest iterations. The constant doesn't represent when the mesh visually disappears on the client — it represents when the cell sends the `RefreshAppearance(holstered=true)` packet. The user reported an end-of-animation flicker: the animation visually ended at ~T, then the blend tree returned to the idle-armed pose (weapon snaps back to hand) for a few frames before the mesh-removal packet landed.

The fix is to send Phase 2 BEFORE the animation visually ends so the `ComponentList` swap on the client coincides with animation completion. Trade-off: the mesh disappears during the last ~100ms of the holster motion (pistol already at thigh socket — barely perceptible) instead of after it (clearly visible snapback). 600ms hit the sweet spot. The constant is easy to tune up or down if a future archetype's animation length differs.

## #333 cleanup (consolidated)

- `BSF_HOLSTER` constant removed from `cell/combat/state.rs` and the re-export
- Dead clear-on-fire write removed from `cell/abilities/use_ability.rs`
- Dead clear-on-reload write removed from `cell/cell_methods/player/world.rs`
- `requestHolsterWeapon` (in `cell_methods/combatant.rs`) now calls `set_weapon_holstered()` instead of toggling the dead bit
- New doc: [`docs/architecture/state-field-bits.md`](docs/architecture/state-field-bits.md) — verified bit→dispatch table and the bit 8 retirement notice, pinned to the Ghidra addresses used to verify
- [`docs/architecture/state-flag-conventions.md`](docs/architecture/state-flag-conventions.md) updated to drop the stale `BSF_HOLSTER` row and cross-link to the retirement notice

## Bug fixes from PR review

| Bug | Source | Fix |
|---|---|---|
| Non-weapon abilities blocked when weapon attack queued | Copilot review | Gated the `queued_attack_already_pending` early reject on `is_weapon_attack` so heals/buffs/self-casts bypass the queue |
| Reload second-press during draw window forces premature Phase B | Copilot review | Added `now < pending_reload_at` short-circuit at top of Phase B |
| Unequip last weapon leaves mesh visible | Playtest | `sync_bandolier_after_inventory_change` no longer early-returns past `refresh_player_appearance` for the empty-bandolier case |
| Stale `BSF_HOLSTER` row in state-flag conventions doc | Copilot review | Removed row, added cross-link |
| `docs/readme.md` architecture count off by one | Copilot + CodeRabbit | "11 documents" → 12 |
| `handle_grant_item` INFO log is high-volume | Copilot review | Demoted to debug; still available via `RUST_LOG=debug` for targeted diagnostics |
| Hard-coded slot ranges in `resolve_auto_equip_target` | Copilot review | Uses `bag_max_slots()`; private `first_empty_slot()` helper |
| ~32 `PR #338` references in source comments | CodeRabbit + Copilot | Stripped — provenance belongs in PR body |

## Test plan

- [x] `cargo clippy -p cimmeria-entity --all-targets -- -D warnings` clean
- [x] **749 services tests** pass locally (started at 716 pre-PR; ~33 regression guards added across the holster + bandolier work)
- [x] **160 entity tests** pass locally
- [x] In-game playtest: spawn → weapon on back, weapon-down pose
- [x] In-game playtest: dialog grants pistol → equip animation plays, weapon appears
- [x] In-game playtest: combat enter (interact with NPC) → unholster animation + first-shot queue → shot fires
- [x] In-game playtest: combat exit → 10s grace → holster animation + mesh removal, no flicker
- [x] In-game playtest: reload while holstered → draw animation → reload animation → ammo refill
- [x] In-game playtest: right-click weapon in main bag → equip animation + mesh attach
- [x] In-game playtest: right-click weapon in bandolier → holster animation + mesh removal
- [x] In-game playtest: drag weapon main bag → bandolier → equip animation
- [x] In-game playtest: drag weapon bandolier → main bag → holster animation
- [ ] In-game playtest: F1/F2/F3/F4 slot swap with two armed slots → holster-then-unholster (added in 650c94aa, awaiting confirmation)

## Out of scope (tracked in #340)

- File-size hard-cap violations introduced or worsened by this work (`base_messages.rs` 1790, `world.rs` 1005, `ticks.rs` 997, `cell_entity/mod.rs` 845, `use_ability.rs` 834). Cataloged in #340 with suggested split directions.
- Test inventory refresh in `docs/testing/inventory/` — likely past the 5% drift threshold after this PR. Same maintenance PR.
- Doc sweep per `CLAUDE.md`'s "what changed → what to update" map. Same maintenance PR.
- Dead `weapon_visual` field on `CellEntity` (only set in tests). Lower priority cleanup.

## Closes

- #249 (players spawn holstered)
- #333 (`BSF_HOLSTER` bit 8 retirement)

## Summary by CodeRabbit

* **New Features**
  * Two‑phase weapon holster/draw with out‑of‑combat auto‑holster grace, spawn‑holstered appearances, and choreographed bandolier swaps.
  * Ability to queue a weapon attack, reload, or slot swap during draw/holster windows; appearance rebroadcasts ensure mesh/animation sync.
  * Server now includes archetype info to drive item equip/unequip/reload animations.

* **Tests**
  * Expanded unit and integration tests covering holster flows, draw‑queueing, and slot‑swap choreography.

* **Chores**
  * Updated state‑flag and architecture docs reflecting holster behavior.


MyPalClara:
## Category 1: Correctness Issues (G1-G5)

### G1 — `appearance_components()` string comparison

**Location:** `crates/entity/src/cell_entity/mod.rs` and `crates/services/src/mercury/types.rs`

**What's Wrong:** The filter uses exact string equality. If `weapon_visual = "pistol"` but `components` contains `"Pistol"` (case mismatch) or `"pistol "` (trailing whitespace), the filter fails silently and the weapon stays visible when holstered.

**Why It Probably Doesn't Matter:** The data comes from the same DB query pipeline, so case/whitespace mismatches are unlikely. But it's fragile.

**Recommended Fix:** Either add a debug assertion that `weapon_visual` exists in `components`, or document the invariant explicitly.

**Effort:** 10 minutes

---

### G2 — `appearance_components()` clones Vec every call

**Location:** Both implementations

**What's Wrong:** Every call allocates a new `Vec<String>`. This is called on every `BeingAppearance` emit. For 100 players in AoI, that's 100 allocations per broadcast tick.

**Why It Probably Doesn't Matter:** Allocation is cheap, and this isn't a hot path yet.

**Recommended Fix (if profiling shows it's hot):** Return `Cow<'_, [String]>` instead.

**Effort:** 30 minutes

---

### G3 — `set_weapon_holstered()` return semantics

**What's Wrong:** Returns `false` for both "no change" and "no weapon". A caller can't distinguish them. The doc explains this, but the API is footgun-shaped.

**Why It's Probably Fine:** Current callers don't need to distinguish.

**Effort:** 45 minutes (if needed later)

---

### G4 — `requestHolsterWeapon` doesn't rebroadcast

**What's Wrong:** Phase 2 is deferred. Server-side state changes but witnesses don't see it. The holster button is still broken, just broken differently.

**Why It's Acceptable:** PR body explicitly calls this out.

**Recommended Fix:** File a follow-up issue for Phase 2 and add a TODO comment.

**Effort:** 15 minutes

---

### G5 — SQL split creates two round-trips

**What's Wrong:** `query_player_load_data` now does two queries instead of one.

**Why It's Probably Fine:** Both queries hit the same table with indexed lookups. Latency impact is negligible.

**Effort:** 30 minutes (if needed)

MyPalClara:
## Category 2: Missing Test Coverage (G6-G10)

### G6 — No test for duplicate weapon visual in components

**What's Missing:** What if `components` contains `["torso", "pistol", "pistol", "head"]`? Filter removes all occurrences — probably correct, but untested.

**Effort:** 10 minutes

---

### G7 — No integration test for world-entry holster state

**What's Missing:** Unit tests verify `appearance_components()` filtering, but no test verifies the full world-entry path emits correct wire bytes with the weapon filtered out.

**Effort:** 45 minutes

---

### G8 — No test for `requestHolsterWeapon` cell method

**What's Missing:** The dispatch calls `set_weapon_holstered()`, but there's no test that verifies the cell method actually updates the entity state.

**Effort:** 30 minutes

---

### G9 — No test for weapon swap while holstered

**What's Missing:** If `weapon_holstered = true` and player changes active bandolier slot, `weapon_visual` updates. New weapon should also be filtered. Untested.

**Effort:** 15 minutes

---

### G10 — No test for `PlayerLoadData::appearance_components` invariant

**What's Missing:** The doc says "when `weapon_visual` is `Some(v)`, the same string is also present in `components`" — but this invariant isn't tested.

**Effort:** 15 minutes

MyPalClara:
## Category 3: Documentation Gaps (G11-G13)

### G11 — `state-field-bits.md` doesn't cross-reference the Bible

**What's Missing:** The new doc is excellent, but it doesn't link to:
- `mercury-wire-format.md` — packet structure
- `entity-property-sync.md` — `BeingAppearance.ComponentList` wire format

**Recommended Fix:** Add a "Related Documentation" section with links.

**Effort:** 10 minutes

---

### G12 — No Phase 2 tracking issue

**What's Missing:** The PR body explains Phase 2 (runtime holster toggle rebroadcast), but there's no:
1. In-code `// TODO(#xxx)` comment
2. Follow-up issue filed
3. Link between the two

A future developer might not know that `requestHolsterWeapon` is half-implemented.

**Recommended Fix:** File issue + add TODO comment in `combatant.rs`.

**Effort:** 15 minutes

---

### G13 — `PlayerLoadData::weapon_visual` invariant not enforced

**What's Missing:** The doc says "when `weapon_visual` is `Some(v)`, the same string is also present in `components`" — but this invariant is not enforced at construction time, not tested, and not validated with a debug assertion.

**Recommended Fix:** Add a debug assertion in the constructor or validation method.

**Effort:** 20 minutes

MyPalClara:
## Category 4: Edge Cases (G14-G15)

### G14 — What if `weapon_visual` not in `components`?

**What's Missing:** The filter is a no-op if `weapon_visual` isn't found in `components`. This violates the documented invariant (G13). It's probably a data bug upstream, but the function silently accepts it.

**Recommended Fix:** Log a warning when the invariant is violated:
```rust
if filtered.len() == self.components.len() {
    tracing::warn!(
        weapon = %weapon,
        components = ?self.components,
        "weapon_visual not found in components — invariant violated"
    );
}
```

**Effort:** 15 minutes

---

### G15 — What if `refresh_player_appearance` fails?

**What's Missing:** The function is called with `await` but the result isn't checked. If the appearance refresh fails, the equipment move still succeeded but witnesses don't see it.

**Why It's Probably Fine:** Silent failure is acceptable here. The move is the primary operation; the appearance broadcast is best-effort.

**Recommended Fix (if you want observability):**
```rust
if let Err(e) = refresh_player_appearance(...).await {
    tracing::warn!(error = %e, "Failed to refresh player appearance");
}
```

**Effort:** 10 minutes

MyPalClara:
## Category 5: Code Quality (G16-G17) + Summary

### G16 — Duplicate `appearance_components()` implementations

**What's Wrong:** `CellEntity::appearance_components()` and `PlayerLoadData::appearance_components(bool)` are nearly identical. The only difference is where `holstered` comes from (field vs parameter).

**Recommended Fix:** Extract a shared helper function:
```rust
pub fn filter_holstered_weapon(
    components: &[String],
    weapon_visual: Option<&str>,
    holstered: bool,
) -> Vec<String>
```

**Effort:** 30 minutes

---

### G17 — `TEST_FLAG` naming in state.rs

**What's Wrong:** The tests use bit 12 as `TEST_FLAG`. The name is generic. A future developer might not understand why it exists.

**Recommended Fix:** Use a more descriptive name like `BSF_TEST_UNUSED_BIT_12` with a doc comment explaining it's in the unused bit range (8-15).

**Effort:** 5 minutes

---

## Summary Table

| Category | Count | Severity | Total Effort |
|----------|-------|----------|--------------|
| Correctness Issues (G1-G5) | 5 | Low-Medium | ~2 hours |
| Missing Test Coverage (G6-G10) | 5 | Low-Medium | ~2 hours |
| Documentation Gaps (G11-G13) | 3 | Low | ~45 min |
| Edge Cases (G14-G15) | 2 | Low | ~25 min |
| Code Quality (G16-G17) | 2 | Low | ~35 min |
| **Total** | **17** | | **~5.5 hours** |

---

## Priority Matrix

🔴 **Before Merge:** Nothing — PR is ready to merge

🟡 **Immediately After Merge (~1 hour):**
- G4/G12 — File Phase 2 tracking issue + add TODO comment

🟢 **Follow-up PR (~2 hours):**
- G7 — Integration test for world-entry holster
- G8 — Test for `requestHolsterWeapon` cell method

⚪ **Tech Debt / Nice-to-Have:**
- G16 — Dedupe implementations
- G2 — `Cow` optimization (only if profiling shows it's hot)
- Everything else

---

## The Verdict

**This PR is ready to merge.** 

18 files, +554/-132 lines, touching 5 crates. The fix is correct, the root cause analysis is perfect, the documentation is exemplary, and the dead code removal is complete.

The 17 gaps are all Low or Low-Medium severity. None are blocking. The most important follow-up is filing the Phase 2 tracking issue (G4/G12) so the half-implemented `requestHolsterWeapon` doesn't get forgotten.

**Ship it.** 🎯

Cadacious:
@coderabbitat review

Cadacious:
**Phase 2 pushed** (commit `742a8e42`) — the runtime toggle the original PR noted as deferred is now in scope.

## What landed

- `CellEntity::sync_holster_to_combat(in_combat)` — pins the invariant that `weapon_holstered = !BSF_InCombat`.
- `cell::combat::threat::enter_player_combat` and `exit_player_combat` flip `weapon_holstered` alongside the existing `BSF_InCombat` change.
- New `CellToBaseMsg::RefreshAppearance { entity_id, player_id, holstered }` variant + base-side handler. The handler maintains a `ConnectedClientState::weapon_holstered` cache (default `true`) and rebuilds + broadcasts `BeingAppearance` via the existing `refresh_player_appearance` pipeline.
- `refresh_player_appearance` now reads the cached flag instead of hardcoding `holstered=true`, so a mid-combat equip / slot swap doesn't silently re-holster.
- `request_appearance_refresh` cell-side helper called from: `damage_apply` (on `generate_threat` → enter combat), `death.rs` (on `clear_dead_npc_from_all_player_threat` → exit combat), and `requestHolsterWeapon` (the TODO Phase 1 left in place).

## Symptom this fixes

From the in-game UAT log after Phase 1 deployed: fists animated correctly on combat enter/exit (that's `BSF_InCombat` driving `UpdateCombatStanceWeaponSet` — independent path, always worked), but no weapon was ever visible because Phase 1 emitted exactly one `BeingAppearance` per session. Phase 2 fires a fresh broadcast on every combat transition and on the explicit holster button.

## Test additions

- `sync_holster_to_combat_inverts_combat_flag` — invariant pin.
- `enter_player_combat_draws_weapon_when_bsf_flips` + `exit_player_combat_holsters_weapon_when_bsf_clears` — bug-shape regression guards. Both fail if someone refactors `weapon_holstered` away from `enter_player_combat` / `exit_player_combat`.
- `request_holster_weapon_dispatches_refresh_appearance` + `request_holster_weapon_no_op_skips_refresh_appearance` — pins the state-change-gated cell→base dispatch.

725 services + 160 entity tests pass. `cargo fmt`, `cargo clippy -- -D warnings` clean. Ready for re-review.

PR #343, opened 2026-05-21
feat(launcher): replace Tauri prototype with egui + Azure Blob seed/patch manifest

BangRocket:
## Summary

- **Scraps** the Tauri 2 WebView2 launcher in `crates/launcher/` (~6,989 lines of JS/HTML/Rust + Tauri configs + auto-generated schemas).
- **Replaces** with a native **egui (eframe)** Windows app driven off a **seed + patch JSON manifest on Azure Blob Storage**. Single ~5 MB .exe, no JS toolchain.
- Adds **3 GitHub Actions workflows** mirroring the server release pattern: path-filtered build CI, `workflow_dispatch` release, and `/release-launcher` ChatOps on merged PRs.
- Adds **2 new docs** (user/operator guide + Azure runbook) and supersedes the two Tauri-era plans.

## What the new launcher does

```text
1. Fetch manifest.json from Azure Blob (anonymous GET).
2. Diff manifest vs <install>/launcher-installed.json.
3. Install / Update:
     - If seed sha mismatches → download seed.zip (Range-resumable) → sha256 verify → extract.
     - For each missing patch in declared order: download → verify → extract overlay.
     - Byte-patch SGW.exe .rdata: replace `www.stargateworlds.com` with configured server host.
4. Launch buttons appear based on file detection:
     - SGW.exe → "Launch SGW.exe"
     - AteraLoader.exe + AtreaGameDebug.bat → "Launch Atera Debug"
     - AtreaFixASLR.bat → "Fix ASLR"
5. Upload Debug Logs: zip Binaries/sgwdebuglog* + Binaries/sessions/** → single PUT
   to Azure SAS URL. Content-digest dedupe ledger means repeat clicks against unchanged
   logs make zero HTTP requests.
```

State files: `launcher-config.json` next to `.exe`, `launcher-installed.json` inside the game dir (survives launcher reinstall), `uploaded.json` next to `.exe`.

## Crate layout (10 flat src files)

| File | Purpose |
|------|---------|
| `main.rs` + `app.rs` | eframe entry + egui panels |
| `manifest.rs` | schema-1 JSON parser + validate (forward-ref + duplicate-id checks) |
| `install.rs` | resumable seed/patch download + sha256 verify + zip extract |
| `patch_rdata.rs` | 22-byte `.rdata` hostname patch (salvaged from old launcher) |
| `launch.rs` | Atera detection + SGW.exe / batch spawning |
| `logs.rs` | log collection + zip + single Azure Blob PUT |
| `state.rs` | InstalledState + UploadedLedger |
| `worker.rs` | tokio worker, Command/Event channels |

## Workflows

| File | Trigger |
|------|---------|
| [`.github/workflows/launcher-build.yml`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/.github/workflows/launcher-build.yml) | Path-filtered PRs touching `crates/launcher/**`. Runs fmt (ubuntu) + clippy/build/test (windows-latest). |
| [`.github/workflows/launcher-release.yml`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/.github/workflows/launcher-release.yml) | `workflow_dispatch`. Builds release exe with `LAUNCHER_LOG_SAS_URL` injected from secrets, creates GitHub Release tagged `launcher-<date>-<sha7>`. |
| [`.github/workflows/launcher-release-on-comment.yml`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/.github/workflows/launcher-release-on-comment.yml) | `issue_comment` `/release-launcher` on merged PRs (mirror of `release-on-comment.yml`). |

The launcher stays in `WORKSPACE_EXCLUDES` in `test.yml` so eframe Linux deps don't slow the main CI pipeline — `launcher-build.yml` owns its full cycle on `windows-latest`.

## Docs

- [`docs/client/sgw-launcher.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/sgw-launcher.md) — rewritten design doc
- [`docs/client/launcher-guide.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/launcher-guide.md) — **new** user + operator guide (player workflow + how to publish patches)
- [`docs/client/launcher-storage-setup.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/launcher-storage-setup.md) — **new** Azure container / SAS minting / lifecycle runbook
- `docs/plans/2026-03-06-sgw-launcher-{design,plan}.md` — marked superseded with link forward
- `crates/README.md`, `docs/readme.md`, `CLAUDE.md` — index entries updated

## Verification (local, Windows)

```text
cargo check  -p sgw-launcher                                 OK (4m07s fresh)
cargo clippy -p sgw-launcher --all-targets -- -D warnings    clean
cargo test   -p sgw-launcher                                 40/40 pass
cargo fmt    -p sgw-launcher -- --check                      exit 0
launcher GUI                                                 opens, panels
                                                             render, manifest
                                                             fetch path runs
```

## Out-of-band setup required before release is functional

Beyond merging this PR:

1. Create the Azure storage account + `sgw` container per [`launcher-storage-setup.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/launcher-storage-setup.md).
2. Mint a SAS URL with `c` (create) permission scoped to the container.
3. Add it as the repo secret `LAUNCHER_LOG_SAS_URL`.
4. Build + publish the seed zip and (eventually) patch zips per [`launcher-guide.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/launcher-guide.md) Part 2.

Without the secret, the release workflow still builds — log upload just stays permanently disabled in that binary.

## Test plan

- [ ] CI: `launcher` workflow (path-filtered) runs and is green
- [ ] CI: existing `ci` workflow stays green (launcher excluded via `WORKSPACE_EXCLUDES`)
- [ ] Manual: download an artifact from a workflow run, launch it on a clean Windows machine, confirm window renders + manifest-error UI displays when storage account is not yet created
- [ ] Manual (post-storage-setup): manifest fetch succeeds → Install / Update applies seed + patches → SGW.exe `.rdata` patched → Launch SGW.exe works
- [ ] Manual: drop Atera files into install dir → "Launch Atera Debug" + "Fix ASLR" buttons appear
- [ ] Manual: click "Upload Debug Logs" twice in a row with unchanged logs → second click reports "Already uploaded" without an HTTP request

## Summary by CodeRabbit

* **New Features**
  * Native Windows launcher with modern GUI
  * Manifest-driven install pipeline with resumable downloads and integrity checks
  * Ed25519-signed manifest verification for tamper protection
  * Debug-log collection, deterministic zipping, and single-request upload with dedupe
  * Improved install progress/status and multiple launch modes (normal, debug, ASLR-fix)

* **Documentation**
  * Added comprehensive user guide, implementation doc, and distribution/runbook for launcher

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/343?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MyPalClara:
## Category 1: Security Concerns (G1-G3)

---

### G1: Zip-slip protection comment needed
**Location:** `crates/launcher/src/install.rs` — `extract_zip()`
**Severity:** Low

The code correctly uses `entry.enclosed_name()` which rejects absolute paths and `..` components. But there's no comment explaining this is the security gate.

**Risk:** A future maintainer might "simplify" to `entry.name()` and introduce zip-slip vulnerability.

**Recommendation:**
```rust
// SECURITY: enclosed_name() is critical here — it rejects absolute paths
// and path traversal (../) to prevent zip-slip attacks. Do NOT replace
// with entry.name() or entry.mangled_name().
let Some(relative) = entry.enclosed_name() else {
    continue;
};
```

**Effort:** 2 minutes

---

### G2: SAS URL baked into binary
**Location:** `crates/launcher/src/logs.rs` — `AZURE_SAS_URL`
**Severity:** Low-Medium

The SAS URL is compiled into the binary. If leaked (decompilation, memory dump), anyone can upload arbitrary blobs to your Azure container. The SAS has `c` (create) permission only, so they can't read/delete, but they can fill storage with garbage.

**Options:**
1. **Accept the risk** — document it in the Azure runbook, set up storage alerts for unusual upload volume
2. **Rotate via server** — launcher fetches short-lived SAS from your API on each upload
3. **Use managed identity** — if launcher ever runs in Azure context

**Recommendation:** Option 1 is fine for a game launcher. Document the risk and set up monitoring.

**Effort:** 30 minutes (documentation + Azure alert)

---

### G3: No hostname validation in patch_rdata
**Location:** `crates/launcher/src/patch_rdata.rs`
**Severity:** Low

No validation that `server_host` is a valid hostname. If someone enters malicious input as the server host, it gets written into the .exe. The .exe won't execute it (it's just a string in .rdata), but it's still weird and could cause confusion.

**Recommendation:**
```rust
fn is_valid_hostname(s: &str) -> bool {
    s.len() <= 22
        && s.chars().all(|c| c.is_ascii_alphanumeric() || c == '.' || c == '-')
        && !s.starts_with('-')
        && !s.ends_with('-')
}
```

**Effort:** 15 minutes

---

**Category 1 Total Effort:** ~45 minutes

MyPalClara:
## Category 2: Error Handling (G4-G6)

---

### G4: Range request handling — ACTUALLY FINE
**Location:** `crates/launcher/src/install.rs` — `download_to_file()`
**Severity:** None (I was wrong)

I initially flagged this as a bug, but on closer inspection the code is correct:

```rust
let mut file = if resumed && existing_len > 0 {
    tokio::fs::OpenOptions::new().append(true).open(dest).await?
} else {
    tokio::fs::File::create(dest).await?  // truncates and starts fresh
};
```

If the server returns 200 (doesn't support Range), `resumed` is false, and `File::create` truncates the existing partial file. **This is correct behavior.**

**Status:** No action needed.

---

### G5: Unbounded event channel
**Location:** `crates/launcher/src/worker.rs`
**Severity:** Low

The event channel from worker to UI is unbounded. If the UI thread stalls (e.g., during a long egui layout), events pile up unboundedly.

**Why this is probably fine:** The launcher UI is simple — no complex layouts, no blocking operations on the UI thread. Events are small (enum variants with a few fields). You'd need to stall for minutes to accumulate meaningful memory.

**If you want to fix it:**
```rust
let (tx, rx) = tokio::sync::mpsc::channel(100);  // bounded
// Then handle SendError by dropping oldest or blocking
```

**Effort:** 15 minutes (if desired)

---

### G6: Config save failure doesn't block install
**Location:** `crates/launcher/src/app.rs`
**Severity:** Low

Config is saved on every "Install / Update" click. If the save fails, the install still proceeds. The error is logged to status, but the user might not notice.

**Current behavior:**
```rust
if let Err(e) = self.config.save() {
    self.status.push(format!("Warning: failed to save config: {e}"));
}
// install proceeds anyway
```

**Why this might be intentional:** Config save failing (disk full, permissions) shouldn't block the user from installing the game. The config will be re-saved on next successful operation.

**If you want stricter behavior:**
```rust
if let Err(e) = self.config.save() {
    self.status.push(format!("ERROR: Cannot save config: {e}"));
    return;  // block install
}
```

**Effort:** 5 minutes (if desired)

---

**Category 2 Total Effort:** ~20 minutes (all optional)

MyPalClara:
## Category 3: Missing Test Coverage (G7-G10)

---

### G7: No test for download resume path
**Location:** `crates/launcher/src/install.rs`
**Severity:** Medium

The resume logic (Range header, 206 status, append mode) is the most complex code path in the file, but it's untested.

**Recommendation:**
```rust
#[tokio::test]
async fn download_resumes_partial_file() {
    // Start a mock server that supports Range requests
    let server = MockServer::start().await;
    let full_content = b"0123456789ABCDEF";
    
    Mock::given(method("GET"))
        .and(header("Range", "bytes=8-"))
        .respond_with(ResponseTemplate::new(206)
            .body(&full_content[8..])
            .insert_header("Content-Range", "bytes 8-15/16"))
        .mount(&server).await;
    
    // Create partial file with first 8 bytes
    let dir = tempfile::tempdir().unwrap();
    let dest = dir.path().join("test.bin");
    std::fs::write(&dest, &full_content[..8]).unwrap();
    
    // Resume should append remaining bytes
    download_to_file(&server.uri(), &dest, &tx, token).await.unwrap();
    
    assert_eq!(std::fs::read(&dest).unwrap(), full_content);
}
```

**Effort:** 45 minutes (including mock server setup)

---

### G8: No test for cancellation during download/extract
**Location:** `crates/launcher/src/install.rs`
**Severity:** Low-Medium

The `CancellationToken` is threaded through, but there's no test that verifies cancellation actually stops the operation mid-stream.

**Recommendation:**
```rust
#[tokio::test]
async fn download_respects_cancellation() {
    let token = CancellationToken::new();
    let token_clone = token.clone();
    
    // Cancel after 100ms
    tokio::spawn(async move {
        tokio::time::sleep(Duration::from_millis(100)).await;
        token_clone.cancel();
    });
    
    // Start a slow download
    let result = download_to_file(&slow_server_url, &dest, &tx, token).await;
    
    assert!(result.is_err());
    // Partial file should exist but be incomplete
}
```

**Effort:** 30 minutes

---

### G9: No tests for worker.rs
**Location:** `crates/launcher/src/worker.rs`
**Severity:** Low

The worker is the glue between UI and async operations. It's simple (just a command dispatcher), but untested.

**What to test:**
- Command::FetchManifest → Event::ManifestFetched or Event::Error
- Command::Install → Event::Progress... → Event::InstallComplete
- Command::Cancel → operations stop

**Effort:** 1 hour

---

### G10: No tests for UI state transitions
**Location:** `crates/launcher/src/app.rs`
**Severity:** Low

The `drain_events` function has complex state transitions:
- `Event::ManifestFetched` → updates `manifest`, `installed_state`
- `Event::InstallComplete` → clears `installing`, updates `installed_state`
- `Event::Error` → clears `installing`, pushes to `status`

These are untested. UI state machines are notoriously buggy.

**Why this is hard to test:** egui apps are typically tested via screenshot comparison or manual testing, not unit tests. The state transitions could be extracted into a pure function for testing.

**Effort:** 1-2 hours (if desired)

---

**Category 3 Total Effort:** ~3-4 hours

MyPalClara:
## Category 4: UX Concerns (G11-G12)

---

### G11: Status log memory leak
**Location:** `crates/launcher/src/app.rs`
**Severity:** Low

The status log display is capped at 100 lines, but the underlying `Vec<String>` grows unboundedly:

```rust
// Display caps at 100
for line in self.status.iter().rev().take(100) {
    ui.label(line);
}

// But the Vec keeps growing
self.status.push(format!("Downloaded {} bytes", n));
```

Over a long session with many operations, this leaks memory.

**Recommendation — Ring buffer:**
```rust
const MAX_STATUS_LINES: usize = 100;

fn push_status(&mut self, msg: String) {
    if self.status.len() >= MAX_STATUS_LINES {
        self.status.remove(0);
    }
    self.status.push(msg);
}
```

Or use `VecDeque` for O(1) removal from front.

**Effort:** 10 minutes

---

### G12: No download speed or ETA display
**Location:** `crates/launcher/src/app.rs` — progress bar
**Severity:** Low (nice-to-have)

The progress bar shows bytes downloaded and total, but not:
- Download speed (MB/s)
- Estimated time remaining

For a multi-GB seed download, users want to know how long they'll be waiting.

**Recommendation:**
```rust
struct DownloadProgress {
    started_at: Instant,
    bytes_downloaded: u64,
    bytes_total: u64,
}

impl DownloadProgress {
    fn speed_mbps(&self) -> f64 {
        let elapsed = self.started_at.elapsed().as_secs_f64();
        if elapsed > 0.0 {
            (self.bytes_downloaded as f64 / 1_000_000.0) / elapsed
        } else {
            0.0
        }
    }
    
    fn eta_seconds(&self) -> Option<u64> {
        let speed = self.speed_mbps();
        if speed > 0.0 {
            let remaining = self.bytes_total - self.bytes_downloaded;
            Some((remaining as f64 / (speed * 1_000_000.0)) as u64)
        } else {
            None
        }
    }
}
```

**Effort:** 30 minutes

---

**Category 4 Total Effort:** ~40 minutes

MyPalClara:
## Category 5: Documentation (G13-G14)

---

### G13: No schema versioning strategy documented
**Location:** `crates/launcher/src/manifest.rs`
**Severity:** Low

The code rejects `schema != 1`, but there's no doc explaining how to add schema 2 in the future.

**Questions a future maintainer will have:**
- Do I bump to schema 2 for any breaking change?
- Should the launcher support multiple schema versions simultaneously?
- What's the deprecation policy for old schemas?

**Recommendation:** Add to `docs/launcher/manifest-schema.md`:
```markdown
## Schema Versioning

The manifest `schema` field is an integer version number.

### When to bump the schema version:
- Adding required fields → bump
- Removing fields → bump  
- Changing field semantics → bump
- Adding optional fields → no bump needed

### Backwards compatibility:
The launcher supports only the current schema version. When bumping:
1. Update `SUPPORTED_SCHEMA` constant
2. Update manifest generation in CI
3. All users must update their launcher

### Current schema: 1
```

**Effort:** 20 minutes

---

### G14: No troubleshooting section in launcher guide
**Location:** `docs/launcher/launcher-guide.md`
**Severity:** Low

Users will encounter issues. The guide should help them self-diagnose.

**Recommendation:** Add a troubleshooting section:
```markdown
## Troubleshooting

### "Failed to fetch manifest"
- Check your internet connection
- Verify the manifest URL is accessible in a browser
- Check if your firewall/antivirus is blocking the launcher

### "Hash mismatch" after download
- The download may have been corrupted
- Click "Install / Update" to re-download
- If persistent, the CDN cache may be stale — wait 15 minutes

### "SGW.exe won't launch"
- Ensure Visual C++ Redistributable 2015-2022 is installed
- Try running as Administrator
- Check Windows Event Viewer for crash details

### "Stuck at 0%" download
- Large seed files may take time to start
- Check if another download is in progress
- Try cancelling and restarting

### Logs location
Launcher logs are stored in: `<install_dir>/Logs/`
Upload them via the "Upload Logs" button for support.
```

**Effort:** 30 minutes

---

**Category 5 Total Effort:** ~50 minutes

---

## Summary

| Category | Gaps | Severity | Effort |
|----------|------|----------|--------|
| Security | G1-G3 | Low-Medium | 45 min |
| Error Handling | G4-G6 | Low | 20 min |
| Test Coverage | G7-G10 | Low-Medium | 3-4 hr |
| UX | G11-G12 | Low | 40 min |
| Documentation | G13-G14 | Low | 50 min |
| **Total** | **14** | | **~6 hours** |

---

## Priority

| Priority | Gaps | Action |
|----------|------|--------|
| 🟢 Quick wins | G1, G11 | Zip-slip comment, ring buffer (15 min) |
| 🟡 Should do | G2, G14 | Document SAS risk, add troubleshooting (1 hr) |
| 🟡 Follow-up PR | G7, G8 | Resume + cancellation tests |
| ⚪ Nice-to-have | Everything else |

---

**Total: 14 gaps, ~6 hours of follow-up work, none blocking.**

BangRocket:
## Review feedback addressed in `ec2e78d`

Tackled the quick wins + medium items locally, full pipeline green (`cargo fmt --check` / `clippy -D warnings` / `test`, **49/49 passing** — up from 40 with 9 new tests on push_status cap, blob_url query/fragment stripping, and hostname validation).

### Items addressed in this fixup

**Code:**
- ✅ **app.rs** — Status `Vec<String>` capped at 1000 via `push_status()` helper. *(MyPalClara G11 + CodeRabbit)*
- ✅ **install.rs** — Replaced panic-prone `unwrap_err()` with new `InstallError::UnexpectedStatus` variant; comment now matches the invariant. *(Copilot + CodeRabbit)*
- ✅ **install.rs** — `SECURITY:` comment above `enclosed_name()` in `extract_zip` documenting the zip-slip gate. *(MyPalClara G1)*
- ✅ **manifest.rs** — `blob_url` now strips `?query` and `#fragment` from the manifest URL before splitting. SAS-tokened manifest URLs now produce valid blob URLs. *(Copilot)*
- ✅ **patch_rdata.rs** — Hostname validation before writing to `.rdata`: ASCII alnum + `.` + `-`, no leading/trailing `-`/`.`, non-empty. New `PatchError::InvalidHostname`. *(MyPalClara G3)*

**Workflows:**
- ✅ **launcher-release.yml** — Pinned all 4 third-party actions to immutable commit SHAs (`actions/checkout`, `dtolnay/rust-toolchain`, `Swatinem/rust-cache`, `actions/upload-artifact`) with version comments. `persist-credentials: false` on checkout. This is the workflow that consumes the SAS secret + creates Releases, so the supply-chain surface matters most here. *(CodeRabbit + zizmor)*
- ✅ **launcher-build.yml** — `persist-credentials: false` on all 4 checkout steps as cheap defence-in-depth.
- N/A — **launcher-release-on-comment.yml** has no `actions/checkout`, only `gh api` calls. Nothing to harden.

**Docs:**
- ✅ **sgw-launcher.md** — Icon link points at `icons/icon.ico` not the `icons/` dir. *(CodeRabbit)*
- ✅ **launcher-storage-setup.md** — Grammar fix on "For higher security preferential is…". *(Copilot)*
- ✅ **launcher-storage-setup.md** — New **Threat model — SAS in the client binary** section with concrete operator action items (storage cost cap, upload-rate anomaly alert, 6-month rotation cadence). *(MyPalClara G2 + CodeRabbit — accept-and-document was the chosen disposition)*
- ✅ **launcher-guide.md** — New **Schema Versioning Policy** subsection in Step 3, covering when to bump vs not, the multi-schema transition procedure, and pointers for migrations that don't fit the additive-optional shape. *(MyPalClara G13)*
- ✅ **launcher-guide.md** — New **Troubleshooting** section covering manifest errors, hash mismatch, HTTP status codes, launch failures, SAS expiry, dedupe-skipped uploads, and where to find diagnostic files. *(MyPalClara G14)*

### Deferred to follow-up issues

Per reviewer classification ("Follow-up PR" / "Nice-to-have") and the chosen scope for this PR:

- **G7 / G8** — Integration tests for the download-resume and cancellation paths. Real coverage gaps; want a mock-HTTP-server harness rather than ad-hoc tests.
- **G9 / G10** — `worker.rs` orchestration tests + `app.rs` UI state-machine tests. Higher-lift; extracting state transitions into a pure fn first would make this cleaner.
- **Copilot perf items** — `spawn_blocking` for `extract_zip` and streaming file reads in `logs::build_log_zip`. Real but bounded: seed/patch zips run during user-driven install (UI freeze is fine) and log inputs are small (`sgwdebuglog` + sessions/, well under 100 MB in practice).
- **G12** — Download speed + ETA display.
- **SAS broker endpoint** — The "Major" CodeRabbit asked for. Requires a server-side issuer that doesn't exist yet. Chosen disposition: accept + document, with the threat-model section and operator action items providing mitigation through monitoring + lifecycle policy. Revisit if the SAS ever gets extracted in the wild.

CI should rerun on the new commit; expect all 15 checks green again.

Cadacious:
## Does it do what it says on the tin?

Mostly yes. Verified against the code:
- Resumable download via HTTP Range ([`install.rs:177-208`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/install.rs#L177))
- SHA-256 verify on every artifact
- Patches applied in declared order, dedup via `applied_patches` ([`install.rs:88-95`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/install.rs#L88))
- State file lives inside install dir, survives launcher reinstall ([`state.rs:31-34`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/state.rs#L31))
- Content-digest dedupe survives re-zips (digest is over `(relpath, bytes)`, not zip bytes)
- `enclosed_name()` prevents zip-slip ([`install.rs:289`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/install.rs#L289))

### Bug — hostname re-patch when `server_host` changes

`needs_patching(data)` is `find_pattern(data, "www.stargateworlds.com").is_some()` ([`patch_rdata.rs:52`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/patch_rdata.rs#L52)).

Sequence that breaks:
1. User installs with `server_host = play.cimmeria.gg` → patch succeeds, original CME string is gone.
2. User edits the host field to `staging.cimmeria.gg` → clicks Install / Update.
3. `needs_patching` returns `false` (CME string already replaced) → patch is **skipped**.
4. Client connects to old host, no warning to the user.

**Remediation** — store the patched-in host in state, repatch when it differs:

```rust
// state.rs
pub struct InstalledState {
    pub applied_patches: Vec<String>,
    pub seed_sha256: Option<String>,
    #[serde(default)]
    pub patched_host: Option<String>,
}
```

```rust
// install.rs - replace the current needs_patching gate
let needs_repatch = patch_rdata::needs_patching(&data)
    || state.patched_host.as_deref() != Some(ctx.server_host);

if needs_repatch {
    patch_rdata::patch_exe_any(
        &exe,
        ctx.server_host,
        state.patched_host.as_deref(),  // fallback pattern to match
    )?;
    state.patched_host = Some(ctx.server_host.to_string());
    state.save(ctx.install_dir)?;
}
```

Add a `patch_exe_any` helper that searches for either the original CME string **or** the previously-patched host (zero-padded to 22 bytes). Regression test reproduces the sequence above and fails before the fix — see Test Gaps comment §7c.

### Minor: hostname patch is null-terminated C-string territory

[`patch_rdata.rs:45-47`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/patch_rdata.rs#L45) zero-fills the 22-byte slot. Correct for C strings; if SGW.exe ever reads the field length-prefixed (it almost certainly doesn't) the zero-pad would be wrong. Worth a one-line note in the module doc.

### Minor: Cancel doesn't clean up `.tmp-*` files

Intentional — enables resume on next run. But if cancel was triggered by *correcting* a wrong manifest URL, the stale tmp lingers and the next install's Range request asks an unrelated URL to continue from byte N. SHA verify catches it. Either document in `launcher-guide.md` ("if you change manifest URLs, delete `.tmp-*` from the install dir") or add a "Reset downloads" button.

Cadacious:
## Feature gaps vs. what a real launcher needs

Solid v1. Gap matrix vs. typical MMO launcher functionality:

| Capability | Status | Notes |
|---|---|---|
| Install game from manifest | yes | |
| Auto-update on launcher start | manual button | Should auto-apply or prompt |
| **Self-update of the launcher itself** | **no** | Every launcher fix = manual re-download from GH Releases |
| **Repair / verify game files** | **no** | Only recovery from corruption = nuke `launcher-installed.json` + re-seed (multi-GB) |
| News / patch notes | no | Manifest could carry `news_url` or `changelog` field |
| Server selection | single host field | OK for one shard; blocker for beta+live |
| EULA / first-run | no | Standard for binary patching of someone else's game |
| Crash detection / auto-log-upload | manual button | Detect abnormal SGW.exe exit, prompt to upload |
| Disk-space precheck | no | Mid-install ENOSPC just dumps an IO error |
| Pause / in-session resume | cancel only | Resume is per-launcher-run |
| ETA / speed display | no | Bytes only |
| Multi-channel (beta/live) | no | One manifest URL |
| Atomic install | no | Failed extract leaves half-applied patch |
| Concurrent-launcher protection | no | Two instances race on state.json + .tmp files |
| Admin-rights detection | no | Default `C:\Program Files` requires admin → opaque mid-extract failure |
| AV / SmartScreen | no | Unsigned .exe → warning on every first download |

### The two biggest — self-update and repair/verify

Without **self-update**, every launcher bug = a "please redownload from GH Releases" announcement.
Without **repair**, every AV-quarantined-a-file report = "delete the whole game, reinstall."

Both inexpensive.

**Self-update sketch:**

```rust
// New manifest field
pub struct Manifest {
    ...
    #[serde(default)]
    pub launcher: Option<LauncherEntry>,
}
pub struct LauncherEntry {
    pub version: String,
    pub blob: String,
    pub sha256: String,
}
```

```rust
// In LauncherApp::new, after manifest fetch:
if let Some(l) = &manifest.launcher {
    if l.version > env!("CARGO_PKG_VERSION") {
        // download to sgw-launcher.new.exe, rename current to .old,
        // rename new to current, std::process::Command::new(current).spawn(),
        // std::process::exit(0). Standard Windows self-update pattern.
    }
}
```

**Repair sketch** — store per-file sha for the seed:

```rust
// InstalledState gains
pub struct InstalledState {
    ...
    #[serde(default)]
    pub seed_files: std::collections::HashMap<String, String>,
}
// Populated at install time from a manifest of seed contents (publish step
// emits this alongside seed.zip).

// "Verify" button hashes each tracked file, lists mismatches, re-downloads
// seed.zip and overlays only the bad files.
```

### Admin-rights / install path

Default install path is `C:\Program Files\Stargate Worlds` — requires admin to write. UAC isn't requested, so the install fails opaquely mid-extract with "Access denied".

**Pick one:**

1. **Best UX** — default to `%LOCALAPPDATA%\Stargate Worlds` (no UAC). Matches Steam's new-install default.
2. **Adequate** — runtime check, prompt to re-launch elevated.
3. **Adequate** — early writability probe, bail before downloading:

```rust
fn install_dir_writable(dir: &Path) -> bool {
    let _ = std::fs::create_dir_all(dir);
    let probe = dir.join(".launcher-write-probe");
    let ok = std::fs::write(&probe, b"x").is_ok();
    let _ = std::fs::remove_file(&probe);
    ok
}
```

Gate the Install button on this, with a friendly error when it fails.

### Atomic install

Mid-patch extract failure currently leaves half-applied state. Wrap each patch in a staging dir + rename:

```rust
async fn apply_patch(ctx: &InstallContext<'_>, patch: &PatchEntry) -> Result<(), InstallError> {
    let staging = ctx.install_dir.join(format!(".staging-{}", safe_id(&patch.id)));
    std::fs::create_dir_all(&staging)?;

    // download + verify into .tmp-*, extract into staging
    extract_zip(&tmp, &staging, ...)?;

    // Only after extract succeeds, overlay staging onto install_dir.
    move_overlay(&staging, ctx.install_dir)?;
    std::fs::remove_dir_all(&staging).ok();
    Ok(())
}
```

`move_overlay` walks staging and `std::fs::rename`s each file into the install dir (same volume → atomic per-file). If anything fails before `move_overlay`, the staging dir is the only thing left to clean up.

### Single-instance lock

```rust
// main.rs - before eframe::run_native
use fs2::FileExt;
let lock_path = config::exe_dir().join("launcher.lock");
let lock = std::fs::File::create(&lock_path)?;
if lock.try_lock_exclusive().is_err() {
    // Show a native message box or just stderr + exit
    eprintln!("Another Stargate Worlds launcher is already running.");
    std::process::exit(1);
}
// hold `lock` for process lifetime
```

Add `fs2 = "0.4"` to `Cargo.toml`. Test in §7f.

Cadacious:
## Security improvements beyond the SAS

### 4a. Manifest is unsigned — biggest gap

`fetch_manifest` does `reqwest::get(url)`. Anyone who can serve `manifest.json` (compromised Azure key, social-engineered support, MITM with HTTP downgrade) can ship any seed/patch — SHA verification just confirms what the attacker said the hash should be.

**Remediation — embed an Ed25519 pubkey, ship detached signature:**

```rust
// manifest.rs
const SIGNING_PUBKEY: &[u8; 32] = include_bytes!("../keys/manifest-signing.pub");

pub async fn fetch_manifest(client: &reqwest::Client, url: &str)
    -> Result<Manifest, ManifestError>
{
    let body = client.get(url).send().await?.error_for_status()?.bytes().await?;
    let sig = client.get(format!("{url}.sig")).send().await?
        .error_for_status()?.bytes().await?;

    use ed25519_dalek::{Signature, VerifyingKey, Verifier};
    let vk = VerifyingKey::from_bytes(SIGNING_PUBKEY.try_into().unwrap())?;
    let signature = Signature::from_slice(&sig)?;
    vk.verify(&body, &signature).map_err(|_| ManifestError::BadSignature)?;

    let manifest: Manifest = serde_json::from_slice(&body)?;
    manifest.validate()?;
    Ok(manifest)
}
```

Add `ed25519-dalek = { version = "2", default-features = false, features = ["pkcs8"] }` to `Cargo.toml`.

Operator workflow change: signing step in the publish flow runs `ed25519sign manifest.json > manifest.json.sig` with an offline private key. Document in [`launcher-guide.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/launcher-guide.md) Part 2.

**This is the single highest-leverage security improvement.** Removes manifest tampering and seed/patch substitution from the threat model. Anonymous-read on the container is fine once the manifest is signed.

### 4b. Replay protection — monotonic version

Once signed, the remaining attack is replaying an old (validly signed) manifest to roll back patches. Defense:

```rust
pub struct Manifest {
    pub schema: u32,
    pub version: u64,                                  // monotonic
    pub published_at: chrono::DateTime<chrono::Utc>,
    pub seed: SeedEntry,
    pub patches: Vec<PatchEntry>,
}

// state.rs
pub struct InstalledState {
    ...
    #[serde(default)]
    pub highest_manifest_version: Option<u64>,
}

// install.rs - reject regressions
if let Some(seen) = state.highest_manifest_version {
    if manifest.version < seen {
        return Err(InstallError::ManifestReplay {
            seen, got: manifest.version,
        });
    }
}
state.highest_manifest_version = Some(manifest.version);
```

### 4c. `SGW.exe` write is destructive without backup

[`patch_rdata.rs:56-61`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/patch_rdata.rs#L56): `read → modify → write`. Power loss between truncate and finish leaves a corrupt .exe with no recovery beyond re-seeding (multi-GB).

**Fix — atomic via rename:**

```rust
pub fn patch_exe(exe_path: &Path, new_host: &str) -> Result<usize, PatchError> {
    let mut data = std::fs::read(exe_path)?;
    let offset = patch_hostname(&mut data, new_host)?;
    let tmp_path = exe_path.with_extension("exe.patching");
    std::fs::write(&tmp_path, &data)?;
    std::fs::rename(&tmp_path, exe_path)?;  // atomic on NTFS, same volume
    Ok(offset)
}
```

### 4d. HTTPS-only enforcement

Belt-and-braces against a user typing `http://...` in the manifest URL field:

```rust
// manifest.rs - top of fetch_manifest
if !url.starts_with("https://") {
    return Err(ManifestError::InsecureUrl(url.to_string()));
}
```

Same in `logs.rs::upload_blob` for the SAS URL. Plus on the `reqwest::Client` itself:

```rust
reqwest::Client::builder()
    .https_only(true)  // refuses to follow http:// redirects too
    .build()
```

### 4e. Concurrent install / state race

Covered functionally in the Feature Gaps comment (single-instance lock). Security framing: a second launcher process can write a malicious `launcher-installed.json` claiming patches are applied, causing the first to skip them. The file lock closes both.

### 4f. SAS-in-binary trade-offs

The SAS lives in `.rdata` of the launcher .exe. Anyone can `strings sgw-launcher.exe | grep blob.core` to extract it. `c`-only scope limits blast radius to:

- Upload spam / DoS storage-account quota
- Egress cost amplification
- Arbitrary content under `logs/` (SAS is container-scoped per [`launcher-storage-setup.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/launcher-storage-setup.md) "Important" callout)

**Mitigations short of removing the SAS:**

- Per-prefix SAS via **stored access policy** (closes #3):
  ```bash
  az storage container policy create \
    --account-name cimmeriastorage \
    --container-name sgw \
    --name logs-write-only \
    --permissions c \
    --expiry "$EXPIRY"
  az storage account generate-sas ... --policy-name logs-write-only
  ```
- **Lifecycle policy on `logs/`** to auto-delete after 30 days — caps total $ leak. (Runbook addition.)
- **Azure Front Door** rate limit by IP in front of storage — closes #1 and #2 partially.

**To eliminate the SAS entirely** — see Alternatives comment.

### 4g. `LauncherConfig` schema drift

No `schema_version` field. Future schema changes will silently revert to defaults via `.ok().unwrap_or_default()`. Add:

```rust
pub struct LauncherConfig {
    #[serde(default = "current_schema_version")]
    pub schema_version: u32,
    pub install_path: PathBuf,
    pub server_host: String,
    pub manifest_url: String,
}

fn current_schema_version() -> u32 { 1 }

impl LauncherConfig {
    pub fn load(path: &Path) -> Result<Self, ConfigError> {
        let cfg: Self = serde_json::from_str(&std::fs::read_to_string(path)?)?;
        if cfg.schema_version != current_schema_version() {
            return Err(ConfigError::SchemaVersion {
                got: cfg.schema_version, expected: current_schema_version()
            });
        }
        Ok(cfg)
    }
}
```

### 4h. `launch.rs` path-prefix sanity check

`Command::new(install_dir.join("SGW.exe"))` — install_dir is user-controllable via the config field. Not really an escalation (the user chose the path), but cheap hardening:

```rust
fn spawn(install_dir: &Path, file: &str, via_cmd: bool) -> Result<u32, LaunchError> {
    let install_dir = install_dir.canonicalize()?;
    let path = install_dir.join(file).canonicalize()?;
    if !path.starts_with(&install_dir) {
        return Err(LaunchError::PathEscape(path));
    }
    ...
}
```

Cadacious:
## Alternatives to SAS / anonymous auth for blob hosting

Ranked by cost vs. security gain.

### Option A — GitHub Releases as the CDN (cheapest, zero infra)

The launcher itself is already shipped via GH Releases. Ship `seed.zip`, `patches/*.zip`, and `manifest.json` as release assets too.

```
https://github.com/SandboxServers/Cimmeria/releases/download/launcher-2026-05-20-abc1234/manifest.json
https://github.com/SandboxServers/Cimmeria/releases/download/launcher-2026-05-20-abc1234/seed.zip
```

- **Cost:** $0. Free egress.
- **Security:** `actions/attest-build-provenance` provides sigstore attestation during the release workflow — solves §4a manifest signing as a side effect.
- **Downside:** 2GB per-asset limit may force seed splitting; rate limits exist (generous but real for a popular emu); no per-blob ACL.

**Implementation sketch:**

```yaml
# launcher-release.yml additions
- name: build seed + patches
  run: ./tools/build-seed.sh    # already-existing publish flow
- name: attest provenance
  uses: actions/attest-build-provenance@v1
  with:
    subject-path: 'release/seed.zip,release/patches/*.zip,release/manifest.json'
- name: upload assets
  run: |
    gh release upload "${TAG}" release/seed.zip release/patches/*.zip release/manifest.json
```

### Option B — Cloudflare R2 + Worker auth (moderate cost, strong)

R2 has zero egress fees. Tiny Worker validates a launcher-supplied token (signed by your auth server, ~5min TTL) and either redirects to a signed R2 URL or proxies bytes.

- **Cost:** ~free at MMO emu scale (R2 free tier + Workers free tier 100k req/day).
- **Security:** Short-lived per-session tokens, rate-limited per-IP at the edge, container fully private.
- **Downside:** Couples launcher to the auth server. Bootstrap problem: a brand-new user has no account. Solve with a separate anonymous "install token" endpoint (rate-limited, IP-throttled) or accept the coupling.

### Option C — Azure Front Door + Function gateway (Azure-native)

Container fully private; Function App validates and rewrites to authenticated blob URLs.

- **Cost:** ~$22/mo Front Door base + small Function spend.
- **Security:** Full control, IP allowlist, rate limit, token validation.
- **Downside:** Most engineering for a v1.

### Option D — Self-hosted nginx with signed-URL middleware

`secure_link` module on a $5/mo VPS. URLs like `https://patches.cimmeria.gg/seed.zip?md5=...&expires=...`.

- **Cost:** $5/mo + ops time.
- **Security:** Same model as CloudFront signed URLs.
- **Downside:** Back in the "ops a server" business.

### Option E — Sign the manifest, keep blobs anonymous

Covered in Security comment §4a. **Highest-leverage, lowest cost** (~50 LOC + a signing step in publish flow). Anonymous-read on Azure is fine afterward.

### Recommendation

1. **Do manifest signing (§4a) immediately** regardless of which option you pick for blob hosting.
2. To drop SAS-in-binary: **Option A (GH Releases)** — lowest friction, already in your release pipeline, sigstore attestation free.
3. **Move log upload off Azure SAS entirely** — small POST endpoint on the Cimmeria server with per-session rate limiting. Eliminates SAS attack surface and gives you a place to dedupe server-side too.

### What a future ADR should capture

Drop into `docs/architecture/launcher-distribution.md`:

- **Decision:** which option (or hybrid).
- **Why not the others:** specifically the trade-offs being accepted.
- **Bootstrap:** how a brand-new user with no account gets the launcher running.
- **Operator workflow:** how a release is published (signing step, attestation, upload).
- **Disaster recovery:** signing-key rotation flow. Embed key *version* in the launcher and accept multiple key versions during a transition window so a leaked key doesn't strand existing installs.

Cadacious:
## Code-level issues

### 6a. `install_path: String` should be `PathBuf`

[`config.rs:30`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/config.rs#L30), and `default_install_path() -> String`. Works today; tomorrow it bites someone with a long path or non-UTF-16-safe Unicode boundary.

```rust
pub struct LauncherConfig {
    pub install_path: PathBuf,    // was String
    pub server_host: String,
    pub manifest_url: String,
}
```

Propagates to `app.rs` (TextEdit needs `.display().to_string()` for editing, parse back on save) and `worker.rs`.

### 6b. `&PathBuf` should be `&Path` (`clippy::ptr_arg`)

[`config.rs:46,51`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/config.rs#L46). Textbook lint. CI is `-D warnings` — either it's allowed somewhere or hasn't fired yet on this PR. Either way:

```rust
pub fn load(path: &Path) -> Result<Self, ConfigError> { ... }
pub fn save(&self, path: &Path) -> Result<(), ConfigError> { ... }
```

### 6c. State writes are non-atomic

`InstalledState::save`, `UploadedLedger::save`, `LauncherConfig::save` all do `std::fs::write` directly. Power loss mid-write corrupts the JSON; the load path falls back to `default()`, which silently means "nothing is installed" → re-download the entire seed.

**Fix everywhere** (promote `tempfile` to a runtime dep, or just rename a `.tmp` file manually):

```rust
// new utility in a shared module
pub fn atomic_write(path: &Path, bytes: &[u8]) -> std::io::Result<()> {
    if let Some(p) = path.parent() {
        std::fs::create_dir_all(p)?;
    }
    let tmp = path.with_extension("tmp");
    std::fs::write(&tmp, bytes)?;
    std::fs::rename(&tmp, path)?;  // atomic on NTFS, same volume
    Ok(())
}
```

Use in:
- `state.rs::InstalledState::save`
- `state.rs::UploadedLedger::save`
- `config.rs::LauncherConfig::save`
- `patch_rdata.rs::patch_exe` (covered in Security comment §4c)

### 6d. `status: Vec<String>` grows unbounded

[`app.rs:22`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/app.rs#L22). UI displays last 100; Vec keeps everything. Long-session leak.

```rust
const STATUS_CAP: usize = 500;

impl LauncherApp {
    fn push_status(&mut self, line: String) {
        self.status.push(line);
        if self.status.len() > STATUS_CAP {
            let drop_n = self.status.len() - STATUS_CAP;
            self.status.drain(0..drop_n);
        }
    }
}
```

Replace every direct `self.status.push(...)` with `self.push_status(...)`.

### 6e. Install-dir rescan runs on UI thread

[`app.rs:128`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/app.rs#L128) — `last_refresh.elapsed() > 2s` triggers `refresh_install_state` on the UI thread. On a slow network drive or Defender-scanned path, this stutters the UI. Move to the worker:

```rust
// worker.rs
pub enum Command {
    ...
    RefreshInstallState { install_dir: PathBuf },
}
pub enum Event {
    ...
    InstallStateRefreshed {
        installed: InstalledState,
        launch_opts: LaunchOptions,
    },
}

fn spawn_refresh(&self, install_dir: PathBuf) {
    let tx = self.events_tx.clone();
    // Refresh involves file I/O; use a blocking task.
    self.runtime.spawn_blocking(move || {
        let installed = InstalledState::load(&install_dir);
        let launch_opts = LaunchOptions::detect(&install_dir);
        let _ = tx.send(Event::InstallStateRefreshed { installed, launch_opts });
    });
}
```

App requests refresh on the 2s timer, applies the event when it arrives.

### 6f. Cancel token reuses across back-to-back installs

[`worker.rs:74-77`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/worker.rs#L74). If two installs are dispatched back-to-back (rapid double-click), they share the cancel token and one cancel fires both. Per-task tokens:

```rust
pub struct Worker {
    runtime: Arc<Runtime>,
    pub events_rx: mpsc::UnboundedReceiver<Event>,
    events_tx: mpsc::UnboundedSender<Event>,
    current_install_cancel: Option<CancellationToken>,
}

fn spawn_install(&mut self, config: LauncherConfig, manifest: Manifest) {
    let cancel = CancellationToken::new();
    self.current_install_cancel = Some(cancel.clone());
    // pass `cancel` into the InstallContext
    ...
}

fn cancel_current_install(&mut self) {
    if let Some(t) = self.current_install_cancel.take() {
        t.cancel();
    }
}
```

### 6g. `reqwest::Client` is built per call in 3 places

[`install.rs:177`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/install.rs#L177), [`logs.rs:123`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/logs.rs#L123), [`manifest.rs:87`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/manifest.rs#L87). Single shared `Client` reuses the connection pool across seed → patch1 → patch2:

```rust
// worker.rs
pub struct Worker {
    ...
    http: reqwest::Client,
}

impl Worker {
    pub fn new(runtime: Arc<Runtime>) -> Self {
        let http = reqwest::Client::builder()
            .https_only(true)          // §4d defense-in-depth
            .build()
            .expect("rustls client");
        Self { ..., http }
    }
}
```

Pass `&Client` into `install_all`, `fetch_manifest`, `upload_blob`.

### 6h. `UploadedLedger` never prunes

A player who clicks "Upload Logs" 1000 times grows the ledger to 1000 entries.

```rust
const LEDGER_KEEP: usize = 100;

pub fn record(&mut self, sha256: String, blob_name: String) {
    self.entries.push(UploadedEntry {
        sha256,
        blob_name,
        uploaded_at: chrono::Utc::now(),
    });
    if self.entries.len() > LEDGER_KEEP {
        let drop_n = self.entries.len() - LEDGER_KEEP;
        self.entries.drain(0..drop_n);
    }
}
```

### 6i. Tmp filename keys patches by `id` only, not `(id, sha)`

[`install.rs:144`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/crates/launcher/src/install.rs#L144) uses `.tmp-patch-{safe_id}.zip`. If a patch is republished with the same id but a new sha (operationally wrong, but possible during a hot-fix), the launcher resumes the old tmp via Range → SHA fails → user must manually delete. Cheap fix: include `&p.sha256[..12]` in the tmp filename.

```rust
let safe_sha = &patch.sha256[..12];
let tmp = ctx.install_dir.join(
    format!(".tmp-patch-{safe_id}-{safe_sha}.zip")
);
```

### 6j. `pub use` discipline for module structure

10 flat files in `src/`. Per [CLAUDE.md](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/CLAUDE.md) file-organization rules ("flat names for 2-3 siblings; directory for 4+"), this is on the high side. They *are* different themes (manifest, install, launch, logs, state, worker, app, config, patch_rdata, main), each at one file, so flat is justified — but as the launcher grows you'll want to promote some to directories (e.g., `install/` for download/extract/state-update split). Not a v1 concern; flag for future.

Cadacious:
## Test gaps

Coverage of the building blocks is strong (40 tests, well-targeted unit-level). Orchestration is what's missing — and §2's hostname re-patch bug would have been caught by a single regression test.

### 7a. End-to-end `install_all` with local HTTP fixture

Add `wiremock = "0.6"` (or `mockito`) as a dev-dep. Test seed downloads → patches apply in declared order → state file reflects truth → re-run is a no-op (zero HTTP requests on the second pass).

```rust
// install_e2e_tests.rs
use wiremock::{MockServer, Mock, ResponseTemplate};
use wiremock::matchers::{method, path};

#[tokio::test]
async fn install_all_seed_then_patches_then_idempotent() {
    let server = MockServer::start().await;

    // Build synthetic seed.zip + patch zips as byte buffers
    let seed_bytes = build_zip(&[("Binaries/SGW.exe", b"www.stargateworlds.com..." as &[u8])]);
    let seed_sha = sha256_hex(&seed_bytes);

    let manifest = json!({
        "schema": 1,
        "seed": { "blob": "seed.zip", "size": seed_bytes.len(), "sha256": seed_sha },
        "patches": []
    });

    Mock::given(method("GET")).and(path("/seed.zip"))
        .respond_with(ResponseTemplate::new(200).set_body_bytes(seed_bytes.clone()))
        .expect(1)              // exactly once across the whole test
        .mount(&server).await;

    let dir = tempfile::tempdir().unwrap();
    let (tx, _rx) = tokio::sync::mpsc::unbounded_channel();
    let ctx = InstallContext {
        manifest_url: &format!("{}/manifest.json", server.uri()),
        install_dir: dir.path(),
        manifest: &serde_json::from_value(manifest).unwrap(),
        server_host: "play.cimmeria.gg",
        cancel: CancellationToken::new(),
        progress: tx.clone(),
    };
    install_all(ctx).await.unwrap();

    // Second pass: should be a no-op, mock expects exactly 1 GET.
    let ctx2 = InstallContext { /* same */ };
    install_all(ctx2).await.unwrap();
    // wiremock verifies .expect(1) on drop — second pass would push to 2 and fail.
}
```

### 7b. Reseed invalidates `applied_patches`

```rust
#[tokio::test]
async fn reseed_clears_applied_patches() {
    let dir = tempfile::tempdir().unwrap();
    let state = InstalledState {
        applied_patches: vec!["a".into(), "b".into()],
        seed_sha256: Some("OLD-SHA".into()),
        ..Default::default()
    };
    state.save(dir.path()).unwrap();

    // Run install_all with a manifest whose seed sha is NEW.
    install_all(/* manifest with seed_sha=NEW */).await.unwrap();

    let loaded = InstalledState::load(dir.path());
    assert!(loaded.applied_patches.is_empty(), "reseed should clear patch list");
    assert_eq!(loaded.seed_sha256.as_deref(), Some("NEW"));
}
```

### 7c. Repatches when `server_host` changes — the §2a bug as a failing guard

```rust
#[test]
fn repatches_when_server_host_changes() {
    let dir = tempfile::tempdir().unwrap();
    let exe = dir.path().join("SGW.exe");
    let mut data = vec![0u8; 100];
    data[40..62].copy_from_slice(b"www.stargateworlds.com");
    std::fs::write(&exe, &data).unwrap();

    // First patch with host A.
    patch_rdata::patch_exe(&exe, "play.cimmeria.gg").unwrap();
    let after_first = std::fs::read(&exe).unwrap();
    assert!(after_first.windows(b"play.cimmeria.gg".len())
        .any(|w| w == b"play.cimmeria.gg"));

    // Change host to B. After the fix from §2a, this must replace
    // play.cimmeria.gg with staging.cimmeria.gg in the binary.
    let state = InstalledState {
        patched_host: Some("play.cimmeria.gg".into()),
        ..Default::default()
    };
    patch_rdata::patch_exe_any(&exe, "staging.cimmeria.gg",
                               state.patched_host.as_deref()).unwrap();

    let after_second = std::fs::read(&exe).unwrap();
    assert!(after_second.windows(b"staging.cimmeria.gg".len())
        .any(|w| w == b"staging.cimmeria.gg"),
        "second patch with different host must update the binary");
    assert!(!after_second.windows(b"play.cimmeria.gg".len())
        .any(|w| w == b"play.cimmeria.gg"),
        "old host must be gone");
}
```

This test fails today — that's the point. It locks in the bug shape from the hostname re-patch finding.

### 7d. Resume from partial `.tmp-*`

```rust
#[tokio::test]
async fn resumes_partial_download_via_range_header() {
    let server = MockServer::start().await;

    Mock::given(method("GET"))
        .and(path("/seed.zip"))
        .and(header("range", "bytes=512-"))
        .respond_with(ResponseTemplate::new(206).set_body_bytes(&second_half))
        .expect(1)
        .mount(&server).await;

    let dir = tempfile::tempdir().unwrap();
    let tmp = dir.path().join(".tmp-seed-abc123.zip");
    std::fs::write(&tmp, &first_half).unwrap();  // pre-existing partial

    download_to_file(&format!("{}/seed.zip", server.uri()), &tmp, ...).await.unwrap();

    let final_bytes = std::fs::read(&tmp).unwrap();
    assert_eq!(final_bytes, [&first_half[..], &second_half[..]].concat());
}
```

### 7e. Explicit zip-slip lock-in

`enclosed_name()` handles it, but lock the behavior so a future zip-crate version can't regress:

```rust
#[test]
fn extract_zip_rejects_path_traversal() {
    let dir = tempfile::tempdir().unwrap();
    let zip_path = dir.path().join("evil.zip");
    {
        let f = std::fs::File::create(&zip_path).unwrap();
        let mut zw = zip::ZipWriter::new(f);
        let opts: zip::write::FileOptions<()> = zip::write::FileOptions::default();
        zw.start_file("../../../evil.txt", opts).unwrap();
        zw.write_all(b"pwned").unwrap();
        zw.finish().unwrap();
    }
    let out = dir.path().join("out");
    let (tx, _) = tokio::sync::mpsc::unbounded_channel();
    extract_zip(&zip_path, &out, &tx, "test", &CancellationToken::new()).unwrap();

    // Nothing written outside `out`:
    assert!(!dir.path().join("evil.txt").exists());
    assert!(!std::path::Path::new("/tmp/evil.txt").exists());
}
```

### 7f. Single-instance lock refusal

After the §3 single-instance lock lands:

```rust
#[test]
fn second_instance_cannot_acquire_lock() {
    use fs2::FileExt;
    let dir = tempfile::tempdir().unwrap();
    let lock_path = dir.path().join("launcher.lock");

    let first = std::fs::File::create(&lock_path).unwrap();
    first.try_lock_exclusive().unwrap();

    let second = std::fs::File::create(&lock_path).unwrap();
    assert!(second.try_lock_exclusive().is_err());
}
```

### 7g. Manifest signature verification (when §4a lands)

```rust
#[test]
fn fetch_manifest_rejects_bad_signature() {
    // Build a manifest, sign with key A, verify with key B → reject.
}

#[test]
fn fetch_manifest_accepts_valid_signature() {
    // Build a manifest, sign with the embedded test key, verify → accept.
}
```

Tests run against a `MockServer` serving `manifest.json` + `manifest.json.sig`.

Cadacious:
@coderabbitai review

BangRocket:
## Round 2 feedback addressed — 6 new commits

This is a big swing. Per scope choices (everything except self-update/repair + manifest signing + GH Releases pivot), the work landed across **6 focused commits** on top of the previous fixup:

| SHA | Headline |
|---|---|
| `e7e54dc` | **🐛 Bug:** Re-patch SGW.exe when `server_host` changes + Cady #1 nits |
| `9e01ffd` | Atomic writes, `PathBuf`, `schema_version`, ledger prune, `%LOCALAPPDATA%` default |
| `44bcfb3` | Shared `reqwest::Client`, cancel-per-install, sha-keyed tmp, atomic exe write |
| `ed8c7b7` | Security hardening: single-instance file lock, HTTPS-only, path-prefix check, writability probe |
| `e300b5b` | **🔐 Manifest signing — Ed25519** (Cady #4a) |
| `af6db5e` | **🚀 GH Releases pivot** (Cady #4 Option A) |

**Tests**: 49 → 69 (+20). All gates green locally (`fmt --check`, `clippy -D warnings`, full suite).

---

## What's now fixed

### Real bug (Cady §1)
- Hostname re-patch silent skip when `server_host` changes. New `patch_hostname_any` / `host_differs` / `patched_host` state field. UI re-enables Install/Update when the only drift is the hostname. Regression test in `e7e54dc`.

### Cady's code-level (§5)
- ✅ **6a** `install_path: String` → `PathBuf` (with `install_path_text: String` UI buffer)
- ✅ **6b** `&PathBuf` → `&Path` in fn signatures
- ✅ **6c** Atomic JSON writes via shared `atomic_write(path, bytes)` (`InstalledState`, `UploadedLedger`, `LauncherConfig`) + atomic `SGW.exe` write via `.exe.patching` + rename
- ✅ **6d** Status `Vec` cap (already fixed in previous fixup)
- ✅ **6f** Cancel token per install (`current_install_cancel: Option<CancellationToken>`)
- ✅ **6g** Shared `reqwest::Client` with `https_only(true)` from the worker
- ✅ **6h** `UploadedLedger` prunes to `LEDGER_KEEP = 100`
- ✅ **6i** Tmp filename includes sha prefix (`.tmp-patch-{id}-{sha12}.zip`)

Deferred: **6e** (rescan on worker) and **6j** (file layout) — not v1 concerns per Cady.

### Cady's security (§3)
- ✅ **4a** Manifest signing — Ed25519 detached signatures. Pubkey embedded via `LAUNCHER_MANIFEST_PUBKEY_HEX` env at compile time; dev default works for PR/CI builds. No unsigned-manifest fallback.
- ✅ **4c** Atomic `SGW.exe` write
- ✅ **4d** HTTPS-only enforcement (both `Client::https_only(true)` and explicit prefix checks in `fetch_manifest` / `upload_blob`)
- ✅ **4g** `LauncherConfig.schema_version` with explicit unsupported-schema rejection
- ✅ **4h** `launch.rs` canonicalize + `starts_with(install_dir)` check, returns `LaunchError::PathEscape` on violation
- ✅ **4e** Closed by the single-instance file lock

Deferred: **4b** Manifest replay protection — needs the signing foundation that just landed; clean follow-up.

### Cady's feature gaps (§2)
- ✅ Single-instance lock via `fs4::FileExt::try_lock_exclusive` in `main.rs`
- ✅ Default install path → `%LOCALAPPDATA%\Stargate Worlds` (no UAC required)
- ✅ Writability probe gates the Install button with a friendly error
- ✅ Manifest signing addresses the auth/signing column

Deferred (per "skip only self-update/repair"): self-update, repair/verify, news, server selection, EULA, crash detection, ETA display.

### CodeRabbit's 3 new nits
- ✅ `InvalidHostname` error text now mentions both `.` and `-`
- ✅ Troubleshooting doc heading updated to "Unexpected HTTP &lt;status&gt;"
- ✅ Regression test for `InstallError::UnexpectedStatus` (wiremock-based)

---

## GH Releases pivot ([`af6db5e`](https://github.com/SandboxServers/Cimmeria/commit/af6db5e))

Two release-tag families:
- **`content-current`** (mutable rolling): holds `manifest.json` + `manifest.json.sig`. Operator publishes a new manifest via `gh release upload content-current --clobber`.
- **`content-YYYY-MM-DD-NNN`** (immutable per-publication): holds `seed.zip` and per-patch zips. Never overwritten.

Manifest references blobs by absolute URL into the immutable tags. `blob_url()` now passes absolute URLs through unchanged; relative paths still work for backwards compat. Default `manifest_url` flipped to the GH Releases convention.

The Azure Blob runbook (`launcher-storage-setup.md`) was rewritten as [`launcher-distribution-setup.md`](https://github.com/SandboxServers/Cimmeria/blob/feat/egui-launcher-azure-blob/docs/client/launcher-distribution-setup.md), covering all three operator concerns in one place: GH Releases publish flow, manifest signing keypair, Azure Blob SAS for log uploads. Log uploads stay on Azure SAS (GH Releases doesn't support unauthenticated writes), with the threat-model section unchanged.

---

## Operator setup needed before the first release

Beyond merging this PR:

1. **Generate the manifest signing keypair** offline. Private key → password manager. Public key → repo secret `LAUNCHER_MANIFEST_PUBKEY_HEX` (64 hex chars).
2. **Create the `content-current` release** with `gh release create content-current --notes "Rolling manifest pointer"`.
3. **Stand up the Azure storage** for log uploads (same as before — runbook in `launcher-distribution-setup.md` Part 3).
4. **Build the first content drop**: seed.zip + manifest.json + signed manifest.json.sig.

The dev-default pubkey lets CI builds and local dev work without operator setup. Only the release pipeline injects the real key.

---

## Filed as follow-up issues (not in this PR)

These were deferred to keep PR scope tractable. Each is its own coherent change:

| # | Item | Source |
|---|---|---|
| 1 | Atomic install via staging dir + rename overlay | Cady §2 |
| 2 | E2E `install_all` test with wiremock (seed → patches → idempotent re-run) | Cady #7a |
| 3 | Resume from partial `.tmp-*` test | Cady #7d / MyPalClara G7 |
| 4 | Explicit zip-slip lock-in test | Cady #7e |
| 5 | Single-instance lock refusal test | Cady #7f |
| 6 | Manifest replay protection via `manifest.version` monotonic | Cady #4b |
| 7 | Move install-dir rescan to worker (UI thread stutter on slow paths) | Cady #6e |
| 8 | Self-update via `manifest.launcher` field | Cady §2 |
| 9 | Repair / verify per-file sha map | Cady §2 |

The bug fix in `e7e54dc` already has a regression test (`patch_hostname_any_repatches_when_host_changes`) — that's the one Cady asked for in #7c.

---

## CI

`launcher-build.yml` should run on the push and exercise the full Windows pipeline against all 6 commits. With 69 tests (up from 40 in the original PR) and the manifest-signing path exercised by 5 new unit tests, the surface area is meaningfully better-covered.

Watching for: the test suite to pass on `windows-latest`, no clippy regressions, and the workflow itself to handle the new `LAUNCHER_MANIFEST_PUBKEY_HEX` secret being unset gracefully (it should — `option_env!` falls back to the dev pubkey).

Issue #288, opened 2026-05-16
[Critical] Fix world-entry spawn glitches: camera-through-floor + character model load delay


Cadacious:
## Background

Audit: [docs/audits/mercury-rust-conformance-2026-05-15.md](https://github.com/SandboxServers/Cimmeria/blob/main/docs/audits/mercury-rust-conformance-2026-05-15.md) — finding **#4** plus the appearance-cascade sequencing sub-finding documented in §12.

Two user-confirmed observable bugs that share the same world-entry flow:

1. **Camera through floor on initial spawn.** The camera "snaps/moves from (0,0,0) to the new spawn position", visibly clipping through the world floor.
2. **No proper character model load before forced appearance update.** The player entity is rendered with no body model until `BeingAppearance` arrives later in the `mapLoaded` sequence.

## Root cause

**Bug 1 mechanism (finding #4):** `crates/services/src/mercury/world_data/phases::build_enter_world_body` emits `forcedPosition` (msg `0x31`) with the spawn position at offsets 12-23 and zeros at offsets 24-35. The spec confirms (§1.10.6 + §1.16 Q3 closure, anchored to Ghidra `ProcessForcedEntityPosition` at `0x00dd9ee0`) that those 12 bytes are the **previous-position reference vector**, not velocity. The client copies them into `pPrevPos` for the entity. The first frame, the camera-attach code interpolates from prev_pos `(0,0,0)` to current_pos `(spawnX, spawnY, spawnZ)` — the path passes through the world floor.

**Bug 2 mechanism:** `crates/services/src/mercury/world_data/map_loaded::build_map_loaded_body_inner` sends `BeingAppearance` (method 26) as step 15 of ~21 entity-method calls. With the bundle fragmented across ~5 packets, BeingAppearance can land 50-200 ms after `createCellPlayer`. During that window the client has a player entity that exists but has no body model.

## Proposed change

**For bug 1:** in `phases::build_enter_world_body`, replace the `[0.0f32, 0.0, 0.0]` literal at lines ~107-112 with `info.pos` so the previous-position reference equals the spawn position. Interpolation distance becomes zero, no visible movement. Also rename the local from `vel` to `prev_pos` and update the comment to cite §1.10.6 + §1.16 Q3.

Extend `build_forced_position` signature in `crates/services/src/mercury/aoi/update` to accept a `prev_pos: [f32; 3]` parameter (currently hardcodes zeros). Update both callers:
[...]

PR #350, opened 2026-05-23
fix(world-entry): cut reliable TX-window pressure on cold-cache login (closes #345)

### LINKED ISSUE #345: world-entry: reduce reliable TX window pressure from resourceFragment chunking and misplaced chat methods
## Background

PR #337 fixed the immediate TX-window-saturation bug by moving tickSync onto the unreliable counter. This issue tracks the remaining growth surface on the reliable side so the window never gets tight again as the game adds content.

## Current burst anatomy

A cold-cache world-entry login produces the following reliable packets before the window can drain:

| Phase | Source | Packets |
|---|---|---|
| Char-select (ACKed before world-entry) | charList | 1 |
| Cache handshake (mostly ACKed, but can overlap) | versionInfo ×N categories | ~5–10 |
| Cache handshake | resourceFragment chunks for overridden missions | 4 overrides × ceil(xml/1000 B) |
| Terrain-load wait (ACKed during terrain load) | `build_create_player` | 1 |
| **mapLoaded burst (nothing ACKed yet)** | `build_enter_world` standalone | **1** |
| **mapLoaded burst** | mapLoaded fragments | **~3–6 (empty inventory), up to ~9 (100 items)** |

With today's 4 mission overrides at roughly 3 000–5 000 B of XML each, the worst-case concurrent in-flight count (resourceFragments not yet ACKed when mapLoaded fires) is roughly **17–22 packets** — well within the 32-slot cap. But the cap shrinks as we add more mission overrides, and the two items below address the two biggest levers.

---

### Math

`MAX_CHUNK = 1000` bytes in `cooked_data.rs` (three sites: `push_overridden_elements`, `send_category_resources`, `handle_element_data_request`).

Mercury `MAX_BODY_LENGTH` is **1411 bytes**. The per-packet overhead for a resource-fragment body is:

| Fragment position | Overhead fields | Bytes used by overhead | Max XML bytes |
|---|---|---|---|
| First | BASEMSG(1) + WORD_LEN(2) + data_id(2) + chunk_id(1) + frag_flags(1) + msgType(1) + categoryId(4) + elementId(4) | 16 | **1 395** |
| Non-first | BASEMSG(1) + WORD_LEN(2) + data_id(2) + chunk_id(1) + frag_flags(1) | 7 | **1 404** |

At `MAX_CHUNK = 1000` we use **72%** of the available body. Raising to **1 390** (the tighter of the two limits, with 5 bytes safety margin) would use **99.6%**.

### Impact on fragment count

| XML element size | Chunks at 1 000 | Chunks at 1 390 | Savings |
|---|---|---|---|
| 3 000 B | 3 | 3 | 0 |
| 5 000 B | 5 | 4 | -1 |
| 8 000 B | 8 | 6 | -2 |
| 12 000 B | 12 | 9 | -3 |

**Zero savings today** — all 4 current mission overrides produce patched XML smaller than 3 000 B. But every new mission override or large dialog/item PAK element added in future costs `ceil(size/1 000)` packets instead of `ceil(size/1 390)`. The fix is a one-liner change at three constant sites.

### Action

- Add startup logging of per-element XML byte counts so we know when an element crosses chunk boundaries (the `tracing::info!` in `push_overridden_elements` already logs `bytes`; extend `handle_element_data_request` similarly)
- Change `MAX_CHUNK` from `1000` to `1390` at all three sites in `cooked_data.rs`
- Add a test asserting that `build_resource_fragment` with a 1 390-byte chunk fits within `MAX_BODY_LENGTH` (no such guard exists today)

---

### Evidence from the original Python

The source comment in `map_loaded.rs` line 374 already flags this:

```rust
// 24. onChatJoined — notify client about default channels
//     Reference: python/base/SGWPlayer.py onClientReady -> ChannelManager.playerLoggedIn
```

`onClientReady -> ChannelManager.playerLoggedIn` means the original C++/Python server wired chat joins in response to the **onClientReady** base method, not during the mapLoaded entity-method bundle. Our implementation is misplaced.

### Byte savings

| Method | Total bytes |
|---|---|
| onChatJoined × 8 (say, emote, yell, team, squad, command, server, tell) | ~156 B |
| onPlayerCommunication (welcome message) | ~155 B |
| **Total** | **~311 B** |

311 bytes out of a ~2 700-byte body (worst-case test fixture, no items) is unlikely to change fragment count in isolation. But combined with item 3 and inventory growth, it creates real headroom.

### Safety

Moving these to `handle_on_client_ready` introduces a one-RTT window (~100–300 ms) where the player is in the world but chat channels are not yet registered. **This is identical to the original game's behavior** — `ChannelManager.playerLoggedIn` ran on `onClientReady`, not in mapLoaded. If another player sends a chat message during that window, the client drops it silently (no channel registered). Acceptable: same as original.

`onPlayerCommunication` (welcome message) is a cosmetic send with no correctness requirement on timing.

### Action

- Move the 8 `onChatJoined` calls from `build_map_loaded_body_inner` to `handle_on_client_ready`
- Move `onPlayerCommunication` from `build_map_loaded_body_inner` to `handle_on_client_ready`
- Renumber / clean up the step comments in `map_loaded.rs` (steps 24 and 27 become gaps)

---

## Also: extend the fragment-count guard

The existing test `build_map_loaded_fragment_count_fits_within_reliable_tx_window` uses `items: vec![]` and `blueprint_ids: vec![]`. A character with a full main bag (40 items × ~40 B = 1 600 B) and a full mission bag (100 items × ~40 B = 4 000 B) adds ~5 600 B to the body — roughly 4 more fragments. The guard should include a realistic max-inventory fixture so it fails before a regression ships.

## Not in this issue

- Gating mapLoaded until resourceFragment ACKs drain (needs a state-machine change; overkill given current numbers)
- Switching resourceFragments to unreliable (client doesn't retry; the invalidation handshake depends on reliable delivery)

Cadacious:
## Summary

Two independent budget cuts on the cold-cache world-entry burst so it
stays well under Mercury's 32-slot reliable TX window as content grows.

**Item 3 — resourceFragment chunk size 1000 → 1390**
Single shared module constant in [`cooked_data.rs`](crates/services/src/base/cooked_data.rs); chunks pack at 99.6% utilization vs the historical 72%. No fragment savings today (all 4 patched mission overrides are sub-3 kB) but every new PAK override or large dialog element now costs `ceil(size/1390)` packets instead of `ceil(size/1000)`. Three byte-exact guard tests in `mercury::protocol::tests` pin first-fragment / non-first-fragment / production-chunker decrypt-cleanly against `MAX_BODY_LENGTH`.

**Item 4 — onChatJoined ×8 + onPlayerCommunication out of mapLoaded**
Moved into `handle_on_client_ready` to match the original `python/base/SGWPlayer.py onClientReady -> ChannelManager.playerLoggedIn` flow. The Cimmeria placement padded ~311 B onto the worst-case fragment burst. The one-RTT window where the player is in-world without registered chat channels is identical to original game behavior; the welcome message is cosmetic. Regression guard added that mirrors the `onPlayMovie` guard from #288.

**Fragment-count guard hardening**
`build_map_loaded_fragment_count_fits_within_reliable_tx_window` now loads a full-capacity inventory fixture (40 main + 100 mission + 100 crafting + 4 bandolier + 8 armor slots) instead of `items: vec![]`. The original would have let a +5 kB body growth silently slip past until a real player logged in.

## Test plan

- [x] `cargo check -p cimmeria-services`
- [x] `cargo clippy -p cimmeria-services --all-targets -- -D warnings`
- [x] `cargo test -p cimmeria-services --lib` — 779/779 pass
- [x] `cargo fmt --all`
- [x] New tests: 3 resource_fragment size guards + chat-omission guard
- [x] Existing fragment-count guard extended with the full-inventory fixture
- [ ] CI: full workspace clippy + nextest + live-DB

Closes #345

## Release Notes

* **Improvements**
  * Chat channel registration and welcome messages now occur earlier during player connection establishment, ensuring chat functionality is available sooner for connecting players.

* **Tests**
  * Added regression tests to validate chat message delivery and wire protocol compliance during client initialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/350?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/services/src/base/cooked_data.rs
+++ b/crates/services/src/base/cooked_data.rs
@@ -362,7 +362,7 @@ pub(crate) async fn handle_element_data_request(
     // INFO matches `push_overridden_elements`: every patched-XML push is
     // load-bearing for cache consistency, and the per-element byte count
     // is what surfaces a future PAK element crossing a chunk boundary
-    // and adding fragments to the cold-cache login burst (issue #345).
+    // and adding fragments to the cold-cache login burst.
     if is_override {
         tracing::info!(
             %addr, category_id, element_id,
@@ -414,3 +414,21 @@ pub(crate) async fn handle_element_data_request(
 
     Ok(())
 }
+
+/// Compile-time lower-bound guard for `MAX_CHUNK`. The wire-level tests in
+/// `mercury::protocol::tests` decrypt a resource fragment at the current
+/// `MAX_CHUNK` and assert it fits Mercury's body cap — but they'd happily
+/// pass at `MAX_CHUNK = 1000` too, since 1000-byte chunks also fit. This
+/// const assertion pins the lower bound at compile time: a future revert
+/// that drops `MAX_CHUNK` below 1390 fails to build instead of silently
+/// restoring the wasted ~28% per-packet headroom and re-widening the
+/// cold-cache login burst against the 32-slot reliable TX window.
+///
+/// Bump in tandem if a deliberate retreat from 1390 is ever needed (and
+/// document the reason in `MAX_CHUNK`'s docstring above).
+const _: () = assert!(
+    MAX_CHUNK >= 1390,
+    "MAX_CHUNK dropped below the 1390 floor — see the doc comment above \
+     for the cold-cache burst rationale; update the docstring + this \
+     guard together if the retreat is intentional."
+);
--- a/crates/services/src/base/cooked_data.rs
+++ b/crates/services/src/base/cooked_data.rs
@@ -13,6 +13,24 @@ use super::helpers::{drain_acks_and_seq, get_active_entity_id};
 use super::resources::{CategoryData, ResourceCache};
 use super::ConnectedClientState;
 
+/// Maximum XML bytes per `BASEMSG_RESOURCE_FRAGMENT` packet.
+///
+/// Mercury `MAX_BODY_LENGTH` is **1411 bytes**. The first fragment of a
+/// resource transfer carries 16 bytes of overhead inside the body — that
+/// is, `BASEMSG`(1) plus `WORD_LEN`(2) plus `data_id`(2) plus
+/// `chunk_id`(1) plus `frag_flags`(1) plus `msg_type`(1) plus
+/// `category_id`(4) plus `element_id`(4). Non-first fragments carry only
+/// 7 bytes (`BASEMSG` + `WORD_LEN` + `data_id` + `chunk_id` + `frag_flags`).
+///
+/// Picking 1390 leaves a 5-byte safety margin against the tighter
+/// first-fragment cap (1395) and packs the XML body at 99.6% utilization;
+/// the historical value of 1000 wasted ~28% of every packet.
+///
+/// The fragment-size guard test in `mercury::protocol::tests` pins this
+/// against `MercuryEncryption::decrypt(...)` for both first and non-first
+/// fragment shapes; bump in tandem if the wire-format overhead changes.
+const MAX_CHUNK: usize = 1390;
+
 /// Handle `versionInfoRequest` (0xC0).
 ///
 /// Client payload: [categoryId: u32][version: u32]
@@ -145,8 +163,6 @@ async fn push_overridden_elements(
     cache: &ResourceCache,
     connected: &Arc<Mutex<HashMap<SocketAddr, ConnectedClientState>>>,
 ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    const MAX_CHUNK: usize = 1000;
-
     for &element_id in element_ids {
         let xml_data = match cache.get(category_id, element_id) {
             Some(data) => data,
@@ -225,8 +241,6 @@ pub(crate) async fn send_category_resources(
     cat: &CategoryData,
     connected: &Arc<Mutex<HashMap<SocketAddr, ConnectedClientState>>>,
 ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    const MAX_CHUNK: usize = 1000;
-
     // Sort element IDs for deterministic ordering.
     let mut element_ids: Vec<u32> = cat.elements.keys().copied().collect();
     element_ids.sort();
@@ -326,18 +340,9 @@ pub(crate) async fn handle_element_data_request(
         }
     };
 
-    // Promote the log to INFO when we're shipping a Cimmeria-overridden
-    // element — operationally the load-bearing question is "did the
-    // patched mission XML actually go out to this client", and that
-    // answer is too important to bury at debug level on a noisy DB.
+    // Override status drives log level below: every patched-XML push is
+    // load-bearing for cache consistency, so it deserves INFO. Reads only.
     let is_override = cache.overridden_elements(category_id).contains(&element_id);
-    if is_override {
-        tracing::info!(
-            %addr, category_id, element_id,
-            bytes = xml_data.len(),
-            "Shipping overridden element to client"
-        );
-    }
 
     // Allocate a data_id for this transfer
     let data_id = {
@@ -351,19 +356,31 @@ pub(crate) async fn handle_element_data_request(
         }
     };
 
-    // Fragment into 1000-byte chunks
-    const MAX_CHUNK: usize = 1000;
     let chunks: Vec<&[u8]> = xml_data.chunks(MAX_CHUNK).collect();
     let total_chunks = chunks.len();
 
-    tracing::debug!(
-        %addr,
-        element_id,
-        total_size = xml_data.len(),
-        total_chunks,
-        data_id,
-        "Fragmenting resource data"
-    );
+    // INFO matches `push_overridden_elements`: every patched-XML push is
+    // load-bearing for cache consistency, and the per-element byte count
+    // is what surfaces a future PAK element crossing a chunk boundary
+    // and adding fragments to the cold-cache login burst (issue #345).
+    if is_override {
+        tracing::info!(
+            %addr, category_id, element_id,
+            bytes = xml_data.len(),
+            total_chunks,
+            data_id,
+            "Fragmenting overridden resource data"
+        );
+    } else {
+        tracing::debug!(
+            %addr,
+            element_id,
+            total_size = xml_data.len(),
+            total_chunks,
+            data_id,
+            "Fragmenting resource data"
+        );
+    }
 
     for (i, chunk) in chunks.iter().enumerate() {
         let frag_flags = match (i == 0, i == total_chunks - 1) {
--- a/crates/services/src/base/mod.rs
+++ b/crates/services/src/base/mod.rs
@@ -32,6 +32,7 @@ mod service;
 pub(crate) mod tick_sync;
 pub(crate) mod world_entry;
 pub(crate) mod world_entry_appearance;
+pub(crate) mod world_entry_chat;
 
 #[cfg(test)]
 mod smoke_tests;
--- a/crates/services/src/base/world_entry_appearance.rs
+++ b/crates/services/src/base/world_entry_appearance.rs
@@ -133,11 +133,17 @@ pub(crate) async fn handle_on_client_ready(
         }
     }
 
-    let pending = {
+    // Take the pending finalization AND copy out the player_name in the
+    // same lock so the welcome-message send below doesn't need a second
+    // round-trip. `player_name` is set during `playCharacter` and stays
+    // for the session, so reading it here is safe.
+    let (pending, player_name) = {
         let mut clients = connected.lock().map_err(|_| "connected lock poisoned")?;
-        clients
-            .get_mut(&addr)
-            .and_then(|c| c.pending_client_ready.take())
+        let entry = clients.get_mut(&addr);
+        match entry {
+            Some(c) => (c.pending_client_ready.take(), c.player_name.clone()),
+            None => (None, None),
+        }
     };
 
     let Some(pending) = pending else {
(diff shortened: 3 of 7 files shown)


MyPalClara:

---

### G1: `MAX_CHUNK = 1390` safety margin is tight
**Location:** `crates/services/src/mercury/cooked_data.rs`
**Severity:** Low

The 5-byte margin is tight. First-fragment overhead is 16 bytes, so max chunk is 1395. You chose 1390 for a 5-byte margin. If anyone adds a single field to the first-fragment header (e.g., a version byte), the margin evaporates.

**The doc comment explains this well**, but consider whether 10 bytes would be safer.

**Recommendation:**
```rust
/// Production chunk size: 1385 bytes (10-byte safety margin)
/// This gives headroom for future header fields without breaking the invariant.
pub const MAX_CHUNK: usize = 1385;
```

**Effort:** 2 minutes (if desired)

---

### G2: `handle_on_client_ready` player_name fallback
**Location:** `crates/services/src/base/world_entry/methods/on_client_ready.rs`
**Severity:** Low

If `player_name` is `None`, the welcome message is skipped entirely. The player sees nothing — no welcome, no player ID confirmation.

**Recommendation:** Add a fallback message:
```rust
let welcome_msg = match &state.player_name {
    Some(name) => format!("Welcome to Stargate Worlds. Your player id is: {}.", state.entity_id),
    None => "Welcome to Stargate Worlds.".to_string(),
};
// Always send the welcome, even without player_name
send_player_communication(&state, "System", &welcome_msg).await;
```

Or log at `warn!` level so the missing name is visible in production logs.

**Effort:** 10 minutes

---

### G3: Chat channel registration — 8 sequential reliable sends
**Location:** `crates/services/src/base/world_entry/methods/on_client_ready.rs`
**Severity:** Low (observation)

8 sequential `send_to_witness_reliable` calls means 8 lock acquisitions in sequence. For 8 channels, that's 8 round-trips through the lock → build → encrypt → send path.

**Not a bug**, but if this becomes a hot path (e.g., mass login event), consider batching into a single entity-method bundle:

```rust
// Future optimization: batch all 8 onChatJoined into one entity-method bundle
let chat_methods = channels.iter()
    .map(|ch| build_on_chat_joined_record(ch))
    .collect();
send_entity_method_bundle(&state, chat_methods).await;
```

**Effort:** 1-2 hours (if needed later)

---

**Category 1 Total Effort:** ~15 minutes for G1+G2, G3 is future optimization

MyPalClara:

---

### G4: `onClientReady` timing window verification
**Location:** `crates/services/src/base/world_entry/methods/on_client_ready.rs`
**Severity:** Low

The PR body says "the one-RTT window where the player is in-world without registered chat channels is identical to original game behavior."

The original Python flow was:
1. `mapLoaded()` completes
2. Client sends `onClientReady`
3. Server receives `onClientReady`, calls `ChannelManager.playerLoggedIn`
4. Chat channels registered

Cimmeria's flow appears identical, but this should be verified against the Python source to confirm the timing matches exactly.

**Recommendation:** Add a comment citing the Python source:
```rust
/// Chat registration moved here to match original Python flow:
/// SGWPlayer.py:onClientReady() -> ChannelManager.playerLoggedIn()
/// This creates a one-RTT window where player is in-world without chat,
/// identical to the 2009 client behavior.
```

**Effort:** 5 minutes

---

### G5: Full-inventory fixture magic numbers
**Location:** `crates/services/src/mercury/world_data/tests/map_loaded.rs`
**Severity:** Low (maintainability)

The slot counts are hardcoded:
```rust
fill_bag(&mut items, INV_MAIN, 40);
fill_bag(&mut items, INV_MISSION, 100);
fill_bag(&mut items, INV_CRAFTING, 100);
fill_bag(&mut items, INV_BANDOLIER, 4);
// ... 8 armor slots
```

If `python/common/Constants.py` changes these values, the test won't catch the drift.

**Recommendation:** Either import the constants from a shared source, or add a comment citing the source:
```rust
// Slot counts from python/common/Constants.py:
// INV_MAIN_SLOTS = 40, INV_MISSION_SLOTS = 100, INV_CRAFTING_SLOTS = 100
// INV_BANDOLIER_SLOTS = 4, EQUIPMENT_SLOTS = 8 (head/face/neck/chest/hands/waist/legs/feet)
fill_bag(&mut items, INV_MAIN, 40);
```

**Effort:** 5 minutes

---

**Category 2 Total Effort:** ~10 minutes

MyPalClara:
## Category 3: Missing Test Coverage (G6-G8)

---

### G6: No positive test for `onChatJoined` in `handle_on_client_ready`
**Location:** `crates/services/src/base/world_entry/methods/on_client_ready.rs`
**Severity:** Low-Medium

The test `build_map_loaded_omits_chat_joined_and_player_communication_from_bundle` verifies chat is NOT in mapLoaded, but there's no test verifying it IS in onClientReady.

A future refactor could accidentally delete the chat registration entirely, and the existing test would still pass.

**Recommendation:** Add a positive test:
```rust
#[tokio::test]
async fn handle_on_client_ready_sends_chat_joined_for_all_channels() {
    let (state, rx) = make_on_client_ready_test_state();
    
    handle_on_client_ready(&state).await;
    
    let packets: Vec<_> = rx.try_iter().collect();
    let chat_joined_count = packets.iter()
        .filter(|p| is_on_chat_joined_method(p))
        .count();
    
    assert_eq!(chat_joined_count, 8, 
        "Expected 8 onChatJoined calls (one per channel)");
}
```

**Effort:** 30 minutes

---

### G7: No test for welcome message content
**Location:** `crates/services/src/base/world_entry/methods/on_client_ready.rs`
**Severity:** Low

The welcome message format is `"Welcome to Stargate Worlds. Your player id is: {entity_id}."` — this matches Python, but there's no test pinning the exact format.

**Recommendation:** Add a content test:
```rust
#[tokio::test]
async fn handle_on_client_ready_sends_welcome_message_with_player_id() {
    let (state, rx) = make_on_client_ready_test_state();
    state.entity_id = 12345;
    
    handle_on_client_ready(&state).await;
    
    let welcome = find_player_communication_packet(&rx);
    assert!(welcome.contains("Welcome to Stargate Worlds"));
    assert!(welcome.contains("12345"));
}
```

**Effort:** 20 minutes

---

### G8: No test for `MAX_CHUNK` constant value
**Location:** `crates/services/src/mercury/cooked_data.rs`
**Severity:** Low

The wire-level tests verify the chunk FITS, but don't verify `MAX_CHUNK == 1390`. If someone changes `MAX_CHUNK` to 1000 (reverting the fix), the tests still pass — they just test a smaller chunk.

**Recommendation:** Add a constant-value guard:
```rust
#[test]
fn max_chunk_is_at_least_1390_for_tx_window_budget() {
    // This test guards against accidental reversion of the chunk-size optimization.
    // See PR #350 for the TX window budget analysis.
    assert!(
        MAX_CHUNK >= 1390,
        "MAX_CHUNK was reduced below 1390; this will increase TX window pressure. \
         If intentional, update this test and document the reason."
    );
}
```

**Effort:** 5 minutes

---

**Category 3 Total Effort:** ~55 minutes

MyPalClara:
## Category 4: Documentation (G9) + Summary

---

### G9: No spec update for chunk-size budget
**Location:** `docs/drafts/spec/mercury-wire-format.md` §1.7
**Severity:** Low

The 32-slot TX window constraint and the chunk-size budget should be in the Bible. Currently this knowledge lives only in the `MAX_CHUNK` doc comment and PR #350 body.

**Recommendation:** Add to `mercury-wire-format.md` §1.7:
```markdown
### Resource Fragment Sizing

| Constant | Value | Source |
|----------|-------|--------|
| MAX_BODY_LENGTH | 1411 bytes | Mercury plaintext body cap |
| First-fragment overhead | 16 bytes | BASEMSG + WORD_LEN + data_id + chunk_id + frag_flags + msg_type + category_id + element_id |
| Non-first overhead | 7 bytes | No msg_type/category_id/element_id |
| Production MAX_CHUNK | 1390 bytes | 1395 - 5 byte safety margin |

**Rationale:** At 1390 bytes per chunk vs the historical 1000 bytes, 
utilization improves from 70.9% to 98.5%. This reduces the number of 
reliable packets in the world-entry burst, protecting the 32-slot TX window 
budget (see §1.7 R9).

**Constraint:** Any change to first-fragment header fields MUST update 
MAX_CHUNK to maintain the 5-byte safety margin.
```

**Effort:** 15 minutes

---

## Summary Table

| Category | Gaps | Severity | Effort |
|----------|------|----------|--------|
| Correctness | G1-G3 | Low | 15 min |
| Potential Issues | G4-G5 | Low | 10 min |
| Test Coverage | G6-G8 | Low-Medium | 55 min |
| Documentation | G9 | Low | 15 min |
| **Total** | **9** | | **~1.5 hours** |

---

## Priority Matrix

| Priority | Gaps | Action |
|----------|------|--------|
| 🟢 Quick wins | G5, G8, G9 | Source citations + constant guard + spec update |
| 🟡 Follow-up | G6 | Positive test for chat in onClientReady |
| ⚪ Nice-to-have | G1, G2, G4, G7 | Margin increase, fallback message, timing comment, content test |
| ⚪ Future | G3 | Batch optimization if needed |

Cadacious:
Thanks for the thorough pass. Disposition for the two review tracks:

### Copilot inline (×12) — all addressed

| Comment | Action |
|---|---|
| 6× source-comment "issue #345" references | Scrubbed all 9 instances; rationale stays, citation moves to PR body per `.github/copilot-instructions.md:44` |
| Inventory fixture missing INV_BACK / ARTIFACT1 / ARTIFACT2 | Added; doc reconciled to "11 equipment slots" with the full slot list inline |
| Hardcoded indices 28/31 in chat-omission guard | Replaced with `method_idx::ON_PLAYER_COMMUNICATION` / `method_idx::ON_CHAT_JOINED` |
| "five wearables" but 8 listed | Fixed in same edit as the equipment-slot addition |
| `resource_fragment` tests compute body_len from assumed sizes | Extracted `measure_resource_fragment_body` that reads the u16 LE WORD_LEN prefix from the decrypted body; all 3 guards now assert against ground truth, not the spec interpretation. A future header overhead change in `build_resource_fragment` will break these tests. |

### MyPalClara categorised review (G1–G9)

| Gap | Action |
|---|---|
| **G1** 5-byte margin tight | **Kept at 1390.** The 5-byte margin is the documented invariant, pinned by `_at_cooked_data_max_chunk_has_safety_margin` (`>= 5`) and the new compile-time `const _: () = assert!(MAX_CHUNK >= 1390)`. Bumping to 1385 would silently overshoot the documented invariant. |
| **G2** player_name=None silent skip | **Fixed.** Falls back to a "Server" speaker so the welcome still lands; warn log still fires. Test pins the fallback string. |
| **G3** 8 sequential reliable sends | **Deferred.** Not a bug; current world-entry rate doesn't make this hot. |
| **G4** Python source citation | **Already present** at the chat-join loop site (`SGWPlayer.py onClientReady -> ChannelManager.playerLoggedIn`). |
| **G5** Magic numbers in fixture | Doc already cites `python/common/Constants.py`. |
| **G6** Positive test for chat in onClientReady | **Fixed via helper extraction** — `DEFAULT_CHAT_CHANNELS` slice pinned byte-exact; `build_chat_joined_args` swept over the full default set. |
| **G7** Welcome message content test | **Fixed via helper extraction** — `build_welcome_message_args` byte-exact + entity_id substitution + fallback speaker tests. |
| **G8** `MAX_CHUNK >= 1390` constant guard | **Fixed.** Compile-time `const _: () = assert!(...)` next to the constant; revert below 1390 fails to build, not at test time. |
| **G9** Spec doc update | **Deferred.** Source-side `MAX_CHUNK` docstring captures the budget; a docs/drafts/spec/ update would be a separate PR. |

### Codecov

The 54 missing lines in `world_entry_appearance.rs` were the inline chat-join + welcome block. Pulled into pure helpers (`build_chat_joined_args` / `build_welcome_message_args`) so the byte-building logic is covered by 6 new unit tests; the remaining uncovered region in `handle_on_client_ready` is the wire-side `send_to_witness_reliable` loop that needs the async harness G3 also flags.

### Validation

- `cargo fmt --all` clean
- `cargo clippy -p cimmeria-services --all-targets -- -D warnings` clean
- `cargo test -p cimmeria-services --lib` — 785/785 pass (6 new + 779 existing)

Issue #360, opened 2026-05-23
Mercury Bundle migration: deferred call families (follow-up to #356)


Cadacious:
## Context

Issue #356 introduced the `ChannelBundle` accumulator + cross-entity bundling rule and migrated the **AoI EnteredAoI burst** as a conservative first slice. That work landed the abstraction, helper, ADR, and the byte-equivalence + burst-shape regression guards.

The original #356 scope sketched migrating ~41 call sites across a dozen files. Doing all of that in one PR risks both review fatigue AND a missed transaction-state hazard (see the [ADR](docs/architecture/mercury-bundle.md) for the "one bundle == one client frame" rule and why a same-entity-after-CREATE bundle silently drops messages). This issue tracks the deferred families so each can ship as its own reviewable PR.

## Migration playbook

Follow [docs/architecture/mercury-bundle.md § Migration playbook for follow-up call families](docs/architecture/mercury-bundle.md). For each family below:

1. Audit the transaction-state surface — list every entity touched, flag CREATE_ENTITY paired with same-entity follow-ups.
2. Extract body-only `compose_*` from the existing `build_*` packet builder.
3. Add a byte-equivalence regression guard (`compose_*` vs decrypted `build_*` body).
4. Migrate the caller(s) to `ChannelBundle` + `send_bundle_to_witness_reliable`.
5. Add a burst-shape regression guard asserting the new packet count.

## Deferred call families

Counts are `send_to_witness_reliable | append_entity_method | shadow_register_reliable_send` occurrences in each file as of the merge of #356.

### High-value (burst shape, biggest TX-window win)

- [ ] **`crates/services/src/base/world_entry_appearance.rs`** (9 calls) — appearance burst at world entry. Likely has CREATE_ENTITY adjacency hazard with `BeingAppearance`, mirroring the map_loaded split. Audit carefully.
- [ ] **`crates/services/src/mercury/world_data/phases.rs`** (8 calls) — world-entry payload phases. Has the manual fragmentation `map_loaded.rs` deliberately keeps as a separate bundle. Migration must preserve the existing bundle boundary (CELL_PLAYER + same-player methods stay split).
- [ ] **`crates/services/src/base/world_entry/methods/progression/mod.rs`** (7 calls) — level-up / XP burst. Multiple property updates fire together; safe to bundle if all target the player entity AFTER any CREATE_ENTITY committed.

### Medium-value (per-handler latency-tolerant)

[...]

PR #357, opened 2026-05-23
feat(mercury): deferred-send queue + defer cell creates until onClientReady (#354 fixes 2+3)

Cadacious:
## Summary

Two of the three #354 interim fixes for Mercury TX-window overflow. Sibling fix 1 (Bundle abstraction) was discovered to be 5× the original scope estimate and has been re-scoped as its own issue (#356) instead of bundling into this PR.

- **Fix 3 — deferred-send queue:** replaces the "downgrade reliable send to best-effort" path with a per-channel queue that holds overflow entries until ACKs free TX-window slots. Closes the silent-reliability-loss correctness bug that broke the `debug/lomiada-broke-in-hallway02/` session.
- **Fix 2 — defer cell-driven AoI until `onClientReady`:** buffers cell→base AoI dispatches while the client is still in the pre-`onClientReady` world-entry window, then drains the buffer when the client signals ready.

Combined, the silent-best-effort downgrade is eliminated from the world-entry path entirely. See the Commits tab for the two-commit split (one per fix).

## The bug shape (from `debug/lomiada-broke-in-hallway02/`)

| | Before this PR | After fix 3 alone | After fixes 3 + 2 (this PR) |
|---|---|---|---|
| World-entry burst hits TX window | ~17 reliable packets silently downgraded to best-effort | Overflow queued + retransmittable; zero silent loss | Cell AoI portion deferred until ready; smaller initial burst; queue absorbs flush burst |
| Lost overflow packet | No recovery — silent state corruption | Retransmits via deferred queue | Retransmits via deferred queue |
| Cell AoI fires during 3.5s pre-`mapLoaded` window | Backlog accumulates in TX window | Backlog accumulates in TX window | Buffered out-of-band; doesn't pressure window |

Tested against the lomiada session logs concretely:

- **Server log** `debug/lomiada-broke-in-hallway02/server-logs.txt:142` — first TX-window-full warning, 3 ms after `mapLoaded`
- **Server log** `debug/lomiada-broke-in-hallway02/server-logs.txt:887-888` — TX-window-full at the exact instant of `onBeginAidWait` (Δ = 464μs)
- **Client log** `debug/lomiada-broke-in-hallway02/2026-05-23_13-27.log:2074-2510` — 210 packets buffered waiting for the missing seq 1148, then `disconnectClientArgs`

## What's NOT in this PR

[...]

Issue #354, opened 2026-05-23
Mercury TX-window relief: Bundle + defer-creates + deferred-send queue (one PR)


Cadacious:
## TL;DR

Three server-side fixes that interlock to eliminate (or recover from) the TX-window overflow we observe at every world-entry and every player defeat. All three land in **one PR** because they share a code surface (the reliable-send path in `crates/services/src/base/helpers.rs` and the `Channel` abstraction in `crates/mercury`) and because they only collectively close the bug class — any one in isolation leaves visible failure modes.

| # | Fix | Failure mode it addresses |
|---|---|---|
| 1 | Mercury **Bundle** abstraction | World-entry / defeat bursts overrun the 32-slot TX window because every entity method is its own packet |
| 2 | **Defer** cell-driven AoI creates until `onClientReady` | 33-packet backlog accumulates during the 3.5s mapLoaded terrain-load gap while the client cannot ACK |
| 3 | Replace overflow **downgrade** with a deferred-send queue | When the window does fill, the current path silently un-tracks reliable packets (no retransmit, no ACK accounting) — a correctness bug |

Sibling to #353 (long-term: patch the client to widen the window to 64). The three fixes here are valuable independently of the client patch — Bundle and defer reduce traffic, the deferred-send queue is a correctness floor that should exist regardless of window size.

## Background

`debug/lomiada-broke-in-hallway02/` captures the failure shape concretely. Two sessions of the same player, both ending the same way: client goes silent → server's 60s inactivity timer fires → session dropped.

Two distinct overflow bursts in every session:

- **World-entry burst** at `server-logs.txt:142` (session 1) and `:960` (session 2): 19 reliable packets downgraded to best-effort within 3ms of `mapLoaded`. Cause: server fires `BeingAppearance`, entity creates, prewarm visuals, etc. as separate reliable methods with no batching.
- **Pre-mapLoaded backlog**: server's seq counter advances 30 → 63 over 3.5s while the client loads terrain (cannot ACK). When mapLoaded finally fires, the TX window is already mostly full before the burst even starts. Cause: cell-driven AoI creates fire during the terrain-load window.
- **Defeat burst** at `server-logs.txt:887-888`: TX window full at the exact instant of `onBeginAidWait` (Δ = 464μs). Cause: defeat state-change burst (Defeat Window UI, BSF flag changes, animation triggers) overflows by the same mechanism as world-entry.

[...]

Issue #353, opened 2026-05-23
Widen Mercury TX window from 32 → 64 via SGW.exe binary patch


Cadacious:
## TL;DR

Our server's per-channel Mercury TX window is pinned at 32 packets. Every player world-entry overflows it in milliseconds (19+ reliable methods, plus a 33-packet backlog accumulated during the 3.5-second terrain load when the client cannot ACK), silently downgrading critical packets to best-effort. The same overflow fires at every player defeat. Lost packets mean lost appearance state, lost combat-state flags, and silent session corruption — see the `debug/lomiada-broke-in-hallway02/` session as a concrete failure.

The 32 cap was based on an **incorrect understanding of the SGW client's reliability layer**. Ghidra-verified disassembly proves the client has no 32-bit ACK bitmap; the slot store is a heap-allocated hash table with runtime-configurable capacity derived from a value the constructor reads at `[config+0x2C]`. **A 3-byte patch to `SGW.exe` at VA `0x0158C801` makes the client honor a 64-slot window** with zero protocol or wire-format changes.

This issue tracks the full work: client patch, server constant bump, launcher integration, dev-mode tooling, doc/spec/figure corrections, and verification.

### The failure mode

`debug/lomiada-broke-in-hallway02/` captures two back-to-back sessions of the same player. Both ended the same way: client went silent → server's 60s inactivity timer fired → session dropped. The proximate trigger in each case was a **TX-window-full burst** that downgraded ~19+ reliable packets to best-effort:

| Session | When | Burst seq range | Packets downgraded |
|---|---|---|---|
| 1 (no debug log) | 11:15:42 (3 ms after `mapLoaded`) | seq=63..81 | 19 |
| 1 (no debug log) | 11:27:35 (instant of player defeat) | seq=911 | 1 (then silence) |
| 2 (with debug log) | 11:29:01 (1 ms after `mapLoaded`) | seq=65..83 | 19 |
| 2 (with debug log) | 11:35:20 → 11:35:49 | seq=1391..1420 | 30 (1/sec for 30s after silence) |

Every world-entry triggers this. Every death triggers this. In a healthy network it's masked by the fact that UDP-best-effort usually delivers — but the moment one of those downgraded packets is lost, it's gone forever, and the player loses an appearance component or a defeat-window state flag with no recovery.

Concrete evidence from the session logs:
[...]

PR #361, opened 2026-05-23
feat(mercury): ChannelBundle accumulator + AoI burst migration (#356)

### LINKED ISSUE #356: Mercury Bundle abstraction: batch entity methods per packet
## TL;DR

Wire BigWorld's `Bundle` abstraction into our Mercury send path so multiple application-level messages can be packed into a single UDP packet, rather than one packet per method as today. This is the largest single piece of TX-window relief but the biggest scope of the three #354 interim fixes — broken out into its own issue because the original 1-2 day estimate was off by an order of magnitude.

**Sibling to #354.** Tracks the work that was originally scoped as "fix #1" in the umbrella issue. The other two fixes (deferred-send queue, defer-cell-creates-until-onClientReady) shipped as their own commits on the #354 branch.

## What I thought when filing #354

> #354 fix #1: Bundle abstraction — One Bundle per Channel, accumulate via `start_message`/`end_message`, finalize on tick boundary into ≤1348-byte packets. Migrate ~30 `send_to_witness_reliable` call sites. ~1-2 weeks of work.

## What I actually found on opening the work

1. **The existing `crates/mercury/src/bundle.rs` is a stub** — 209 lines, only referenced from its own tests, with a wire format of `[u8 msg_id][u16 LE len][payload]` repeated. This does NOT match SGW's actual entity-method wire format. Not reusable as the production batching path.

2. **The real wire-format encoder lives elsewhere** — `crates/services/src/mercury/mod.rs:246` (`append_entity_method`) correctly handles the SGW direct-vs-extended encoding:
   - Direct (method_index 0-60): `[(index | 0x80): u8][word_len: u16][entity_id: u32][args...]`
   - Extended (method_index ≥ 61): `[0xBD: u8][word_len: u16][entity_id: u32][(index - 61): u8][args...]`

3. **`append_entity_method` is called 41 times across 11 production files** — but each call site builds its OWN self-contained packet body, then sends as one packet. There is no accumulator that batches methods across call sites.

4. **Each send site allocates its own seq + ACKs independently** — via [`drain_acks_and_seq`](crates/services/src/base/helpers.rs#L173) (or equivalent) at call time. Bundle semantics need a single seq + ACK piggyback per *finalized packet*, not per *message*.

5. **The world-entry path manually composes multi-packet fragmented bundles** — `crates/services/src/base/world_entry/map_loaded.rs:80-142` builds a body via `build_map_loaded_body` then explicitly fragments it via `fragment_map_loaded`. That manual fragmentation is what Bundle's `finalize` would subsume.

## Scope (revised)

Estimated **1-2 weeks of focused work**, in three layers:

### Layer A — Bundle accumulator + Channel integration

- Decide: extend the existing `bundle.rs` (rewrite the wire format) or build a new module (`channel_bundle.rs`) alongside. **Recommend: new module**, leave the existing `bundle.rs` for whatever original intent it had until that's traced.
- New type: `ChannelBundle` with:
  - `start_entity_method(method_index: u16, entity_id: u32) -> BundleWriter` (returns a writer that appends args)
  - `start_base_message(msg_id: u8) -> BundleWriter` (for non-entity Mercury messages — e.g. CREATE_BASE_PLAYER, RESET_ENTITIES)
  - `end_message()` (or RAII via writer Drop)
  - `is_empty() -> bool`
  - `finalize(reliable: bool) -> Vec<Packet>` returning N packets ≤ `MAX_BODY = 1348` body bytes each
  - Carries `acks: Vec<u32>` so piggybacked ACKs ride with the first finalized packet only
- Channel ownership: one `ChannelBundle` per `Channel`, recycled each tick. Access via `Channel::current_bundle() -> &mut ChannelBundle`.
- Tick driver flush: `Channel::flush_bundle()` called from `crates/services/src/base/tick_sync.rs` per session per tick. Finalizes the bundle, stamps seqs from `next_seq`, mirrors into TX window via the deferred-queue-aware [`register_sent_packet`](crates/mercury/src/channel/mod.rs#L239), hands encrypted packets to the socket.
- Tick-boundary semantics: messages buffer until next flush (every 100ms per `tick_sync`). This is a behavior change from today's at-call sends. Audit any latency-sensitive paths (combat dispatch, damage apply) before flipping them.

### Layer B — Call-site migration

Sweep via `git grep 'send_to_witness_reliable\|shadow_register_reliable_send'`. Each site changes from "build packet, send, register" to "append to channel.current_bundle()".

Known call families (touch counts approximate, confirm via grep):

| Family | Files | Count |
|---|---|---|
| AoI handlers | `crates/services/src/base/world_entry/cell_dispatch/aoi.rs` | 6 |
| World-entry payload | `crates/services/src/base/world_entry/{play_character,enable_entities,map_loaded,reanchor_player,gate_travel/mod}.rs` | 5 |
| World-entry appearance | `crates/services/src/base/world_entry_appearance.rs` | several |
| Inventory / vendor / mail handlers under `crates/services/src/base/world_entry/methods/` | various | ~10+ |
| Cooked-data response | `crates/services/src/base/cooked_data.rs` | 1 |
| Cell-driven entity-method dispatch | indirectly via cell→base messages | already lands at `cell_dispatch/aoi.rs` |

Each migration ≈ 5-20 lines. Tests for individual sites are mostly wire-format byte-pattern tests that will pass unchanged (because the bytes inside the body don't change — just which packet the body ends up in).

### Layer C — Tests + invariants

- New test file `crates/mercury/src/channel/tests/bundle.rs`:
  - `bundle_packs_multiple_messages_into_one_packet`
  - `bundle_fragments_at_1348_byte_boundary`
  - `bundle_finalize_acks_ride_on_first_packet_only`
  - `bundle_empty_finalize_emits_zero_packets`
  - `bundle_recycled_per_tick`
- World-entry integration test: assert that the world-entry burst now finalizes into ≤3 packets (was 5-6).
- Defeat-burst integration test: assert that the BSF state-change burst finalizes into ≤2 packets.
- AoI burst test: 28 NPC CREATE_ENTITY events at world entry pack into ≤3 packets (was 28).

## Why not just ship #353 and call it done

The client patch in #353 widens the TX window from 32 → 64 slots. With #353 alone, the world-entry burst still fits because 19+ packets < 64. So the immediate bug goes away.

But Bundle is independently valuable even with a 64-slot window:

- **Lower packet count = lower per-packet overhead** on the wire (each UDP packet has ~28 bytes of IP+UDP headers; collapsing 19 packets into 3 saves ~450 bytes of overhead per world entry).
- **Fewer ACKs to piggyback** — the client's UnAckedHandler tracks one slot per packet sent, so fewer packets = fewer slots consumed = bigger effective window.
- **Matches BigWorld's intended design** — the C++ Bundle docstring says verbatim *"improve network efficiency of the BW protocol by bundling packets together"* (`deprecated/cpp/src/mercury/bundle.hpp:10-14`). Diverging from this design has accumulated drift; converging back removes the drift.
- **Reduces per-tick processing** on both sides — fewer Mercury packets to encode/decode/dispatch.

## Implementation order within the eventual PR

1. **Layer A first** — build the abstraction in isolation with tests. No call-site changes yet. Lands a new module that nothing uses.
2. **Layer B by family** — migrate one call family at a time, smallest first (cooked_data, world-entry-payload), commit each. AoI family last (largest, most subtle).
3. **Layer C** — integration tests at the end, asserting the burst sizes have actually shrunk.

## File touch map

```
crates/mercury/src/channel/bundle.rs                            # NEW (or rename existing bundle.rs)
crates/mercury/src/channel/mod.rs                               # ChannelBundle field + accessor
crates/mercury/src/channel/tests/bundle.rs                      # NEW
crates/services/src/base/tick_sync.rs                           # flush call per tick
crates/services/src/base/helpers.rs                             # send_to_witness_reliable becomes bundle-append
crates/services/src/base/world_entry/cell_dispatch/aoi.rs       # migrate 6 handlers
crates/services/src/base/world_entry_appearance.rs              # migrate appearance burst
crates/services/src/base/world_entry/play_character.rs          # migrate
crates/services/src/base/world_entry/enable_entities.rs         # migrate
crates/services/src/base/world_entry/map_loaded.rs              # migrate + drop manual fragmentation
crates/services/src/base/world_entry/reanchor_player.rs         # migrate
crates/services/src/base/world_entry/gate_travel/mod.rs         # migrate
crates/services/src/base/world_entry/methods/inventory/**       # migrate ~5 sites
crates/services/src/base/world_entry/methods/vendor/**          # migrate ~3 sites
crates/services/src/base/world_entry/methods/mail.rs            # migrate
crates/services/src/base/cooked_data.rs                         # migrate
docs/architecture/2026-05-23-mercury-bundle.md                  # NEW ADR
docs/drafts/spec/mercury-wire-format.md                         # add Bundle section
```

## Risks & open questions

- **Tick-boundary latency**: today's at-call sends arrive within microseconds; bundled sends wait up to 100ms for next tick flush. Combat damage dispatch and animation triggers may be sensitive — audit before migrating those families. Per `crates/services/src/base/tick_sync.rs` the tick cadence is already 100ms-based, but per-call sends short-circuit it.
- **Existing `bundle.rs` fate**: it might be load-bearing for some future plan I'm not seeing. Verify with git log + commit messages before deleting; rename to `channel_bundle.rs` and leave the old module if there's any uncertainty.
- **ACK piggyback semantics**: today each packet carries its own pending-ACK list. With Bundle the first finalized packet carries all pending ACKs and subsequent packets in the same burst carry none. Verify that the client treats this correctly (it should — ACKs are independent of the body — but worth testing).
- **Migration coexistence**: while Layer B is in progress, both old (per-call) and new (bundled) sends coexist. They share the same `next_seq` counter via `ConnectedClientState.next_seq`. Make sure ordering is preserved (the bundle's stamped seqs come from the same counter, so first-come-first-served).
- **Cell→base message ordering**: the cell sends many `CellToBaseMsg`s per tick to a single witness. With bundling, base accumulates these into one bundle per channel per tick. If two cells send to the same witness mid-tick, ordering between them is whatever order base processes them — same as today.

## Acceptance criteria

- [ ] `ChannelBundle` exists in `crates/mercury` with multi-message-per-packet finalization, fragmentation at 1348 bytes, ack-piggyback-on-first-packet semantics
- [ ] All 41 `append_entity_method` call sites migrated to bundle-append pattern (or explicitly opted-out with a tracking comment for latency-sensitive paths)
- [ ] World-entry burst finalizes into ≤3 packets (down from 5-6) — measured via debug log assertion or integration test
- [ ] AoI burst (28 NPCs at Castle_CellBlock) finalizes into ≤3 packets (down from 56+)
- [ ] All existing tests pass; no behavior regressions in wire format (covered by existing byte-pattern tests)
- [ ] ADR documenting Bundle design + tick-boundary semantics
- [ ] No regression in latency for combat/damage paths (audit, document any migrations of these paths)

## Sibling issues

- **#354** — Umbrella: TX-window relief (Bundle + defer-creates + deferred-send queue, one PR). Fixes #2 and #3 are landing on the `feat/mercury-deferred-send-queue` branch. This issue replaces the "Bundle" portion of the umbrella with proper scope.
- **#353** — Long-term: widen the client's TX window from 32 → 64 via SGW.exe binary patch + launcher integration. Compounds with this issue.
- **#355** — Network chaos test apparatus. Required to write CI-level regression guards for the bundle-burst-size assertions in Layer C.

Cadacious:
## Summary

Conservative-slice implementation of #356. Introduces `ChannelBundle` — a caller-owned accumulator for cross-entity message batching into one client frame — plus the migration of the AoI `EnteredAoI` burst, the largest single TX-window-pressure source on world entry. The 28-NPC Castle_CellBlock first-load burst collapses from **56 reliable packets to ~11**, regression-guarded.

Closes #356. Deferred call families filed as #360.

## Reconnaissance finding that shaped scope

The issue's original design (one per-channel auto-accumulator flushed on tick boundary) would re-introduce the client's HOLD-FOR-TRANSACTION drop documented at [`base/world_entry/map_loaded.rs:66-79`](crates/services/src/base/world_entry/map_loaded.rs#L66):

> Previously we combined everything into one fragmented bundle, which caused BeingAppearance to be silently dropped (HOLD FOR TRANSACTION) because the entity was still in its creation transaction during bundle processing.

`CREATE_ENTITY(X)` puts entity X in transaction for the rest of the bundle; same-entity messages later in the same bundle are silently dropped by the client. A per-channel auto-accumulator can't know to split. The implementation uses **caller-owned bundles** so the bundle-boundary decision sits where the transaction-state semantics are visible — the AoI flush deliberately splits phase-1 (CREATE_ENTITY/UPDATE_AVATAR) and phase-2 (cascade) into separate bundles.

See [`docs/architecture/mercury-bundle.md`](docs/architecture/mercury-bundle.md) for the full rule + design rationale.

### Layer A — `crates/mercury/src/channel_bundle.rs`
- New caller-owned accumulator. Body uses the same wire layout as the services-layer `append_entity_method`. Finalize delegates to the existing `build_fragmented_bundle` for fragmentation at FRAGMENT_BODY_SIZE (1300 B) and ack-on-first-packet semantics.
- 11 unit tests pin: direct/extended encoding boundary, byte-for-byte match against manual composition, multi-message packing, fragmentation, ack-only flush, empty finalize.

### Layer A.5 — `crates/services/src/base/helpers.rs`
- `send_bundle_to_witness_reliable` bridges the bundle to the session UDP socket + Channel TX window: reserves seqs atomically based on `bundle.estimated_packet_count()`, drains pending ACKs into the bundle under the same lock window, encrypts + sends + registers each fragment via `shadow_register_reliable_send`.

### Layer B (conservative) — `cell_dispatch/aoi.rs::flush_deferred_aoi`
- Extracted body-only composers from `build_create_entity_base` and `build_create_entity_cascade` (`compose_*_body`).
- Drained deferred-AoI buffer's `EnteredAoI` events now batch into two cross-entity bundles. Two-bundle split is required by the transaction-state rule.
- `LeftAoI` and `EntityMethodCall` stay on the per-message `send_to_witness_reliable` path — no burst shape, and the same-entity interleave with `EnteredAoI` needs follow-up audit.

### Layer C — tests + ADR
- `compose_create_entity_*_body` regression guards in `aoi/tests.rs`: assert byte-equivalence with the decrypted body of the standalone `build_*` packet so the migrated bundle path emits exactly the same wire bytes.
- `flush_deferred_aoi_bundles_28_npc_burst_under_packet_budget`: 28-NPC burst must emit ≤15 packets (was 56 pre-bundle); ≥2 packets to confirm phase-1 + phase-2 split survives.
- ADR at [`docs/architecture/mercury-bundle.md`](docs/architecture/mercury-bundle.md) covers the design, transaction-state rule, why caller-owned, observed packet-count win, and the migration playbook for follow-up families.

## Observed win

| Pre-bundle | Post-bundle |
|---:|---:|
| 28 × CREATE_ENTITY/UPDATE_AVATAR packet | 1 cross-entity phase-1 bundle, ~1 KB body, 1 packet |
| 28 × cascade packet | 1 cross-entity phase-2 bundle, ~12 KB body, ~10 fragments |
| **56 reliable packets** | **~11 reliable packets** |

Stacks with the deferred-send queue (#357): if a bundle finalizes into more packets than the remaining TX window, the bytes still go on the wire and bookkeeping queues for retransmit promotion as ACKs free slots.

## What did NOT change

- Wire format. Bundle body is byte-identical to a concatenation of per-message bodies the existing `append_entity_method` produces. Pinned by tests.
- Reliability. Bundled packets ride `FLAG_RELIABLE | FLAG_ON_CHANNEL` and register with the TX window the same way as pre-migration packets.
- Sequence counter. Bundle reserves N consecutive seqs via `next_seq.fetch_add(N, Relaxed)` — same counter as the per-packet path, no reordering relative to interleaved per-packet sends.
- The deferred-send queue (#357) still backstops TX-window overflow.

## Test plan

- [x] `cargo clippy -p cimmeria-mercury -p cimmeria-services --all-targets -- -D warnings`
- [x] `cargo test -p cimmeria-mercury -p cimmeria-services --lib` — 828 tests pass
- [x] `cargo test --doc -p cimmeria-commands`
- [x] New test `flush_deferred_aoi_bundles_28_npc_burst_under_packet_budget` reproduces the burst-reduction win and fails if the migration is reverted (manual revert check: pre-bundle path would emit 56 packets, violating the `<= 15` upper bound)
- [x] New test `compose_create_entity_base_body_matches_build_create_entity_base_body` + cascade variant: byte-exact equivalence between the body-only composer and the decrypted standalone-packet body — load-bearing regression guard for the migration
- [x] Existing `flush_deferred_aoi_drains_buffer_and_dispatches_to_aoi_handlers` updated docstring; 1-NPC packet count unchanged (still 4 packets through the new flow)
- [ ] Live client smoke test (Castle_CellBlock world entry) — recommended before merge to verify no transaction-state hazard slipped past the unit tests

## Sibling issues

- #356 — this PR closes it
- #354 — umbrella TX-window relief
- #357 — deferred-send queue (sibling, already shipped)
- #353 — long-term: widen client TX window via SGW.exe binary patch
- #360 — follow-up: per-handler migrations for the remaining ~13 call families

## Summary by CodeRabbit

* **New Features**
  * Implemented message bundling to improve network efficiency when sending multiple entity updates to clients simultaneously.

* **Tests**
  * Added regression tests for bundled message delivery and area-of-interest entity visibility handling.

* **Documentation**
  * Added architecture documentation for the message bundling system, including usage guidelines and transaction-state constraints.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/361?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MyPalClara:

---

### G1: `estimated_packet_count()` ACK-only edge case
**Location:** `channel_bundle.rs`

ACK-only bundle returns 1, but what if body is empty AND acks is empty? The code returns 0, which is correct. But the doc comment says "empty body + acks → 1" — what if someone adds acks AFTER calling `estimated_packet_count()`? The estimate would be stale.

The `send_bundle_to_witness_reliable` helper drains ACKs before calling `estimated_packet_count()`, so this is handled correctly in practice, but the API allows misuse.

---

### G2: `append_raw_message` no length validation
**Location:** `channel_bundle.rs`

No check that `raw_msg` is well-formed. If a caller passes truncated bytes (e.g., a CREATE_ENTITY missing its tail), the bundle body is corrupted but `finalize` succeeds. The client will fail to parse.

Consider a `debug_assert!` that the first byte is a valid msg_id, or document that the caller is responsible for well-formedness.

---

### G3: `send_bundle_to_witness_reliable` partial send
**Location:** `aoi.rs`

If fragment 3 of 10 fails to send, fragments 1-2 are already on the wire and registered. The client will receive a partial bundle and stall waiting for the missing fragments.

The code logs a warning and continues, which is probably the right behavior (let the retransmit driver recover), but the partial-send case is not tested.

---

### G4: Sequence reservation vs finalize mismatch
**Location:** `channel_bundle.rs` / `aoi.rs`

The `debug_assert!` catches this, but only in debug builds. If `estimated_packet_count()` ever disagrees with `finalize().packets.len()`, release builds will silently over-reserve or under-reserve seqs.

Consider promoting to a runtime check with a WARN log, or add a test that exhaustively verifies the contract across edge cases.

MyPalClara:

---

### G5: `EXTENDED_ENCODING_THRESHOLD = 61` duplication
**Location:** `channel_bundle.rs`

This constant is defined in `channel_bundle.rs` AND presumably in `services/src/mercury/mod.rs`. If they drift, the bundle path and the standalone-packet path will emit different bytes.

The byte-equivalence tests catch this, but consider extracting to a shared constant in `cimmeria-mercury`.

---

### G6: `body.len().div_ceil(FRAGMENT_BODY_SIZE)`
**Location:** `channel_bundle.rs`

`div_ceil` is a nightly feature stabilized in Rust 1.73. If the MSRV is lower, this will not compile.

Verify MSRV compatibility or use `(body.len() + FRAGMENT_BODY_SIZE - 1) / FRAGMENT_BODY_SIZE`.

---

### G7: `LeftAoI` and `EntityMethodCall` not migrated
**Location:** Scope

The PR body says these "stay on the per-message `send_to_witness_reliable` path." But `EntityMethodCall` can be bursty (e.g., 28 NPCs all firing `onStatUpdate` on the same tick).

Is there a follow-up issue tracking this? The PR references #360 but does not explicitly list `EntityMethodCall` burst shapes.

MyPalClara:
## Category 3: Missing Test Coverage (G8-G11)

---

### G8: No test for `send_bundle_to_witness_reliable`
**Location:** `aoi.rs`

The helper is 80 lines of async code with lock acquisition, ACK draining, seq reservation, encryption, send, and registration. There is no unit test for it — only the integration-level `flush_deferred_aoi_bundles_28_npc_burst_under_packet_budget` test exercises it indirectly.

---

### G9: No test for partial send failure
**Location:** `aoi.rs`

What if `socket.send_to` fails for one fragment? The code logs and continues, but there is no test verifying the remaining fragments still send and register.

---

### G10: No test for interleaved bundle + single-packet sends
**Location:** `channel_bundle.rs` / `aoi.rs`

What if a caller sends a bundle, then a single packet, then another bundle? The seq counter should stay contiguous. The existing tests do not exercise this interleave.

---

### G11: No test for `append_raw_message` with extended-encoding body
**Location:** `channel_bundle.rs`

The tests use `append_entity_method` for extended encoding, but `append_raw_message` is used for CREATE_ENTITY/UPDATE_AVATAR which have their own wire format. There is no test that verifies `append_raw_message` preserves arbitrary bytes correctly.

MyPalClara:
## Category 4: Documentation + Summary (G12-G13)

---

### G12: ADR does not mention the `debug_assert!` contract
**Location:** `mercury-bundle.md`

The `estimated_packet_count() == finalize().packets.len()` contract is critical for seq reservation correctness. The ADR mentions it in passing ("see the seq-reservation debug_assert") but does not explain what happens if it is violated.

---

### G13: No mention of the 1300-byte `FRAGMENT_BODY_SIZE` origin
**Location:** `mercury-bundle.md`

The ADR says "1300 bytes" but does not cite where this comes from. Is it a Mercury spec constant? A BigWorld constant? A Cimmeria choice? Cross-reference to the spec or add a one-liner explaining the origin.

---

## Summary

| Category | Gaps | Severity |
|----------|------|----------|
| Correctness issues | G1-G4 | Low-Medium |
| Potential issues | G5-G7 | Low |
| Missing test coverage | G8-G11 | Medium |
| Documentation | G12-G13 | Low |

**Total: 13 gaps**

---

## Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟡 Medium | G5 (shared constant) | 15 min |
| 🟡 Medium | G6 (MSRV check) | 5 min |
| 🟡 Medium | G8 (helper test) | 45 min |
| 🟢 Low | G1, G2, G4 (API guards) | 30 min |
| 🟢 Low | G12, G13 (ADR polish) | 15 min |
| ⚪ Follow-up | G3, G9, G10, G11 (edge case tests) | 1.5 hr |
| ⚪ Scope | G7 (EntityMethodCall burst) | Track in #360 |

**Total: ~3.5 hours of follow-up work**

Cadacious:
Addressed all review feedback in `53c9785a`:

## Correctness bugs

- **Copilot C1 — seq wrap in `build_fragmented_bundle`**: fixed with defense-in-depth masking against `SEQUENCE_MASK` on `base_seq`, per-fragment seq, `frag_begin`, and `frag_end`. Two new regression tests pin the wrap behavior at both single-packet and multi-fragment paths.
- **Copilot C2 — `flush_deferred_aoi` reorders `LeftAoI` vs `EntityMethodCall`**: replaced the two separate vecs with a single `TailMsg` enum that preserves cell-emitted encounter order. Buffered `[EntityMethodCall(X), LeftAoI(X)]` now fires the method first.
- **CodeRabbit CR3 + Clara G3 — fragment send failure should abort, not continue**: `send_bundle_to_witness_reliable` now `return`s on first failure with an `error!` log naming the gap range. Pinned by `send_bundle_to_witness_reliable_aborts_on_first_fragment_send_failure`.

## API guards (CR1, Clara G2, G4)

- `append_entity_method` uses `u8::try_from` / `u16::try_from` with explicit panic messages for `method_index > 316` or `args > u16::MAX`. Two new `#[should_panic]` tests.
- `append_raw_message` has a debug-only assertion against empty input.
- The `estimated_packet_count == finalize().packets.len()` contract is now pinned at the fragment boundary with ACKs present (CR2).

## Anti-drift (Clara G5)

- `EXTENDED_ENCODING_THRESHOLD` and `EXTENDED_ENCODING_MARKER` promoted to `pub const` in `cimmeria_mercury::channel_bundle`. The services-layer `append_entity_method` now imports them so the two encoders cannot diverge.

## Comment hygiene (Copilot C4, C5, C7, C8 + CR4, CR5)

- Removed `#356` references from all source comments per repo convention.
- Fixed the stale test-name link (`_into_seven_packets` → `_under_packet_budget`).
- Fixed the inaccurate "method index 122 observed working with direct" comment.

## ADR polish (Copilot C10, C11 + Clara G12, G13)

- Rewrote the `estimated_packet_count` contract section to explain the over-/under-reserve consequences.
- Fixed typo "per-message body bodies" → "per-message bodies".
- Documented the `FRAGMENT_BODY_SIZE = 1300` origin (MAX_BODY_LENGTH 1411 − footer + AES-CBC overhead).
- Documented the seq-masking guarantee.
- Reconciled packet-count claims: ~11 typical, ≤15 regression guard.

## Test coverage additions (Clara G8, G9, G10, G11)

- `send_bundle_to_witness_reliable_is_noop_for_unknown_witness` (G8)
- `flush_deferred_aoi_keeps_reliable_seq_contiguous_across_bundle_and_tail` (G10)
- `append_raw_message_preserves_arbitrary_bytes_in_order` (G11)
- `send_bundle_to_witness_reliable_aborts_on_first_fragment_send_failure` (G3 + G9 combined)

## Tooling

- `tools/lint-md.sh` now has the executable bit set in git (CR outside-diff).

## Out of scope for this PR

- Clara G7 — `EntityMethodCall` burst migration is in scope of the follow-up; commented on #360 to call it out explicitly.
- The TX-window seq-ordering vs concurrent fetch_add race is pre-existing; bundles slightly widen the window between the first and last fragment but don't introduce the bug. Worth its own PR with a redesign of channel locking.

## Verification

- `cargo fmt --check` ✅
- `cargo clippy -p cimmeria-mercury -p cimmeria-services --all-targets -- -D warnings` ✅
- `cargo check --workspace` (excl GUI apps) ✅
- `cargo test -p cimmeria-mercury -p cimmeria-services --lib` — **835 pass** (+3 from prior commit; +7 new tests, net after consolidation) ✅
- `cargo test --doc -p cimmeria-commands` ✅

Adversarial self-review pass confirmed:
- Seq-wrap fix propagates to `fragment_map_loaded` via delegation
- No production caller of `append_raw_message` passes empty bytes
- Ordering fix preserves cell-emitted order even in the pre-existing reorder-prone case (`[LeftAoI(X), EntityMethodCall(X)]` was always going to fire LeftAoI first)
- Workspace check clean

PR #362, opened 2026-05-24
test(holster): pin RefreshAppearance on fire-while-holstered queue (closes #339, #333)

### LINKED ISSUE #333: cleanup(state): BSF_Holster bit 8 does nothing client-side — remove or document the no-op

### LINKED ISSUE #339: feat(cell): wire BeingAppearance rebroadcast on runtime holster toggle (fire / reload / requestHolsterWeapon)
## Plain English

[#338](https://github.com/SandboxServers/Cimmeria/pull/338) landed the **spawn-holstered** half of the holster fix: players join the world weapon-down because the world-entry path emits a `BeingAppearance.ComponentList` with the active bandolier weapon filtered out. The server-side holster state lives on `CellEntity::weapon_holstered` (defaults to `true`).

What's not yet wired: **the runtime toggle paths.** When the player fires, reloads, or hits the holster keybind, the server-side state changes but no `BeingAppearance` rebroadcast goes out, so other players in AoI keep seeing the player in the previous pose. The holster button is no longer broken-as-no-op (server state updates correctly), but it's still broken-as-not-visible-to-witnesses.

This is the followup to make the runtime toggle visible.

## Code sites that need the rebroadcast

1. **`crates/services/src/cell/cell_methods/combatant.rs`** — `requestHolsterWeapon` (method 7). Already calls `set_weapon_holstered(holstered)`. Needs to follow with a `BeingAppearance` recomposite + broadcast to all AoI witnesses (including self for first-person). The `tx` sender is already plumbed in and currently unused (annotated with a TODO pointing at this issue).

2. **`crates/services/src/cell/abilities/use_ability.rs`** — the fire commit path. When a player commits an ability launch, they're drawing the weapon (Atrea convention: a fire-with-holstered-weapon doesn't make sense). The path should call `set_weapon_holstered(false)` on the attacker and trigger the rebroadcast. The previous `state_field &= !BSF_HOLSTER` write at this site was a no-op on the wire (#333); the new equivalent is the rebroadcast.

3. **`crates/services/src/cell/cell_methods/player/world.rs`** — the reload path. Same shape as fire: reloading draws the weapon, so `set_weapon_holstered(false)` + rebroadcast.

## Wire format

`Event_NetIn_BeingAppearance` carries `{bodyset: wstring, components: array<wstring>}`. The runtime rebroadcast emits the same shape the world-entry path already builds via:

- [`PlayerLoadData::appearance_components(holstered: bool)`](crates/services/src/mercury/types.rs) — at world entry, called with `holstered = true`.
- [`CellEntity::appearance_components()`](crates/entity/src/cell_entity/mod.rs) — for runtime, reads the live `self.weapon_holstered`.

Both delegate to the shared free function [`filter_holstered_weapon`](crates/entity/src/cell_entity/mod.rs) which carries the `weapon_visual ∈ components` invariant via `debug_assert!` + a warn-level log.

Verified Ghidra anchors for the wire contract:
- `CompositedAppearanceProxy::ApplyToPawn @ ghidra://SGW.exe@0x00ec0840` — writes `entity+0x3D2` (animation-key weapon-category byte).
- The compositor's weapon-category sources from the `ComponentList`, not from any state-flag bit. (See [`docs/architecture/state-field-bits.md`](docs/architecture/state-field-bits.md) for the full byte-pattern scan that confirmed bits 8-31 of `bStateField` are wire-dead.)

## Proposed shape

Two pieces:

1. **A new helper** — `broadcast_being_appearance(entity_id, tx, space_mgr)` (working name; refine during impl). Reads the entity's current `weapon_holstered` and live `components`/`weapon_visual`, builds the wire args, fans out via the witness-fanout helper landed in #278. Lives in either `crates/services/src/cell/abilities/messaging.rs` (alongside the existing `send_entity_method_to_self_and_witnesses`) or a sibling appearance-broadcast module — choose during impl.
2. **Three callsite migrations** — combatant.rs `requestHolsterWeapon`, use_ability.rs commit path, player/world.rs reload path. Each pattern is `set_weapon_holstered(...)` → `broadcast_being_appearance(...)`.

## Test plan

- [ ] Unit: `set_weapon_holstered(true)` followed by the broadcast helper emits exactly one `BeingAppearance.EntityMethodCall` (self) + N `WitnessEntityMethod` calls (one per AoI witness).
- [ ] Wire-format regression: byte-exact `ComponentList` matches the spawn-holstered shape from [#338](https://github.com/SandboxServers/Cimmeria/pull/338)'s `build_map_loaded_filters_weapon_visual_from_being_appearance` test.
- [ ] Live UAT: player A holsters, player B in same AoI sees A's weapon disappear. Player A fires, player B sees A's weapon reappear. Same for reload.
- [ ] Negative test: `set_weapon_holstered(holstered)` with `weapon_visual = None` (empty bandolier slot) must NOT broadcast — there's nothing visually different to send (already pinned at the entity level via `set_weapon_holstered`'s `false` return).

## Out of scope

- **Reload-completion rebroadcast** for when the reload tick actually completes (vs. when reload starts). That's a separate animation-system concern (the reload animation plays through several states; the visual weapon is in different positions across them). For now, "reload starts → weapon drawn" is enough.
- **Weapon-pickup / equip-while-holstered** flow. If the player picks up a new weapon while their bandolier was previously empty, the new active weapon needs to seed `weapon_visual`. That's the equip path's responsibility ([#240 Gap 2](https://github.com/SandboxServers/Cimmeria/issues/240) territory), not this issue.

## Related

- Parent: [#279](https://github.com/SandboxServers/Cimmeria/issues/279) — BeingAppearance recomposite broadcast on equip-change.
- Sibling: [#278](https://github.com/SandboxServers/Cimmeria/issues/278) — witness-fanout helper (provides the fanout primitive this issue's helper builds on).
- Predecessor: [#338](https://github.com/SandboxServers/Cimmeria/pull/338) — landed spawn-holstered. Closed [#249](https://github.com/SandboxServers/Cimmeria/issues/249) and [#333](https://github.com/SandboxServers/Cimmeria/issues/333) with the runtime piece deferred here.
- Adjacent: [#232](https://github.com/SandboxServers/Cimmeria/issues/232) — death/respawn state-field broadcast (same fanout pattern, different state).

Cadacious:
## Summary

PR #338 already implemented everything #339 and #333 asked for — but the GitHub auto-close didn't fire (the PR body used a `## Closes` markdown header rather than a `Closes #N` keyword), so both issues stayed open. This PR closes the coverage gap I found while auditing and properly closes the tickets.

## What was already done in PR #338

**#333 (BSF_HOLSTER cleanup):**

- `BSF_HOLSTER` constant removed from `cell/combat/state.rs` and the re-export in `combat/mod.rs`.
- Dead clear-on-fire write in `use_ability.rs` and clear-on-reload write in `cell_methods/player/world.rs` deleted.
- `requestHolsterWeapon` switched from the dead bit to `CellEntity::set_weapon_holstered`.
- New canonical doc [`docs/architecture/state-field-bits.md`](docs/architecture/state-field-bits.md) with the verified bit→side-effect table, pinned to Ghidra anchors.
- [`docs/architecture/state-flag-conventions.md`](docs/architecture/state-flag-conventions.md) row dropped + cross-link added.

**#339 (runtime toggle rebroadcast):** all three callsites already wired:

- `crates/services/src/cell/cell_methods/combatant.rs:91` — `requestHolsterWeapon` → `request_appearance_refresh`.
- `crates/services/src/cell/abilities/use_ability/mod.rs:254` — fire-while-holstered queue → `request_appearance_refresh`.
- `crates/services/src/cell/cell_methods/player/world/mod.rs:354` — reload-while-holstered queue → `request_appearance_refresh`.

The `request_appearance_refresh` helper sends `CellToBaseMsg::RefreshAppearance`, which the base-side `refresh_appearance` handler turns into a fan-out of `BeingAppearance` to the owner + all AoI witnesses via `refresh_player_appearance`.

### Test coverage gap

The combatant.rs path (`request_holster_weapon_dispatches_refresh_appearance`) and the reload path (`reload_while_holstered_phase_a_defers_reload`) each have a strict test that drains the `tx` receiver and asserts a `RefreshAppearance` was sent. The use_ability fire path's existing test (`attack_while_holstered_queues_and_draws_without_committing`) only checked state mutation — it discarded `_rx`, so dropping the `request_appearance_refresh` call would silently pass.

Adversarial verification: commenting out `request_appearance_refresh` at `use_ability/mod.rs:254` makes the new assertion fail (`left: 0, right: 1`) — exactly the #339 bug shape (server state correct, AoI witnesses stuck on holstered pose).

### Ghidra re-verification of the wire contract

I re-decompiled both anchor functions against the live `SGW.exe` binary to confirm the design's bedrock claims still hold:

- **`GameBeing_OnStateFieldUpdate @ ghidra://SGW.exe@0x00e01c90`** — XORs current ^ new at `+0x158`, then dispatches on masks `0x01, 0x02, 0x08, 0xC4, 0x10, 0x20`. Bit 8 (mask `0x100`) is **never tested**. Confirms #333's premise.
- **`CompositedAppearanceProxy::ApplyToPawn @ ghidra://SGW.exe@0x00ec0840`** — `iVar2 = *(int*)(this+0x34)` (proxy weapon-category, defaults to 4=WEAP_Melee on zero), then `*(char*)(param_5+0x3d2) = (char)iVar2`. The animation-key posture byte is fed from the appearance compositor, never from `bStateField`. Confirms #339's wire contract.

### Sidewalks

Two stale `BSF_HOLSTER` references in wire-flow diagrams refreshed:

- `docs/gameplay/weapon-ammo-reload.md` — two `onStateFieldUpdate (… | clear BSF_HOLSTER)` annotations updated to reflect the actual mechanism (BSF_IN_COMBAT bit only + separate `BeingAppearance` broadcast).
- `docs/reverse-engineering/findings/animation-system.md` — the "Implication for #249" section consolidated into a resolution note covering #249 + #333 + #339.

## Test plan

- [x] `cargo clippy --workspace --exclude cimmeria-app --exclude cimmeria-content-editor --exclude cimmeria-scene-editor --exclude sgw-launcher --all-targets -- -D warnings` clean
- [x] `cargo test -p cimmeria-services --lib` — **835 / 835 passing**
- [x] `cargo test -p cimmeria-entity --lib` — **160 / 160 passing**
- [x] Adversarial: commented out the `request_appearance_refresh` call at `use_ability/mod.rs:254` → new assertion fails with `left: 0, right: 1`. Restored.
- [x] Ghidra wire-contract verification (above).
- [x] Markdown lint: my two doc edits introduce no new lint violations (pre-existing violations in adjacent docs remain — that's a separate cleanup).

## Closes

- Closes #339
- Closes #333

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for attack behavior while holstered to ensure proper character appearance refresh and unholster state synchronization.

* **Documentation**
  * Updated weapon reload sequence documentation to clarify visual behavior and timing during reload while holstered.
  * Refined animation system documentation to explain how holster state changes drive weapon appearance updates during combat scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/362?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/services/src/cell/abilities/use_ability/tests.rs
+++ b/crates/services/src/cell/abilities/use_ability/tests.rs
@@ -315,7 +315,7 @@ async fn attack_while_holstered_queues_and_draws_without_committing() {
     }
     // required_ammo=1 → triggers the weapon-attack queue.
     mgr.ability_defs.insert(7, make_ability(7, 1, 30));
-    let (tx, _rx) = mpsc::channel(64);
+    let (tx, mut rx) = mpsc::channel(64);
 
     let committed = handle_use_ability(1, 7, 0, &tx, &mut mgr).await;
     assert!(
@@ -346,6 +346,37 @@ async fn attack_while_holstered_queues_and_draws_without_committing() {
     assert_eq!(
         e.bandolier_items[&0].current_ammo, 30,
         "Phase A must NOT consume ammo — ammo check happens in Phase B",
+    );
+
+    // Pin the runtime rebroadcast (#339): Phase A must dispatch exactly
+    // one `RefreshAppearance(holstered=false)` so the base-side handler
+    // unwraps the weapon mesh in the `ComponentList` and rebroadcasts
+    // `BeingAppearance` to self + AoI witnesses. Without this assertion
+    // the test would pass even if a refactor dropped the
+    // `request_appearance_refresh` call — the bug shape #339 was filed
+    // against (server state mutates correctly but other players still
+    // see the player in the holstered pose).
+    let mut refresh_count = 0;
+    while let Ok(msg) = rx.try_recv() {
+        if let CellToBaseMsg::RefreshAppearance {
+            entity_id,
+            holstered,
+            ..
+        } = msg
+        {
+            assert_eq!(entity_id, 1, "RefreshAppearance must target the attacker");
+            assert!(
+                !holstered,
+                "Phase A must broadcast holstered=false (mesh attaches)",
+            );
+            refresh_count += 1;
+        }
+    }
+    assert_eq!(
+        refresh_count, 1,
+        "Phase A must dispatch exactly one RefreshAppearance — \
+         dropping it leaves AoI witnesses stuck on the holstered pose \
+         while the attacker animates an invisible draw",
     );
 }
 
--- a/crates/services/src/cell/abilities/use_ability/tests.rs
+++ b/crates/services/src/cell/abilities/use_ability/tests.rs
@@ -356,27 +356,28 @@ async fn attack_while_holstered_queues_and_draws_without_committing() {
     // `request_appearance_refresh` call — the bug shape this guards
     // against (server state mutates correctly but other players still
     // see the player in the holstered pose).
-    let mut refresh_count = 0;
-    while let Ok(msg) = rx.try_recv() {
-        if let CellToBaseMsg::RefreshAppearance {
-            entity_id,
-            holstered,
-            ..
-        } = msg
-        {
-            assert_eq!(entity_id, 1, "RefreshAppearance must target the attacker");
-            assert!(
-                !holstered,
-                "Phase A must broadcast holstered=false (mesh attaches)",
-            );
-            refresh_count += 1;
-        }
-    }
+    let refreshes: Vec<_> = drain(&mut rx)
+        .into_iter()
+        .filter_map(|m| match m {
+            CellToBaseMsg::RefreshAppearance {
+                entity_id,
+                holstered,
+                ..
+            } => Some((entity_id, holstered)),
+            _ => None,
+        })
+        .collect();
     assert_eq!(
-        refresh_count, 1,
+        refreshes.len(),
+        1,
         "Phase A must dispatch exactly one RefreshAppearance — \
          dropping it leaves AoI witnesses stuck on the holstered pose \
          while the attacker animates an invisible draw",
+    );
+    assert_eq!(
+        refreshes[0],
+        (1, false),
+        "RefreshAppearance must target the attacker with holstered=false",
     );
 }
 
--- a/crates/services/src/cell/abilities/use_ability/tests.rs
+++ b/crates/services/src/cell/abilities/use_ability/tests.rs
@@ -348,12 +348,12 @@ async fn attack_while_holstered_queues_and_draws_without_committing() {
         "Phase A must NOT consume ammo — ammo check happens in Phase B",
     );
 
-    // Pin the runtime rebroadcast (#339): Phase A must dispatch exactly
-    // one `RefreshAppearance(holstered=false)` so the base-side handler
+    // Pin the runtime rebroadcast: Phase A must dispatch exactly one
+    // `RefreshAppearance(holstered=false)` so the base-side handler
     // unwraps the weapon mesh in the `ComponentList` and rebroadcasts
     // `BeingAppearance` to self + AoI witnesses. Without this assertion
     // the test would pass even if a refactor dropped the
-    // `request_appearance_refresh` call — the bug shape #339 was filed
+    // `request_appearance_refresh` call — the bug shape this guards
     // against (server state mutates correctly but other players still
     // see the player in the holstered pose).
     let mut refresh_count = 0;
--- a/docs/gameplay/weapon-ammo-reload.md
+++ b/docs/gameplay/weapon-ammo-reload.md
@@ -98,7 +98,7 @@ Client                       Cell                                       Base / D
   │                           │   reload_slot_id = Some(active_bandolier_slot)  ← pin
   │                           │   abilities.start_ability_cooldown(596, warmup+cooldown)
   │ ◀── onTimerUpdate (m12) ──│  cooldown bar starts
-  │ ◀── onEntityProperty (m25)│  ammo-type sync (cur_ammo_type)
+  │ ◀── onEntityProperty (m7)│  ammo-type sync (cur_ammo_type)
   │                           │
   │  Note: handle_reload does NOT touch bStateField — BSF_IN_COMBAT is
   │  derived from `threatened_mobs` and only flips via combat::generate_threat.
@@ -247,7 +247,7 @@ Client                       Cell                              Base / DB
   │ requestReload(0) ─────────▶│  reload_complete_at = now + 2.0s
   │                            │  start_ability_cooldown(596, 3.0s)
   │ ◀── onTimerUpdate(m12) ────│  cooldown bar
-  │ ◀── onEntityProperty(m25) ─│  ammo-type sync
+  │ ◀── onEntityProperty(m7) ──│  ammo-type sync
   │                            │  (weapon already drawn from the fire above,
   │                            │   so no BeingAppearance — and handle_reload
   │                            │   intentionally does NOT touch bStateField)
--- a/docs/gameplay/weapon-ammo-reload.md
+++ b/docs/gameplay/weapon-ammo-reload.md
@@ -97,13 +97,14 @@ Client                       Cell                                       Base / D
   │                           │   reload_complete_at = now + warmup
   │                           │   reload_slot_id = Some(active_bandolier_slot)  ← pin
   │                           │   abilities.start_ability_cooldown(596, warmup+cooldown)
-  │ ◀── onTimerUpdate (m12) ──│
-  │ ◀── onStateFieldUpdate ───│  (BSF_IN_COMBAT — bit 3 only)
-  │ ◀── BeingAppearance ──────│  ComponentList includes weapon_visual
-  │                           │  (if reload-while-holstered: Phase A first
-  │                           │   draws weapon via RefreshAppearance →
-  │                           │   BeingAppearance, then defers the actual
-  │                           │   reload by UNHOLSTER_DRAW_DURATION)
+  │ ◀── onTimerUpdate (m12) ──│  cooldown bar starts
+  │ ◀── onEntityProperty (m25)│  ammo-type sync (cur_ammo_type)
+  │                           │
+  │  Note: handle_reload does NOT touch bStateField — BSF_IN_COMBAT is
+  │  derived from `threatened_mobs` and only flips via combat::generate_threat.
+  │  A BeingAppearance rebroadcast only fires on the reload-while-holstered
+  │  Phase A path (request_appearance_refresh draws the weapon, the actual
+  │  reload defers by UNHOLSTER_DRAW_DURATION).
   │                           │
   │  (warmup elapses, e.g. 2 s later)
   │                           │
@@ -246,10 +247,10 @@ Client                       Cell                              Base / DB
   │ requestReload(0) ─────────▶│  reload_complete_at = now + 2.0s
   │                            │  start_ability_cooldown(596, 3.0s)
   │ ◀── onTimerUpdate(m12) ────│  cooldown bar
-  │ ◀── onStateFieldUpdate ────│  +BSF_IN_COMBAT (bit 3 only — see
-  │                            │   docs/architecture/state-field-bits.md;
-  │                            │   bit 8 / BSF_HOLSTER was retired in #333)
-  │ ◀── BeingAppearance ───────│  weapon mesh stays attached during reload
+  │ ◀── onEntityProperty(m25) ─│  ammo-type sync
+  │                            │  (weapon already drawn from the fire above,
+  │                            │   so no BeingAppearance — and handle_reload
+  │                            │   intentionally does NOT touch bStateField)
   │
   │   (≈2 s later — next reload_completion_tick after deadline)
   │                            │  refill_active_slot()  → cur=5
--- a/docs/gameplay/weapon-ammo-reload.md
+++ b/docs/gameplay/weapon-ammo-reload.md
@@ -98,7 +98,12 @@ Client                       Cell                                       Base / D
   │                           │   reload_slot_id = Some(active_bandolier_slot)  ← pin
   │                           │   abilities.start_ability_cooldown(596, warmup+cooldown)
   │ ◀── onTimerUpdate (m12) ──│
-  │ ◀── onStateFieldUpdate ───│  (BSF_IN_COMBAT | clear BSF_HOLSTER)
+  │ ◀── onStateFieldUpdate ───│  (BSF_IN_COMBAT — bit 3 only)
+  │ ◀── BeingAppearance ──────│  ComponentList includes weapon_visual
+  │                           │  (if reload-while-holstered: Phase A first
+  │                           │   draws weapon via RefreshAppearance →
+  │                           │   BeingAppearance, then defers the actual
+  │                           │   reload by UNHOLSTER_DRAW_DURATION)
   │                           │
   │  (warmup elapses, e.g. 2 s later)
   │                           │
@@ -241,7 +246,10 @@ Client                       Cell                              Base / DB
   │ requestReload(0) ─────────▶│  reload_complete_at = now + 2.0s
   │                            │  start_ability_cooldown(596, 3.0s)
   │ ◀── onTimerUpdate(m12) ────│  cooldown bar
-  │ ◀── onStateFieldUpdate ────│  +BSF_IN_COMBAT, -BSF_HOLSTER
+  │ ◀── onStateFieldUpdate ────│  +BSF_IN_COMBAT (bit 3 only — see
+  │                            │   docs/architecture/state-field-bits.md;
+  │                            │   bit 8 / BSF_HOLSTER was retired in #333)
+  │ ◀── BeingAppearance ───────│  weapon mesh stays attached during reload
   │
   │   (≈2 s later — next reload_completion_tick after deadline)
   │                            │  refill_active_slot()  → cur=5
--- a/docs/reverse-engineering/findings/animation-system.md
+++ b/docs/reverse-engineering/findings/animation-system.md
@@ -300,8 +300,8 @@ Server sends Event_NetIn_BeingAppearance {BodySet, ComponentList}
 Resolution:
 
 - **#249** (spawn-holstered) — fixed in PR #338. World-entry path emits `BeingAppearance` with `weapon_visual` filtered from `ComponentList`.
-- **#333** (BSF_HOLSTER retirement) — fixed in PR #338. Constant removed; dead clear-on-fire / clear-on-reload writes removed.
-- **#339** (runtime toggle rebroadcast) — fixed in PR #338. `combatant.rs requestHolsterWeapon`, `use_ability.rs` fire-while-holstered queue, and `player/world.rs` reload-while-holstered queue all call `request_appearance_refresh` → base-side `BeingAppearance` rebroadcast to self + AoI witnesses.
+- **#333** (BSF_HOLSTER retirement) — fixed in PR #338. Constant removed; dead clear-on-fire / clear-on-reload writes removed. Test-and-docs follow-up in PR #362 refreshed the wire-flow annotations in [`docs/gameplay/weapon-ammo-reload.md`](../../gameplay/weapon-ammo-reload.md) and added this resolution note.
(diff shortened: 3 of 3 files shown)


MyPalClara:
## Category 1: Issue Coverage Verification

**Issues claimed:** #339 (runtime holster toggle rebroadcast), #333 (BSF_HOLSTER cleanup)

### #339 Coverage — ✅ COMPLETE

The issue required three callsite migrations:
1. `combatant.rs requestHolsterWeapon` → `request_appearance_refresh` — **Already done in PR #338**
2. `use_ability.rs` fire commit path → `request_appearance_refresh` — **Already done in PR #338**
3. `player/world.rs` reload path → `request_appearance_refresh` — **Already done in PR #338**

**This PR's contribution:** Adds the missing test coverage for the `use_ability.rs` path. The existing test discarded `_rx`, so dropping the `request_appearance_refresh` call would silently pass. The new assertion pins `RefreshAppearance { holstered: false }` dispatch.

**Adversarial verification documented:** Commenting out `request_appearance_refresh` at `use_ability/mod.rs:254` makes the new assertion fail (`left: 0, right: 1`).

### #333 Coverage — ✅ COMPLETE

The issue required:
- Delete `BSF_HOLSTER` constant — **Done in PR #338**
- Remove dead clear-on-fire/clear-on-reload writes — **Done in PR #338**
- Add `docs/architecture/state-field-bits.md` — **Done in PR #338**

**This PR's contribution:** Updates stale `BSF_HOLSTER` references in two wire-flow diagrams:
- `docs/gameplay/weapon-ammo-reload.md` — annotations updated
- `docs/reverse-engineering/findings/animation-system.md` — resolution note added

---

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | Test assertion order | **The test drains `rx` AFTER all state assertions.** If a future refactor moves the `request_appearance_refresh` call earlier in the handler (before state mutation), the test would still pass even though the wire order changed. |
| **G2** | `try_recv()` loop | **The loop uses `try_recv()` which is non-blocking.** If the handler is async and the message hasn't arrived yet, the loop exits with `refresh_count = 0`. The test works because `handle_use_ability` is `await`ed before the drain, but this is implicit. |

---

| **G3** | No negative test for `weapon_visual = None` | **Issue #339 explicitly requires:** "Negative test: `set_weapon_holstered(holstered)` with `weapon_visual = None` (empty bandolier slot) must NOT broadcast." This PR doesn't add that test. |
| **G4** | No test for `holstered = true` broadcast | **The test only pins `holstered = false` (draw).** The holster direction (`holstered = true`) via `requestHolsterWeapon` is tested elsewhere, but this file doesn't have symmetry. |

---

## Category 4: Documentation

| Gap | Location | Concern |
|-----|----------|---------|
| **G5** | `animation-system.md` resolution note | **The resolution note references PR #338 three times but doesn't mention this PR.** If this PR is the one that formally closes #339/#333, the note should reference it. |
| **G6** | No update to issue #339's test plan | **Issue #339 has a test plan with checkboxes.** This PR addresses one of them but doesn't tick the checkbox in the issue body. |

---

## Summary

| Issue coverage | 2/2 | ✅ Both issues addressed |
| Correctness issues | 2 | **Low** — implicit ordering |
| Missing test coverage | 2 | **Low-Medium** — G3 is in issue's test plan |
| Documentation | 2 | **Low** — cross-reference polish |

**Total: 6 gaps**

### Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟡 Consider | G3 — negative test for `weapon_visual = None` | 20 min |
| ⚪ Nice-to-have | G1, G2, G4, G5, G6 | ~30 min total |

Cadacious:
Thanks for the reviews. Pushed [6151b21e](https://github.com/SandboxServers/Cimmeria/commit/6151b21e) addressing all the still-valid findings:

**Fixed**

- **CodeRabbit + Copilot**: Stripped `#339` references from the test's inline comments per the repo convention (provenance belongs in PR/commit, not source).
- **Copilot ×2 on weapon-ammo-reload.md**: You were right — my first pass on the diagrams was still wrong. Re-read `handle_reload` and the comment at `player/world/mod.rs:451` explicitly says \"BSF_InCombat is intentionally not touched here.\" Phase B only emits `onTimerUpdate (m12)` + `onEntityProperty (m25)`. The `BeingAppearance` rebroadcast only fires on the reload-while-holstered Phase A path. Both diagrams now reflect that, with the Phase A exception called out in prose.
- **Clara G5**: `animation-system.md` resolution note now credits this PR for the test-gap closure alongside PR #338's primary implementation.

**Skipped with reasons**

- **Clara G1 (drain order)**: The drain happens after `await`ing `handle_use_ability`. Because the function fully completes its `tx.send().await` calls before returning, all messages have landed in the channel by the time the test reads `try_recv()` — drain order vs. state assertions is just readability, not a correctness gap. This pattern is identical to `reload_while_holstered_phase_a_defers_reload` and `request_holster_weapon_dispatches_refresh_appearance`.
- **Clara G2 (`try_recv` non-blocking)**: Same root reason — the `await` upstream guarantees the sends have completed; `try_recv` is the standard drain idiom in this test suite.
- **Clara G3 (negative test for `weapon_visual = None`)**: This requirement in #339's test plan is outdated. The `set_weapon_holstered` change-detection was deliberately rewritten to NOT gate on `weapon_visual` — see the doc comment at [`crates/entity/src/cell_entity/mod.rs:663-670`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/entity/src/cell_entity/mod.rs#L663-L670): \"The `weapon_visual` check that used to live here was wrong: in production the cell entity's `weapon_visual` is always `None` (only the base side populates it from `PlayerLoadData`), so the gate silently suppressed every holster broadcast.\" Adding the test from the issue body would now lock in the bug.
- **Clara G4 (no `holstered = true` test in this file)**: The holster-direction broadcast is already pinned by `request_holster_weapon_dispatches_refresh_appearance` in `cell_methods/combatant.rs` — that's the only path the client uses to ask for a manual holster. The use_ability fire path is only ever `holstered = false` (firing draws the weapon, by definition).
- **Clara G6 (tick #339's checkboxes)**: Will tick them when the issue closes.

**Ghidra evidence reminder**: re-verified before pushing — `GameBeing_OnStateFieldUpdate @ 0x00e01c90` only tests bits 0-7 (masks `0x01, 0x02, 0x08, 0xC4, 0x10, 0x20`); `CompositedAppearanceProxy::ApplyToPawn @ 0x00ec0840` writes `pawn+0x3D2` from `proxy+0x34`, never from `bStateField`. The whole holster-via-appearance design rests on these two anchors.

Cadacious:
Copilot pass 2 — both findings fixed in [4d3f8892](https://github.com/SandboxServers/Cimmeria/commit/4d3f8892):

- **`onEntityProperty (m25)` → `(m7)`** in both diagrams. The constant is `spawnable_entity::ON_ENTITY_PROPERTY = 7` ([client_methods/spawnable_entity.rs:18](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/client_methods/spawnable_entity.rs#L18)). Good catch — that was a straight typo on my part.
- **Use the file-local `drain` helper** in the test. Now reads:
  ```rust
  let refreshes: Vec<_> = drain(&mut rx)
      .into_iter()
      .filter_map(|m| match m { CellToBaseMsg::RefreshAppearance { ... } => Some(...), _ => None })
      .collect();
  ```
  Adversarial re-verification: commenting out the production `request_appearance_refresh` call still makes the test fail with `left: 0, right: 1`, so the refactor didn't weaken the guard.

Issue #330, opened 2026-05-18
feat(cell): aggression-override server→client wire-out (onAggressionOverrideUpdate / Cleared)


Cadacious:
## Plain English

NPCs in SGW have a per-target **aggression level** that controls UI presentation client-side (nameplate color, targeting reticle hostility). Five levels: `HOSTILE`/`SUSPICIOUS`/`NEUTRAL`/`FRIENDLY` plus `DEFAULT` (the "no override active" sentinel). The wire surface is two events: `Event_NetIn_onAggressionOverrideUpdate(int8 aAggressionLevel)` and `Event_NetIn_onAggressionOverrideCleared()` (empty payload), plus a `CELL_PUBLIC` replicated array property `aggressionOverrides`.

**Cimmeria does not implement any of this server side today.** Mob NPCs render with default colors no matter what relationship they should have to the player. This blocks quest flows that need a friendly NPC to look friendly (e.g., a Jaffa standing guard alongside the player should be visually distinct from the hostile Jaffa five meters away).

## Spec evidence

All addresses are in `SGW.exe` at image base `0x00400000`.

### Aggression level enum — verified values

| Atrea name | Value | Lua getter | Source data |
|---|---|---|---|
| `AGGRESSION_HOSTILE` | 1 | `ghidra://SGW.exe@0x00aa2360` (`FLD1` inline) | n/a (inline 1.0) |
| `AGGRESSION_SUSPICIOUS` | 2 | `ghidra://SGW.exe@0x00a9eab0` | double @ `0x01866fe0` = 2.0 |
| `AGGRESSION_NEUTRAL` | 3 | `ghidra://SGW.exe@0x00a9ea90` | double @ `0x0183db60` = 3.0 |
| `AGGRESSION_FRIENDLY` | 4 | `ghidra://SGW.exe@0x00a9e610` | double @ `0x0193f230` = 4.0 |
| `AGGRESSION_DEFAULT` | 5 | (sentinel, no Lua getter) | constructor + Cleared handler write `0x5` |

The `UIAggressionLevel` Lua enum binds exactly four entries (HOSTILE/SUSPICIOUS/NEUTRAL/FRIENDLY at `0x01955b50..0x01955b90` UTF-16). `DEFAULT = 5` is the *sentinel*: the value the client writes to `GameMob+0x16c` when (a) the mob is constructed, or (b) the server sends `onAggressionOverrideCleared`. It is never a valid override *target*.

### Wire format — verified

[...]

Issue #329, opened 2026-05-18
fix(cell): NPC AI should use per-ability range + minimum-range backup + re-tick on launch failure


Cadacious:
## Plain English

The NPC AI tick uses a single global constant `NPC_ATTACK_RANGE = 30.0` for every NPC's "am I in range to fire" check. That's wrong on two axes:

1. **Different abilities have different ranges.** `AbilityDef` already carries `min_range` and `max_range` fields loaded from the DB, but the AI tick ignores them and uses the flat 30.0 constant. A grenade ability with `max_range = 15` is correctly out-of-range at distance 20, but the AI sees "in range" and tries to fire; the `useAbility` path checks the ability's actual range and rejects, so the NPC just stands there.
2. **No minimum-range backup behavior.** When a target is *inside* an ability's `min_range` (e.g., a sniper rifle that can't fire under 5m), the NPC can't fire and has no logic to step back. It stops in place, perpetually trying to fire, until the player moves or kills it.

Additionally, on `useAbility` launch failure (target dead between range-check and launch, animation lock, etc.), the AI does not re-schedule a tick — it has to wait for the next 2-second AI tick to retry. Adds visible "dead frame" lag.

## Spec evidence

Per the client binary (SGW.exe at image base `0x00400000`), every `Ability` carries its own min/max range fields, stored client-side in `AbilityType` struct at offsets `+0x60` (`MinRange`, float) and `+0x64` (`MaxRange`, float). These are deserialized from `CookedDataAbilities.pak`:

- `AbilityType` PAK deserializer: `ghidra://SGW.exe@0x015D4400` — reads `"MinRange"` and `"MaxRange"` field strings directly out of the PAK stream
- Field-name assert strings: `"maxRange >= 0"` at `ghidra://SGW.exe@0x019EA8B4`, `"minRange >= 0"` at `ghidra://SGW.exe@0x019EA8E0` — the client validates non-negative at load
- `CookedData:ItemRangeSetType` (type ID `0x18` in the cooked-data dispatch table at `ghidra://SGW.exe@0x015E5FF0`) is the 24-byte item-range record with `MinRange` at offset `+4` and `MaxRange` at offset `+8`. Deserializer: `ghidra://SGW.exe@0x015DEEB0`.

The client owns the range data and gates ability fire *client-side*; there is no server "out of range" wire response. The range field is read straight from cooked data — no entity property carries it.

[...]

PR #368, opened 2026-05-24
feat(cell): NPC ability buckets + auto-aggro + Castle drone encounter (closes #342)

### LINKED ISSUE #342: feat(cell): NPC ability buckets + three-bucket AI selection + set_aggression auto-aggro (Castle Cellblock drone)
## Plain English

The "Prisoner retrieval unit" drone in Castle_CellBlock (tag `ArmYourself_PrisonerRetrievalUnit`, template 4) is supposed to spring to life when the player snatches the Ambernol vial during Mission 639 ("Find Ambernol"), shoot the player with a slow zat-like projectile ("Energy Shock", ability 221) from its fixed perch, and never chase — its leash is its spawn point. Today only the leash works. The drone doesn't aggro, doesn't fire the right ability, and the chain that's supposed to wake it up is missing the canonical threat seed and the player's reaction dialog.

Three independent gaps in the Rust runtime + two content gaps in the curated chain need to land together for the encounter to play the way it did in 2009.

### The drone

[`db/resources/Entities/Seed/entity_templates.sql:7`](../blob/main/db/resources/Entities/Seed/entity_templates.sql#L7) — template 4:

```sql
INSERT INTO entity_templates (template_id, ..., body_set, ..., ability_set_id, ...) VALUES
  (4, ..., 'MOB_CA_DroneTank.BS_MOB_DroneFlyer', ..., 2, ...);
```

[`db/resources/Worlds/Seed/spawnlist.sql:45`](../blob/main/db/resources/Worlds/Seed/spawnlist.sql#L45) — fixed spawn, leashed in place:

```sql
INSERT INTO spawnlist (spawn_id, x, y, z, heading, world_id, template_id, tag, set_name, is_stationary)
VALUES (10, -220.257, 66.744, -121.375, 4.712, 12, 4, 'ArmYourself_PrisonerRetrievalUnit', NULL, true);
```

### The ability

[`db/resources/Abilities/Seed/ability_sets.sql:9`](../blob/main/db/resources/Abilities/Seed/ability_sets.sql#L9) and [`ability_set_abilities.sql:7`](../blob/main/db/resources/Abilities/Seed/ability_set_abilities.sql#L7):

```sql
INSERT INTO ability_sets (ability_set_id, description) VALUES (2, 'Prisoner retrieval unit ability set');
INSERT INTO ability_set_abilities (ability_set_id, ability_id) VALUES (2, 221);
```

[`db/resources/Abilities/Seed/abilities.sql:5069`](../blob/main/db/resources/Abilities/Seed/abilities.sql#L5069) — ability 221 "Energy Shock":

```sql
INSERT INTO abilities (ability_id, name, ..., type_id, cooldown, ..., is_ranged, ..., velocity, ..., effect_ids, ..., event_set_id, required_ammo, ...) VALUES
  (221, 'Energy Shock', ..., 'ABILITY_TYPE_DD', 2, ..., true, ..., 100, ..., '{264}', ..., 802, 0, ...);
```

`ABILITY_TYPE_DD` direct damage, ranged, slow-velocity projectile, 2s cooldown, 0 ammo. The "zat-like" look comes from effect 264 + event_set 802.

### Canonical Python ordering — `FindAmbernol.py`

[`deprecated/python/cell/missions/Castle_CellBlock/FindAmbernol.py:151-159`](../blob/main/deprecated/python/cell/missions/Castle_CellBlock/FindAmbernol.py#L151) — `interactCb` subscribed to `entity.interact.tag::ArmYourself_AmbernolVial`:

```python
def interactCb(args):
    self.n1_var_Player = self.owner
    self.n1_propagator_Player()
    status = self.n5_var_Player.missions.getStepStatus(639, 2145)
    if status == Constants.MISSION_Active:
        self.n3_trigger_In()                                # see below
        self.n4_var_Player.missions.advance(639, 2144)
    return True
```

`n3_trigger_In` (lines 21-35) → `n6` (destroy) → `n8`/`n9` (aggression + threat) → `displayDialog(2297)` → `n19` (sequence). Combined linear order, with file:line anchors:

| # | Action | Python line | Maps to chain action |
|---|---|---|---|
| 1 | `inventory.pickedUpItem(19, 1)` | `FindAmbernol.py:25` | `add_item 19 (container 0, qty 1)` |
| 2 | `destroyCellEntity(vial)` | `FindAmbernol.py:63` | `destroy_entity ArmYourself_AmbernolVial` |
| 3 | `drone.setAggression(1)` | `FindAmbernol.py:99` | `set_aggression ArmYourself_PrisonerRetrievalUnit {"level":1}` |
| 4 | `drone.threatGenerated(playerId, 1000)` | `FindAmbernol.py:103` | **missing in chain 1032** |
| 5 | `displayDialog(2297)` | `FindAmbernol.py:32` | **missing in chain 1032** |
| 6 | `playSequence(10001, playOnWitnesses=True)` | `FindAmbernol.py:125-127` | `play_sequence 10001` |
| 7 | `missions.advance(639, 2144)` | `FindAmbernol.py:158` | `advance_step 639 2144` |

And the kill-the-drone callback at `FindAmbernol.py:170-173`:

```python
def deadCb(args):
    self.n30_var_Player.missions.advance(639, 2343)
    return True
self.owner.space.subscribe("entity.dead.tag::" + 'ArmYourself_PrisonerRetrievalUnit', deadCb, once=False)
```

Already covered by chain 1033 ✅.

### Curated chain 1032 today

[`db/resources/Content/Seed/castle_cellblock_chains.sql:329-345`](../blob/main/db/resources/Content/Seed/castle_cellblock_chains.sql#L329):

```sql
INSERT INTO content_triggers VALUES (1032, 'interact_tag', 'ArmYourself_AmbernolVial', 'player', false, 0);
INSERT INTO content_conditions VALUES (1032, 'step_status', 639, '2145', 'eq', 'active', 0);
INSERT INTO content_actions VALUES
  (1032, 'add_item',         19,   NULL, '{"container": 0, "qty": 1}',     0, 0),
  (1032, 'destroy_entity',   NULL, 'ArmYourself_AmbernolVial', '{}',       0, 1),
  (1032, 'set_aggression',   NULL, 'ArmYourself_PrisonerRetrievalUnit', '{"level": 1}', 0, 2),
  (1032, 'play_sequence',    10001, NULL, '{}',                            0, 3),
  (1032, 'advance_step',     639,  '2144', '{}',                           0, 4);
```

Two actions missing vs. the Python: `generate_threat 1000` (between `set_aggression` and `play_sequence`) and `display_dialog 2297` (between `generate_threat` and `play_sequence`). Both belong in this PR — same content area, same player-visible flow.

### Gap 1 — `set_aggression` stores a property nothing reads

[`crates/services/src/cell/content/executor/world.rs:64-80`](../blob/main/crates/services/src/cell/content/executor/world.rs#L64):

```rust
pub(super) fn set_aggression(entity_tag: String, agg_level: i32, ...) {
    if let Some(target_id) = space_mgr.find_entity_by_tag(entity_id, &entity_tag) {
        if let Some(target) = space_mgr.get_entity_mut(target_id) {
            target.properties.insert(
                "aggression".to_string(),
                PropertyValue::Int32(agg_level),
            );
        }
    }
}
```

Nothing in the AI tick reads `properties["aggression"]`. The NPC sits idle even after the chain fires. `grep -rn "aggression" crates/services/src/cell/` confirms no consumer.

### Gap 2 — Spawn unconditionally overwrites the template's ability set

[`crates/services/src/cell/space_manager/spawn.rs:38`](../blob/main/crates/services/src/cell/space_manager/spawn.rs#L38) and `:132`:

```rust
e.abilities.add_ability(super::super::combat::NPC_DEFAULT_ABILITY);
```

`template.ability_set_id` is never read. Every NPC, including the drone, ends up with exactly one ability: 592 (Pistol Shot).

### Gap 3 — NPC AI tick hardcodes the ability

[`crates/services/src/cell/service/npc_ai.rs:195-202`](../blob/main/crates/services/src/cell/service/npc_ai.rs#L195):

```rust
super::super::abilities::handle_use_ability(
    npc_id,
    combat::NPC_DEFAULT_ABILITY,    // always 592, regardless of NPC bucket
    target_id as i32,
    tx,
    space_mgr,
).await;
```

No bucket lookup, no cooldown check, no ammo check.

### What's already right

- Leash distance 50 units: [`combat/threat.rs:7-26`](../blob/main/crates/services/src/cell/combat/threat.rs#L7) ✅
- `is_stationary=true` honored — no pathfinding: [`npc_ai.rs:136`](../blob/main/crates/services/src/cell/service/npc_ai.rs#L136) ✅
- `entity_dead_tag` trigger wired (chain 1033 advances to step 2343 on drone death): [`content-engine/src/triggers.rs:213`](../blob/main/crates/content-engine/src/triggers.rs#L213) ✅
- `Action::SetAggression` parses + dispatches from chain — the *consumer* is missing, not the wiring ✅

## Proposed fix

Five changes — three runtime, two content. Land in one PR so the encounter doesn't half-work between commits.

### 1. `set_aggression(level=1)` actually drives combat

Extend [`world.rs::set_aggression`](../blob/main/crates/services/src/cell/content/executor/world.rs#L64) so that after storing the property, when `agg_level >= 1` it seeds threat on the triggering player by calling `combat::generate_threat`. The Python flow uses `setAggression` for the *behavior bit* (NPC will keep auto-acquiring new threats in sight) and `threatGenerated` for the *initial threat seed*. With chain 1032 fixed to add explicit `generate_threat 1000` (see content change below), `set_aggression` doesn't strictly need to seed threat itself — but it should still flip the NPC into a state where the AI tick scans for nearby players each tick.

Implementation:

- Store `aggression: i32` directly on `CellEntity` (not as a string-keyed property bag entry).
- In `npc_ai_tick`, when an NPC has `aggression > 0` and is in `Idle` state, scan AoI for player entities of opposing faction within sight range; seed minimal threat (e.g. `1.0`) on the closest one, which transitions to `Fighting`.
- Aggression persists across kills — a stationary drone with `aggression=1` re-aggros the next player who walks in.
- Aggression `0` is the default (passive — NPC only fights back when threatened).

Confidence: high. The Python's `setAggression(int)` semantics are documented in the Atrea Script Editor convention as "0 = passive, 1 = hostile to nearby enemies"; the canonical chain doesn't pass any other value.

### 2. Load per-template ability bucket on spawn

In the spawn path ([`space_manager/spawn.rs`](../blob/main/crates/services/src/cell/space_manager/spawn.rs)), when `template.ability_set_id` is non-null, query `ability_set_abilities WHERE ability_set_id = ?` and call `e.abilities.add_ability(id)` for each row. Fall back to `NPC_DEFAULT_ABILITY` only when the template has no ability set (defensive default, matches Python convention where un-assigned NPCs use the Pistol Shot moniker as a placeholder).

For the drone: template 4 → ability_set 2 → ability 221 (Energy Shock). The Cellblock Guard (template 15 → ability_set 1 → 592 Pistol Shot) preserves current behavior under the fallback path.

Load timing: ability bucket can be loaded eagerly at spawn (one-shot query per spawn) or lazily on first AI tick — eager is simpler and the data is static.

### 3. Three-bucket ability selection in the AI tick

Rather than the trivial "pick the first ability" — implement the full Python `chooseAbility` model from the start, since we'll need it the moment we have a second multi-ability NPC. Each tick, partition the NPC's ability list into three buckets:

- **usable** — off cooldown, has ammo (or `required_ammo == 0`), target in range/LOS
- **cooling** — on cooldown
- **needs_ammo** — out of ammo

Selection priority: pick from `usable`; if empty, hold fire and tick again next interval (cooldown will tick down). Among `usable`, prefer abilities with higher `threat_level_id` or — for a first pass — just round-robin. If all three buckets are empty (no abilities at all), fall back to `NPC_DEFAULT_ABILITY` so we never wedge.

Cooldowns: track `last_used_at: HashMap<ability_id, Instant>` on the NPC. Honor each ability's `cooldown` field (Energy Shock = 2s; lines up with the 2s AI tick interval, so the drone fires every tick).

This work should *not* duplicate #329 (per-ability range + min-range backup + re-tick on launch failure). The bucket selector reads `min_range`/`max_range` from the same `AbilityDef`; the in-range check #329 introduces is what gates the "target in range/LOS" pillar of the `usable` bucket. Either land #329 first and consume its API, or sequence its work into this PR — but the *range computation* itself belongs to #329, not here.

### 4. Chain 1032: add `generate_threat 1000`

Insert between sort_order 2 (`set_aggression`) and 3 (`play_sequence`):

```sql
(1032, 'generate_threat', NULL, 'ArmYourself_PrisonerRetrievalUnit', '{"threat_level": 1000}', 0, 3),
```

…and bump subsequent sort_orders. This is the explicit initial-threat seed that focuses the drone on the player who grabbed the vial, matching `FindAmbernol.py:103`.

### 5. Chain 1032: add `display_dialog 2297`

Insert before the play_sequence so the player's reaction line plays at the correct beat:

```sql
(1032, 'display_dialog', 2297, NULL, '{}', 0, 4),
```

Matches `FindAmbernol.py:32`. Dialog 2297 is the "what was that?" line; without it the moment plays silent.

### Out of scope for this PR

- **Per-ability range** — covered by #329. This PR depends on that landing (or being inlined) for the `usable` bucket's range gate.
- **Aggression UI override** (HOSTILE / SUSPICIOUS / NEUTRAL / FRIENDLY / DEFAULT) — covered by #330. That's a wire-out for nameplate color; orthogonal to combat-aggression-as-behavior.
- **Effect 264 stun semantics** — Energy Shock fires effect 264. If the effect carries a stun/slow component beyond direct damage, that's effect-system work; surface as a follow-up issue if testing shows the drone hits should pacify the player briefly.
- **Loot drop** — template 4 has `loot_table_id = NULL`. The drone drops nothing. Confirmed not a gap.

### Unit tests

- [ ] `set_aggression(level=1)` on an NPC tagged `T` with a player in AoI flips the NPC to `Fighting` on the next AI tick (with the player in the threat list).
- [ ] `set_aggression(level=0)` leaves the NPC `Idle`.
- [ ] An NPC with `aggression=1` that kills its current target re-aggros the next player who enters AoI.
- [ ] Template-driven bucket: spawning template 4 results in an NPC whose `abilities` contains 221 and *not* 592.
- [ ] Template-driven bucket fallback: spawning a template with `ability_set_id = NULL` results in an NPC with `NPC_DEFAULT_ABILITY` (592).
- [ ] Three-bucket selector: NPC with abilities [221, 592], 221 on cooldown → selector returns 592.
- [ ] Three-bucket selector: NPC with abilities [221, 592], 221 needs ammo (mock) → selector returns 592.
- [ ] Three-bucket selector: all abilities cooling → selector returns `None`, NPC holds fire.
- [ ] Cooldown honored: NPC firing ability 221 twice within 2s fires only once (second tick falls through to next-usable or holds).

### Wire-format tests

- [ ] Byte-exact `onAbilityLaunched(221, ...)` packet emitted when the drone fires (replaces today's `592` emission). Pin via golden bytes in `crates/services/src/cell/abilities/...` tests.

### Chain-replay tests

- [ ] Mission 639 replay: after `interact_tag ArmYourself_AmbernolVial` with step 2145 active, the resolved action sequence is exactly the 7 actions in the Python order table above (add_item → destroy_entity → set_aggression → generate_threat → display_dialog → play_sequence → advance_step).
- [ ] Regression guard for the missing actions: a chain that produces only 5 actions (no `generate_threat`, no `display_dialog`) fails the replay assertion. This is the bug-shape guard.

### Live-DB tests

- [ ] Spawn template 4 + run chain 1032 + advance one AI tick → drone is in `Fighting`, target = the player, last `EntityMethodCall` is `onAbilityLaunched(221)`.
- [ ] Stationary leash preserved: spawn the drone, aggro it, walk the player past `LEASH_DISTANCE` from the spawn point → next AI tick flips drone to `Leashing`, then `Idle`, threat list cleared, health restored.

### Manual UAT (Castle Cellblock walkthrough)

- [ ] Accept Mission 622, loot Capt. Frost, equip pistol → mission 622 completes, 639 accepts.
- [ ] Walk into Region11 (step 2117 → 2145).
- [ ] Right-click the Ambernol vial. Verify in order: item appears in mission inventory, vial disappears, drone visibly aggros (turret swings to face you), Net'an dialog 2297 plays, play_sequence 10001 plays (door opens / cinematic / whatever 10001 does), step advances to 2144.
- [ ] Drone fires Energy Shock at you (slow blue zat-shaped bolt, not the pistol muzzle flash). Cooldown ~2s between shots.
- [ ] Walk well past 50 units from the drone's perch — drone leashes (stops firing, returns to passive). Walk back in range — drone re-aggros (aggression=1 still set).
- [ ] Kill the drone — step advances to 2343 (chain 1033, no change in this PR).
- [ ] Use the Ambernol vial — mission 639 completes, 640 accepts.

## Related

- **Depends on #329** for the per-ability range/LOS gate that the `usable` bucket needs. Land #329 first or fold it in.
- **Orthogonal to #330** — that's UI/nameplate color via `onAggressionOverrideUpdate`, this is combat-behavior aggression that drives the AI tick. Different field, different consumer.
- **Spec authority for the chain ordering** is the original 2009 server's `FindAmbernol.py` (Castle_CellBlock mission set). The Python's seven-action ordering is what the client expects (dialog before sequence, threat seed before VO, etc.). The auto-converted chain 1032 dropped two of those actions; the curated chain is what the player actually exercises.
- Castle_CellBlock is also exercised by the `castle_cellblock_chains.sql` chain pipeline (#216 archetype-routing fix, mission 622 / 638 / 639 / 640 / 641 / 680 chain) — this PR's content edits to chain 1032 sit inside the curated file and won't conflict with the auto-export.

Cadacious:
## Summary

Brings the Castle_CellBlock prisoner-retrieval drone encounter to parity with the canonical Python `FindAmbernol.py`. Five gaps closed — three runtime, two content — landing together so the encounter doesn't half-work between commits.

### Runtime — three independent gaps

| Gap | Was | Now |
|---|---|---|
| `set_aggression` content action | Stored an int in the property bag; nothing read it | Writes `CellEntity::aggression`. The AI tick adds an Idle sweep — for NPCs with `aggression > 0`, scan AoI witnesses, seed threat on the closest opposing-faction player. Aggression persists across kills (stationary drones re-aggro the next player who walks in). |
| Per-template ability seeding | `spawn_npc_from_record` unconditionally added `NPC_DEFAULT_ABILITY` (Pistol Shot 592). Every NPC fired the pistol regardless of template. | `load_spawns_from_db` `LEFT JOIN`s `ability_set_abilities` via `array_agg + COALESCE-to-empty-array`. `SpawnRecord::ability_ids` is non-empty for templates with `ability_set_id`; spawn uses those. Falls back to `NPC_DEFAULT_ABILITY` only when the template has no set. Drone (template 4 → set 2 → `[221]` Energy Shock) now spawns with the right loadout. |
| Three-bucket ability selection | `npc_ai_fight` hardcoded `combat::NPC_DEFAULT_ABILITY` in `handle_use_ability` | `choose_npc_ability` partitions known abilities into **usable** / **cooling** / **needs_ammo** (mirrors `python/cell/SGWMob.py:chooseAbility`). Returns first usable; `None` when all gated → AI holds fire and retries next 2s tick. |

Per-ability range gating is deliberately deferred to #329 (the issue body explicitly says so); until that lands, the existing flat `NPC_ATTACK_RANGE = 30.0` gate before the selector remains the range pillar.

### Content — chain 1032 (`db/resources/Content/Seed/castle_cellblock_chains.sql`)

Two missing actions added between `set_aggression` and `play_sequence`:

- `generate_threat 1000` — focuses the drone on the player who grabbed the vial (matches `FindAmbernol.py:103`). Without it, auto-aggro from `set_aggression` alone picks the closest player, which is wrong in a party.
- `display_dialog 2297` — Net'an's reaction line (matches `FindAmbernol.py:32`). Without it the moment plays silent.

Final ordering matches the canonical Python sequence: `add_item → destroy_entity → set_aggression → generate_threat → display_dialog → play_sequence → advance_step`.

## Test plan

All from #342's test plan. **854 services tests pass**, **160 entity tests pass**, full workspace clippy clean, fmt clean.

### Unit tests (10 new)

- `npc_ai::tests::idle_npc_with_aggression_aggros_opposing_player` — `aggression=1` + opposing-faction witness → NPC enters `Fighting` with player on threat list.
- `npc_ai::tests::idle_npc_without_aggression_stays_idle` — passive default holds.
- `npc_ai::tests::aggression_skips_same_faction_witnesses` — same-faction witness ignored.
- `npc_ai::tests::selector_skips_cooling_ability` — cooldown gate participates in partition.
- `npc_ai::tests::selector_skips_needs_ammo_ability` — `required_ammo > 0` routes to the needs_ammo bucket.
- `npc_ai::tests::selector_returns_none_when_all_cooling` — AI holds fire instead of misfiring.
- `npc_ai::tests::selector_falls_back_to_default_when_no_abilities` — empty-bucket fallback.
- `spawner::tests::spawn_record_with_ability_ids_skips_default_fallback` — template-driven bucket does NOT also pick up `NPC_DEFAULT_ABILITY`.
- `spawner::tests::spawn_record_with_empty_ability_ids_falls_back_to_default` — empty bucket falls back.
- `executor::tests::set_aggression_level_one_writes_entity_field` — content action writes the canonical field, not the old property-bag entry.

### Chain-replay test (live-DB)

- `chain_replay_tests::mission_639::chain_1032_seven_actions_match_python_ordering` — pins the seven-action signature in order, asserts `GenerateThreat` magnitude 1000, asserts `DisplayDialog` id 2297. Skips cleanly when `DATABASE_URL` is unset (`require_db_or_skip!`).

### Adversarial verification

Reverting each runtime gap individually flips the relevant guard to red:

| Revert | Guard that fails |
|---|---|
| Comment out `target.aggression = agg_level` | `set_aggression_level_one_writes_entity_field` → `0 != 1` |
| Restore unconditional `add_ability(NPC_DEFAULT_ABILITY)` | `spawn_record_with_ability_ids_skips_default_fallback` |
| Comment out the `generate_threat` seed call | `idle_npc_with_aggression_aggros_opposing_player` → `Idle != Fighting` |

Bug-shape-pinning verified for all three.

## Sidewalk

[`docs/gameplay/spawn-system.md`](docs/gameplay/spawn-system.md) gains two short sections — **NPC Ability Bucket and Selection** and **NPC Aggression (behavior bit)** — documenting both the spawn-time data flow and the runtime AI gate in one place. [`docs/content/mission-chains.md`](docs/content/mission-chains.md) already described the canonical chain 1032 flow correctly; the SQL is what was lagging.

## Out of scope

- #329 per-ability range / min-range backup / re-tick on launch failure — the issue explicitly says range belongs there.
- #330 aggression-override (UI nameplate color) — distinct field, distinct consumer.
- Effect 264 stun semantics — surface separately if testing shows the drone hits should pacify briefly.

## Closes

- Closes #342

## Summary by CodeRabbit

* **New Features**
  * Aggression-driven NPC behavior enables idle NPCs to auto-aggro and seed threat, causing dynamic combat transitions
  * NPC abilities are now seeded from templates with a defensive fallback to a default ability; ability selection respects cooldowns

* **Bug Fixes**
  * Corrected Ambernol vial encounter action sequence and ordering

* **Documentation**
  * Added spawn-system and NPC aggression/ability behavior docs

* **Tests**
  * Added regression and unit tests covering aggression, ability selection, and encounter ordering

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/368?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/entity/src/cell_entity/mod.rs
+++ b/crates/entity/src/cell_entity/mod.rs
@@ -407,6 +407,19 @@ pub struct CellEntity {
     /// Pin this NPC to its spawn position. AI will attack when the target
     /// is in range + LOS but never pathfind. Loaded from `spawnlist.is_stationary`.
     pub is_stationary: bool,
+    /// NPC behavior-aggression level. `0` = passive (only fights back when
+    /// threatened, the default). `≥1` = hostile-on-sight (the AI idle tick
+    /// scans the NPC's witnesses for opposing-faction players and seeds
+    /// threat to wake the NPC up). Set by the `set_aggression` content
+    /// action when a mission chain wants an NPC to start hunting — e.g.,
+    /// chain 1032 flips the Castle_CellBlock prisoner-retrieval drone to
+    /// `1` when the player grabs the Ambernol vial. Persists across kills
+    /// so a stationary drone re-aggros the next player who walks in.
+    ///
+    /// Distinct from `aggression_override` (UI nameplate color, tracked
+    /// separately under #330). This field drives combat behavior; that one
+    /// drives the client's friend/foe indicator.
+    pub aggression: i32,
 
     // ── Saved mission state (for re-login) ────────────────────────────────────
     /// Saved missions loaded from DB, to be populated before content engine fires.
@@ -584,6 +597,7 @@ impl CellEntity {
             nav_path: VecDeque::new(),
             move_speed: 0.6, // ~0.6 world units per 100ms tick = 6 units/sec
             is_stationary: false,
+            aggression: 0,
             saved_missions_loaded: false,
             loot_table_id: None,
             loot: Vec::new(),
--- a/crates/entity/src/cell_entity/mod.rs
+++ b/crates/entity/src/cell_entity/mod.rs
@@ -416,9 +416,8 @@ pub struct CellEntity {
     /// `1` when the player grabs the Ambernol vial. Persists across kills
     /// so a stationary drone re-aggros the next player who walks in.
     ///
-    /// Distinct from `aggression_override` (UI nameplate color, tracked
-    /// separately under #330). This field drives combat behavior; that one
-    /// drives the client's friend/foe indicator.
+    /// Distinct from any UI-layer aggression-override (nameplate color);
+    /// this field drives combat behavior, not the friend/foe indicator.
     pub aggression: i32,
 
     // ── Saved mission state (for re-login) ────────────────────────────────────
--- a/crates/services/src/cell/content/chain_replay_tests/mission_639.rs
+++ b/crates/services/src/cell/content/chain_replay_tests/mission_639.rs
@@ -111,6 +111,23 @@ async fn chain_1032_seven_actions_match_python_ordering() {
 
     // Per-action invariants the encounter relies on:
     //
+    // - SetAggression must target the drone tag with level=1. Without the
+    //   level, auto-aggro never kicks in; without the tag, the wrong NPC
+    //   gets the behavior bit.
+    let aggression = chain
+        .actions
+        .iter()
+        .find_map(|a| match a {
+            Action::SetAggression { entity_tag, level } => Some((entity_tag.as_str(), *level)),
+            _ => None,
+        })
+        .expect("chain 1032 must include a SetAggression action");
+    assert_eq!(
+        aggression,
+        ("ArmYourself_PrisonerRetrievalUnit", 1),
+        "chain 1032 SetAggression must target the drone with level=1",
+    );
+
     // - GenerateThreat must target the drone tag with magnitude 1000 — this
     //   is what focuses the drone on the player who grabbed the vial rather
     //   than whichever player happens to be closest (auto-aggro from
--- a/crates/services/src/cell/content/chain_replay_tests/mission_639.rs
+++ b/crates/services/src/cell/content/chain_replay_tests/mission_639.rs
@@ -3,6 +3,12 @@
 //! accepts mission 640. The `remove_item` action is the load-bearing piece
 //! — without it, the player keeps the vial after using it (and any chain
 //! gated on "no longer holds vial" stays stuck).
+//!
+//! Chain 1032 is the vial-grab encounter setup: pick up the vial, destroy
+//! it on the world, wake the prisoner-retrieval drone, focus its threat on
+//! the triggering player, play Net'an's reaction VO, fire the cinematic,
+//! advance the mission step. The seven-action ordering is canonical per
+//! `python/cell/missions/Castle_CellBlock/FindAmbernol.py`.
 
 use super::super::engine_loader::load_single_chain_for_test;
 use crate::test_support::require_db_or_skip;
@@ -37,5 +43,106 @@ async fn chain_1034_includes_remove_item_for_ambernol() {
          so the ambernol vial is consumed on use; got {remove_vial_actions} \
          matching actions. Full action list: {:?}",
         chain.actions,
+    );
+}
+
+/// Chain 1032: the seven-action vial-grab sequence. Replay guard pinned to
+/// the canonical ordering from `FindAmbernol.py:21-35 + :151-159`:
+///
+/// 1. `add_item 19`        — `inventory.pickedUpItem(19, 1)`
+/// 2. `destroy_entity`     — `destroyCellEntity(vial)`
+/// 3. `set_aggression 1`   — `drone.setAggression(1)` (durable behavior bit)
+/// 4. `generate_threat`    — `drone.threatGenerated(player, 1000)` (focuses drone)
+/// 5. `display_dialog 2297`— Net'an's reaction VO
+/// 6. `play_sequence 10001`— cinematic
+/// 7. `advance_step 2144`  — mission step transition
+///
+/// Two of these were historically missing (`generate_threat` and
+/// `display_dialog`) — they were dropped during the auto-conversion of the
+/// Python chain and surfaced as "drone doesn't aggro until 2s later" and
+/// "Net'an's line never plays" in the encounter walkthrough. This test
+/// guards against regression of the curated re-insert; the bug shape the
+/// guard catches is a future seed change that drops either action back.
+#[tokio::test]
+async fn chain_1032_seven_actions_match_python_ordering() {
+    use cimmeria_content_engine::actions::Action;
+
+    let pool = require_db_or_skip!();
+    let chain = load_single_chain_for_test(&pool, 1032)
+        .await
+        .expect("DB query for chain 1032 must succeed")
+        .expect("chain 1032 must exist in seeded content_chains");
+
+    // Build a label-only signature so a future addition of a sibling
+    // action surfaces as a mismatch on this list rather than on a
+    // per-action structural comparison (which would also break on cosmetic
+    // changes like swapping container ID or threat magnitude).
+    let signature: Vec<&'static str> = chain
+        .actions
+        .iter()
+        .map(|a| match a {
+            Action::GrantItem { .. } => "add_item",
+            Action::DestroyTaggedEntity { .. } => "destroy_entity",
+            Action::SetAggression { .. } => "set_aggression",
+            Action::GenerateThreat { .. } => "generate_threat",
+            Action::DisplayDialog { .. } => "display_dialog",
+            Action::PlaySequence { .. } => "play_sequence",
+            Action::AdvanceStep { .. } => "advance_step",
+            _ => "OTHER",
+        })
+        .collect();
+
+    let expected = vec![
+        "add_item",
+        "destroy_entity",
+        "set_aggression",
+        "generate_threat",
+        "display_dialog",
+        "play_sequence",
+        "advance_step",
+    ];
+    assert_eq!(
+        signature, expected,
+        "chain 1032 action ordering drifted from the canonical Python sequence. \
+         Got {signature:?}, expected {expected:?}. \
+         Full action list: {:?}",
+        chain.actions,
+    );
+
+    // Per-action invariants the encounter relies on:
+    //
+    // - GenerateThreat must target the drone tag with magnitude 1000 — this
+    //   is what focuses the drone on the player who grabbed the vial rather
+    //   than whichever player happens to be closest (auto-aggro from
+    //   set_aggression alone picks the closest, which is wrong in a party).
+    let threat = chain
+        .actions
+        .iter()
+        .find_map(|a| match a {
+            Action::GenerateThreat {
+                entity_tag,
+                threat_level,
+            } => Some((entity_tag.as_deref(), *threat_level)),
+            _ => None,
+        })
+        .expect("chain 1032 must include a GenerateThreat action");
+    assert_eq!(
+        threat,
+        (Some("ArmYourself_PrisonerRetrievalUnit"), 1000),
+        "chain 1032 GenerateThreat must focus the drone with threat=1000",
+    );
+
+    // - DisplayDialog must be dialog 2297 (Net'an's reaction line).
+    let dialog_id = chain
+        .actions
+        .iter()
+        .find_map(|a| match a {
+            Action::DisplayDialog { dialog_id } => Some(*dialog_id),
+            _ => None,
+        })
+        .expect("chain 1032 must include a DisplayDialog action");
+    assert_eq!(
+        dialog_id, 2297,
+        "chain 1032 DisplayDialog must reference dialog 2297 (Net'an reaction)",
     );
 }
--- a/crates/services/src/cell/content/executor/tests.rs
+++ b/crates/services/src/cell/content/executor/tests.rs
@@ -635,3 +635,49 @@ async fn cross_world_teleport_action_with_unknown_entity_dispatches_gate_travel(
          cell entity is absent — base may still hold the connection"
     );
 }
+
+/// `Action::SetAggression { level: 1 }` writes the `aggression` field on
+/// the tagged NPC's `CellEntity` so the AI idle tick can wake it up. Bug
+/// shape: the previous implementation stored an unread `"aggression"`
+/// property-bag entry; this guard pins that the canonical field is now
+/// the source of truth.
+#[tokio::test]
+async fn set_aggression_level_one_writes_entity_field() {
+    let mut mgr = make_space_mgr();
+    // Tagged NPC the chain will target.
+    mgr.spawn_npc(101, "Agnos", [10.0, 0.0, 10.0], [0.0; 3])
+        .unwrap();
+    if let Some(npc) = mgr.get_entity_mut(101) {
+        npc.tag = Some("Drone".to_string());
+        assert_eq!(npc.aggression, 0, "fresh NPC must start passive");
+    }
+    // Player triggering the chain — must be co-located so
+    // `find_entity_by_tag` resolves "Drone" against the same space.
+    mgr.create_entity(1, "Agnos", [0.0; 3], [0.0; 3]).unwrap();
+    if let Some(p) = mgr.get_entity_mut(1) {
+        p.is_player = true;
+        p.player_id = Some(42);
+    }
+
+    let (tx, _rx) = mpsc::channel(8);
+    let engine = ChainEngine::new();
+    let resolved = ResolvedActions {
+        params: std::collections::HashMap::new(),
+        actions: vec![(
+            1032,
+            Action::SetAggression {
+                entity_tag: "Drone".to_string(),
+                level: 1,
+            },
+        )],
+    };
+
+    execute_actions(resolved, 1, 42, &tx, &mut mgr, &engine).await;
+
+    assert_eq!(
+        mgr.get_entity(101).unwrap().aggression,
+        1,
+        "SetAggression level=1 must write aggression=1 on the target — \
+         the AI idle tick reads this field directly, no property-bag lookup",
+    );
+}
--- a/crates/services/src/cell/content/executor/world.rs
+++ b/crates/services/src/cell/content/executor/world.rs
@@ -58,9 +58,19 @@ pub(super) async fn set_interaction_type(
     }
 }
 
-/// `Action::SetAggression` — store the int32 aggression level in the
-/// tagged entity's properties. Read by combat AI; not broadcast (each
-/// witness reads on its own AoI create).
+/// `Action::SetAggression` — set the tagged NPC's behavior-aggression
+/// level (`0` = passive, `≥1` = hostile-on-sight). The AI idle tick reads
+/// this directly off the entity (no property-bag lookup) and seeds threat
+/// on opposing-faction witnesses when `aggression > 0`.
+///
+/// The Python flow uses `setAggression` for the *durable behavior bit*
+/// and a separate `threatGenerated` for the *initial threat seed* — see
+/// `python/cell/missions/Castle_CellBlock/FindAmbernol.py:99-103`. Chain
+/// 1032 follows the same pattern: this action sets the behavior, then a
+/// `generate_threat` action focuses the NPC on the player who triggered
+/// the chain. Without that explicit seed the drone would aggro on the
+/// next idle tick anyway, but the seed delivers the correct frame
+/// ordering (drone faces the player immediately, not 2s later).
 pub(super) fn set_aggression(
     entity_tag: String,
     agg_level: i32,
@@ -71,10 +81,7 @@ pub(super) fn set_aggression(
     if let Some(target_id) = space_mgr.find_entity_by_tag(entity_id, &entity_tag) {
         tracing::debug!(entity_id, %entity_tag, target_id, agg_level, chain_id, "Content: set aggression");
         if let Some(target) = space_mgr.get_entity_mut(target_id) {
-            target.properties.insert(
-                "aggression".to_string(),
-                cimmeria_entity::base_entity::PropertyValue::Int32(agg_level),
-            );
+            target.aggression = agg_level;
         }
     }
 }
--- a/crates/services/src/cell/service/npc_ai.rs
+++ b/crates/services/src/cell/service/npc_ai.rs
@@ -7,16 +7,10 @@ use tokio::sync::mpsc;
 use super::super::messages::CellToBaseMsg;
 use super::super::space_manager::SpaceManager;
 
-/// NPC AI tick — drive the three Fighting/Leashing/Idle-with-aggression
-/// paths. For NPCs in Fighting: find top threat target, check leash, attack
-/// via the three-bucket selector. For NPCs in Leashing: reset to Idle when
-/// close enough to spawn point, restore health. For NPCs in Idle whose
-/// `aggression > 0`: scan AoI witnesses for opposing-faction players, seed
-/// threat on the closest one (transitions to Fighting on the next tick).
-///
-/// The Idle-with-aggression path is what makes `set_aggression(level=1)`
-/// content actions actually drive combat — see
-/// `python/cell/missions/Castle_CellBlock/FindAmbernol.py:99` and
+/// NPC AI tick — drives Fighting, Leashing, and Idle-with-aggression
+/// NPCs. The `Idle` filter on `aggression > 0` is what makes the
+/// `set_aggression` content action actually trigger combat — without it
+/// the action would be a behavior bit nothing read. See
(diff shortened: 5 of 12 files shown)


MyPalClara:

| **G1** | `set_aggression` dual state | The PR adds `CellEntity::aggression` and the AI tick reads it. Does `set_aggression` in `executor/world.rs` still write to the property bag too? If so, there's redundant state. If not, verify the property bag write was removed. |
| **G2** | `choose_npc_ability` fallback edge case | Falls back to `NPC_DEFAULT_ABILITY` when bucket is empty. But what if the NPC has abilities but ALL are gated (cooling + needs_ammo)? The PR says "returns `None`, AI holds fire" — verify the fallback doesn't fire in that case. |
| **G3** | Aggression persistence across respawn | "Aggression persists across kills" — but what about NPC respawn? If the drone respawns (spawn timer), does it inherit `aggression=1` from the previous instance, or reset to 0? The spawn path should reset to 0 (fresh entity), but verify. |
| **G4** | `generate_threat` magnitude dominance | Chain 1032 uses `threat_level: 1000`. The auto-aggro from `set_aggression` seeds `1.0`. The doc says "explicit `generate_threat 1000` dominates" — is 1000 >> 1.0 enough to guarantee the chain-triggered player is always the target, even if another player is closer? Verify the threat comparison logic. |

MyPalClara:

| **G5** | `LEFT JOIN` on spawn query | `load_spawns_from_db` now `LEFT JOIN`s `ability_set_abilities`. If the join produces multiple rows per spawn (template has multiple abilities), does `array_agg` correctly collapse them? Verify the query doesn't duplicate spawn records. |
| **G6** | Cooldown tracking location | The PR mentions cooldown gating but doesn't specify where the per-ability cooldown state lives. Is it on `CellEntity::abilities`? On a separate map? Verify the cooldown state is correctly initialized and persisted. |
| **G7** | Range gating deferred to #329 | The PR explicitly defers per-ability range to #329. But the `usable` bucket definition includes "target in range/LOS" — is that check actually implemented, or is it using the flat `NPC_ATTACK_RANGE = 30.0` gate? If the latter, document that the bucket's range pillar is a placeholder. |

MyPalClara:
## Category 3: Missing Test Coverage (G8-G10)

| **G8** | No test for multi-ability NPC | The tests cover single-ability (drone) and fallback (no abilities). But what about an NPC with 3 abilities, 2 cooling, 1 usable? Verify the selector picks the usable one. |
| **G9** | No test for `generate_threat` magnitude dominance | The chain-replay test verifies `threat_level=1000` is present, but doesn't verify it dominates auto-aggro's `1.0` in a party scenario. Consider a test with two players (one closer, one farther) that verifies the chain-triggered player is targeted. |
| **G10** | No wire-format test for `onAbilityLaunched(221)` | Issue #342's test plan includes "Byte-exact `onAbilityLaunched(221, ...)` packet emitted." The PR doesn't mention this test. Is it present? |

MyPalClara:
## Category 4: Documentation + Summary (G11-G12)

| **G11** | `spawn-system.md` cooldown tracking | The doc explains ability bucket selection but not where cooldown state lives. Add a one-liner. |
| **G12** | #329 dependency resolution | Issue #342 says "Depends on #329 for per-ability range." Is #329 merged? If not, document how the `usable` bucket's range check works today. |

---

## Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟡 Verify | G1, G2, G3, G5, G6 | ~30 min |
| 🟡 Consider | G4, G9 (party scenario test) | ~45 min |
| 🟢 Quick wins | G11, G12 (doc updates) | ~15 min |
| ⚪ Nice-to-have | G7, G8, G10 | ~1 hr |

**Total: 12 gaps, ~2.5 hours of follow-up work.**

---

## Issue #342 Coverage ✅

All 5 gaps from the issue are addressed:
- ✅ `set_aggression` drives combat
- ✅ Per-template ability bucket  
- ✅ Three-bucket selector
- ✅ Chain 1032 `generate_threat`
- ✅ Chain 1032 `display_dialog`

**The Castle drone encounter is complete.** 🎯

Cadacious:
Pushed [c0190f1c](https://github.com/SandboxServers/Cimmeria/commit/c0190f1c) addressing all valid review findings across CodeRabbit, Copilot, and Clara.

## Critical correctness bug — fixed

Both CodeRabbit and Copilot caught this independently, and it would have been a **showstopper regression on every existing NPC**:

`choose_npc_ability` treated `required_ammo > 0` as "needs_ammo" and skipped. The seeded `NPC_DEFAULT_ABILITY = 592` (Pistol Shot) has `required_ammo = 1` — so every stock NPC's only ability would have been permanently unfireable, and every NPC in the game would have held fire forever. Fix: drop the ammo gate from the NPC selector. NPCs have infinite ammo by convention; the ammo check at the dispatch site (`use_ability.rs:150,160,185`) is player-only.

The new `selector_picks_ammo_bearing_ability_for_npc` regression guard explicitly seeds `ability_defs[592].required_ammo = 1` so the fixture genuinely models the bug — without that seed, `ability_defs.get(&592)` returns `None` and a re-introduced ammo gate would silently fall through to `required_ammo = 0` and pass.

**Adversarial verification:** re-introducing the buggy gate flips the new guard to red (`left: None, right: Some(592)`).

## Convention violations — fixed

- Stripped `#329` / `#330` issue refs from Rust source comments and docs (CodeRabbit, Copilot).
- `"unspeced"` → `"unspecified"` in `spawn.rs` comment and `spawn-system.md` (Copilot).
- Test message `"round-robin"` → `"alternate based on sort order"` — the selector is deterministic, not RR (Copilot).
- Corrected the SQL comment in `spawner/npcs.rs`: actual query is a correlated subquery with `COALESCE`, not a `LEFT JOIN` with `FILTER` (Copilot ×2).

## Test hardening — done

- Chain 1032 replay test now asserts `SetAggression` target+level explicitly (CodeRabbit). Was only checking presence/order; a payload drift could have passed.
- New `selector_skips_multiple_cooling_abilities` test exercises a 3-ability bucket (Clara G8) — verifies the `find` walks past every cooling entry rather than stopping at the first.

## Refactor — done

- Removed the duplicated `aggression > 0` defensive check inside `npc_ai_idle_auto_aggro` (CodeRabbit). The outer `npc_ai_tick` filter is the single source of truth — repo convention: trust internal contracts.
- Trimmed narrative comments around `choose_npc_ability` to keep only the non-obvious "why" lines (CodeRabbit).
- Moved the inline `#[cfg(test)] mod tests` block out of `npc_ai.rs` (now 368 lines, under 500 soft cap) into the existing sibling test file at `service/tests/npc_ai.rs` (CodeRabbit). The new aggression tests route through `npc_ai_tick` (the production filter path); selector tests call `pub(super) choose_npc_ability` directly.

## Doc updates (Clara G3, G11)

- `spawn-system.md` notes cooldown state lives on `AbilityManager::ability_cooldowns` and is cleared by the leash tick.
- Adds a sentence that `CellEntity::new` initializes `aggression: 0`, so respawn-eligible NPCs start passive again unless the mission chain re-fires `set_aggression` against the fresh spawn.

## Clara findings verified already correct (no fix needed)

| Gap | Reason |
|---|---|
| **G1** | `set_aggression` writes only to `CellEntity::aggression`; no property bag entry remains. Verified by `grep aggression crates/services/src/cell/content/executor/world.rs`. |
| **G2** | Distinct paths: `known_count() == 0` → fallback; non-empty-but-all-gated → `None`. The new `selector_returns_none_when_all_cooling` test pins this. |
| **G4** | `threat_list.max_by` is numerical — `1000 >> 1.0` is guaranteed dominance. |
| **G5** | The SQL is a correlated subquery (one row per spawn), not a JOIN; no possible row multiplication. |
| **G7** | Range-gating-deferred is already called out in `spawn-system.md`. |
| **G12** | Issue ref removed from doc per Copilot/CodeRabbit — the deferred-range note now stands on its own without linking. |

## Clara findings skipped (with reason)

- **G9** (party threat-dominance runtime test): the chain-replay test already pins the `1000` magnitude. A full multi-player runtime scenario would duplicate the existing `threat_list` infrastructure tests.
- **G10** (byte-exact `onAbilityLaunched(221)` test): the general `onAbilityLaunched` wire format is regression-guarded elsewhere; a per-ability byte test would be theatre.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --exclude {GUI crates} --all-targets -- -D warnings` clean
- **855 services tests pass** (was 854; +1 for the new multi-cooling test)

PR #369, opened 2026-05-24
fix(combat): credit quest kills from auto-cycle and other tick paths (#367)

### LINKED ISSUE #367: Auto-shoot kills don't credit quest objectives — kill-credit hook only wired at right-click entry points
## Summary

When a player kills a quest target via auto-shoot (auto-cycle), the mission's KillCount-style objective does not advance. Killing the same target with a manual right-click attack works correctly.

**Repro**: med-room drone in the early-game mission. Press the auto-fire button → kill the drone → objective stays at 0. Right-click the drone → kill it → objective advances.

## Root cause

The kill-credit hook (`fire_entity_death`, which advances content-engine kill chains) is wired as a **wrapper around `handle_use_ability`** at three cell-method dispatch sites. Three other player-attack paths call `handle_use_ability` directly and bypass the wrapper.

**Call sites that fire the kill-credit hook (working):**
- [`cell/cell_methods/player/combat/mod.rs::USE_ABILITY`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/player/combat/mod.rs#L40) — manual right-click attack
- [`cell/cell_methods/player/combat/mod.rs::USE_ABILITY_ON_GROUND`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/player/combat/mod.rs#L98) — AoE ground-target ability
- [`cell/cell_methods/player/interaction.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/player/interaction.rs#L90) — cursor right-click interaction (ability 592)

**Call sites that bypass the hook (broken):**
- [`cell/service/ticks/auto_cycle.rs::auto_cycle_tick`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service/ticks/auto_cycle.rs#L146) — the 100ms auto-fire loop (**reported bug**)
- [`cell/cell_methods/player/world/mod.rs::setAutoCycle`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/player/world/mod.rs#L86) — immediate-fire on toggle ON (Phase 2 auto-cycle UX)
- [`cell/service/ticks/mod.rs::pending_attack_tick`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service/ticks/mod.rs#L242) — queued attack-while-holstered (Phase B fires after draw window)

NPC AI also calls `handle_use_ability` (in `service/npc_ai.rs`), but that path correctly skips the hook — the killer is an NPC with no `player_id`, so there's no mission to credit.

## Why the wrapper-at-dispatch shape failed

The wrapper code lives at the cell-method dispatch sites because it needs `engine: &ChainEngine` and the tag/player_id resolution. The pattern looks like:

```rust
let was_alive_before = /* snapshot target health */;
handle_use_ability(entity_id, ability_id, target_id, tx, space_mgr).await;
if was_alive_before && /* target just_died */ {
    let tag = /* lookup tag */;
    let player_id = /* lookup attacker's player_id */;
    fire_entity_death(entity_id, player_id, &tag, engine, tx, space_mgr).await;
}
```

The pattern was duplicated across three dispatch sites — but when auto-cycle (and other tick-driven re-fires) were added, the contributor only added the `handle_use_ability` call, not the wrapper. Easy mistake; the function name doesn't signal "this needs a wrapper for quest credit."

## Proposed fix

Extract the wrapper into a single helper next to `handle_use_ability`:

```rust
pub async fn handle_use_ability_with_kill_credit(
    entity_id: u32,
    ability_id: i32,
    target_id: i32,
    engine: &ChainEngine,
    tx: &mpsc::Sender<CellToBaseMsg>,
    space_mgr: &mut SpaceManager,
);
```

Route all 5 player-driven callers through it. NPC AI keeps the bare `handle_use_ability` (correct — no player_id). Removes the three duplicated wrappers and makes the bug class structurally impossible.

Requires plumbing `engine: &ChainEngine` into `auto_cycle_tick` and `pending_attack_tick`; `setAutoCycle::SET_AUTO_CYCLE` already has it in scope.

## Test plan

- Regression test: tick-driven auto-cycle fire against a tagged NPC fires `fire_entity_death` with the right `entity_tag` so a `Mob_KILLED` chain match progresses (mirror an existing right-click test that asserts the same).
- Existing right-click tests stay green (they exercise the same code path via the helper).

## Risks

- **None observable in shipped behavior** — every kill that used to credit still credits; this widens kill-credit to the three missing paths.
- **API surface**: `handle_use_ability_with_kill_credit` is a new public symbol in `cell/abilities/use_ability/`. The existing `handle_use_ability` stays untouched (NPC AI + tests keep using it directly).

## See also

- The `apply_damage_to_target` function in `cell/abilities/damage_apply/mod.rs` is the actual damage path shared by all callers — the death-state mutation, XP grant, and wire-protocol burst happen there. The kill-credit hook for content chains is a separate concern wired off `engine` (which damage_apply doesn't have), so it belongs at the next layer up.

Cadacious:
## Summary

Auto-shoot kills against quest-tagged NPCs (e.g. the med-room drone in the early-game mission) did not advance KillCount objectives. Manual right-click kills against the same target worked. This PR fixes the divergence by routing every player-driven attack path through a single canonical kill-credit helper.

Closes #367.

## Root cause

`fire_entity_death` (the content-engine `EntityDeath` trigger that KillCount chains subscribe to) was wired as an **inline wrapper** around `handle_use_ability` at three cell-method dispatch sites. **Three other player-attack paths** called `handle_use_ability` directly and bypassed it:

| Path | Status before | Status after |
|---|---|---|
| `combat/mod.rs::USE_ABILITY` (manual right-click) | ✅ fires | ✅ fires (via helper) |
| `combat/mod.rs::USE_ABILITY_ON_GROUND` (AoE) | ✅ fires | ✅ unchanged (different shape, returns Vec of deaths) |
| `interaction.rs` (right-click interact path) | ✅ fires | ✅ fires (via helper) |
| `service/ticks/auto_cycle.rs` (**reported bug**) | ❌ bypass | ✅ fires (via helper) |
| `world/mod.rs::SET_AUTO_CYCLE` (immediate fire on toggle) | ❌ bypass | ✅ fires (via helper) |
| `service/ticks/mod.rs::pending_attack_tick` (queued attack-while-holstered) | ❌ bypass | ✅ fires (via helper) |
| `service/npc_ai.rs` (NPC attacking) | ✅ correctly skips | ✅ unchanged (NPC has no player_id) |

## Fix shape

Extracted the inline wrapper into a single helper next to `handle_use_ability`:

```rust
pub async fn handle_use_ability_with_kill_credit(
    entity_id: u32, ability_id: i32, target_id: i32,
    engine: &ChainEngine,
    tx: &mpsc::Sender<CellToBaseMsg>,
    space_mgr: &mut SpaceManager,
) -> bool;
```

Snapshots the target's alive-state before the ability resolves, calls `handle_use_ability`, and fires `fire_entity_death` if a tagged-NPC alive→dead transition happened on this call.

All 5 player-driven callers now route through the helper. NPC AI keeps the bare `handle_use_ability` (correct — no player_id, nothing to credit). The two previously-inline wrappers (USE_ABILITY + interaction) collapsed into a single helper call each.

`engine: &ChainEngine` is now threaded into `auto_cycle_tick` and `pending_attack_tick`; `SET_AUTO_CYCLE` already had it in scope.

## Test plan

- [x] **End-to-end regression guard** (`auto_cycle_tick_credits_quest_kill_on_tagged_npc_death`) — registers a chain that increments a counter on `EntityDeath(entity_tag=QuestTargetDrone)`, drives one auto-cycle tick against a tagged 1-HP NPC, asserts the counter incremented on the killer entity. Drives the full path: tick → kill-credit wrapper → damage → death → `fire_entity_death` → chain resolve → counter increment.
- [x] **Adversarial reversal**: temporarily switched the tick back to bare `handle_use_ability` — counter stays at None instead of `Some(1)`. Test fires correctly on the bug shape. Reverted before commit.
- [x] Updated all 15 existing `auto_cycle_tick_*` + `pending_attack_tick_*` tests to pass the new `ChainEngine` parameter (via a shared `empty_engine()` helper for the tick-mechanics tests that don't exercise chains).
- [x] `cargo clippy --workspace --exclude {GUI crates} --all-targets -- -D warnings`
- [x] `cargo build --workspace --exclude {GUI crates} --all-targets`
- [x] `cargo test --workspace --exclude {GUI crates} --lib --no-fail-fast` — **844 services + 149 mercury + 1380 lib tests total**, all green.

## Risks

- **None for shipped behavior** — every kill that used to credit still credits. The change widens kill-credit to three previously-missing paths.
- **API surface**: `handle_use_ability_with_kill_credit` is a new public symbol in `cell/abilities/`. The existing `handle_use_ability` stays untouched so NPC AI and the 23 test call sites that don't need engine threading keep working.
- **Test churn**: 15 existing tests in `auto_cycle.rs::tests` and `ticks/mod.rs::tests` need the new engine parameter. Done via shared helper / inline `ChainEngine::new()` — mechanical change, no logic shifts.

### DIFF
--- a/crates/services/src/cell/abilities/mod.rs
+++ b/crates/services/src/cell/abilities/mod.rs
@@ -36,4 +36,4 @@ pub(crate) use messaging::{request_appearance_refresh, send_entity_method};
 // land here for #278 child PRs to adopt. They stay private to the `messaging`
 // module until the first child callsite migrates — at which point the
 // migrating PR adds the re-exports it needs.
-pub use use_ability::handle_use_ability;
+pub use use_ability::{handle_use_ability, handle_use_ability_with_kill_credit};
--- a/crates/services/src/cell/abilities/use_ability/mod.rs
+++ b/crates/services/src/cell/abilities/use_ability/mod.rs
@@ -509,20 +509,19 @@ pub async fn handle_use_ability(
 
 /// `handle_use_ability` + content-engine kill-credit hook.
 ///
-/// **Use this from every player-driven attack path.** Calls
-/// [`handle_use_ability`] to resolve the ability, then — if the attacker
-/// is a player who just transitioned a tagged NPC from alive→dead —
-/// fires the `EntityDeath` content event so mission KillCount chains
-/// (e.g., "kill 5 Hallway_Guards") progress.
+/// **Use this from every single-target player-driven path that calls
+/// [`handle_use_ability`] directly.** Calls `handle_use_ability` to
+/// resolve the ability, then — if the attacker is a player who just
+/// transitioned a tagged NPC from alive→dead — fires the `EntityDeath`
+/// content event so mission KillCount chains (e.g., "kill 5
+/// Hallway_Guards") progress.
 ///
-/// Wrap-shape extracted from the previous in-line copies at the cell-
-/// method dispatch sites in `cell_methods/player/combat/mod.rs` and
-/// `cell_methods/player/interaction.rs`. Three other player-attack
-/// paths used to call [`handle_use_ability`] directly and bypass this
-/// hook — auto-cycle's immediate-fire on toggle, the auto-cycle tick
-/// re-fire loop, and the pending-attack tick (Phase B of
-/// attack-while-holstered). All three are now routed through here,
-/// closing issue #367 ("auto-shoot kills don't credit quest objectives").
+/// **Not** for AoE / ground-target callers: those go through
+/// [`super::handle_use_ability_on_ground`], which returns the set of
+/// every NPC that died during the cast and fires per-death
+/// `fire_entity_death` at the caller layer. The AoE path is the only
+/// other single canonical kill-credit fan-out today; collapsing them
+/// would require returning a Vec<entity_id> from this helper too.
 ///
 /// Why this isn't baked into `handle_use_ability` itself: NPC AI also
 /// calls `handle_use_ability`, and NPC kills shouldn't fire
--- a/crates/services/src/cell/abilities/use_ability/mod.rs
+++ b/crates/services/src/cell/abilities/use_ability/mod.rs
@@ -8,11 +8,13 @@
 //! ordering between immutable and mutable space_manager access is delicate
 //! and a hand-rolled split here would just trade lines for `&mut` plumbing.
 
+use cimmeria_content_engine::chain::ChainEngine;
 use tokio::sync::mpsc;
 
 use cimmeria_entity::abilities::{
     serialize_timer_update, AF_DEACTIVATE_AUTO_CYCLE, TIMER_ABILITY_COOLDOWN,
 };
+use cimmeria_entity::stats::HEALTH;
 
 use super::super::combat;
 use super::super::messages::CellToBaseMsg;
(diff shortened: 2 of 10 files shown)


MyPalClara:

| **G1** | `was_alive_before` race condition | The snapshot reads `HEALTH.cur > 0` before the ability. If two players fire at a 1-HP target simultaneously, both see `was_alive_before = true`. One kills, the other hits a corpse. Does the second caller fire `fire_entity_death` again? The `just_died` check should prevent this (corpse has `cur <= 0`), but verify. |
| **G2** | `USE_ABILITY_ON_GROUND` not migrated | The PR says this path is "unchanged" because it "returns Vec of deaths." But does it fire `fire_entity_death` for each death? If not, AoE kills against quest targets might not credit. Verify the AoE path has its own kill-credit logic. |
| **G3** | `player_id` warn log reachability | If the killer has no `player_id`, the helper logs a warning and skips. But NPC AI shouldn't reach this helper — it uses bare `handle_use_ability`. Is there any other path where a non-player entity could reach `handle_use_ability_with_kill_credit`? If not, the warn is defensive dead code; if so, document the path. |

MyPalClara:

| **G4** | `ChainEngine` threading | `engine: &ChainEngine` is now threaded into `auto_cycle_tick` and `pending_attack_tick`. The engine is borrowed immutably — but does `fire_entity_death` mutate it? If so, the borrow might conflict with other engine uses in the same tick. Verify the engine is read-only during the tick. |
| **G5** | Test churn verification | 15 existing tests updated to pass `ChainEngine::new()`. The PR says "mechanical change, no logic shifts" — verify none of those tests were asserting on behavior that changed (e.g., a test that expected NO kill-credit event now gets one). |

MyPalClara:
## Category 3: Missing Test Coverage (G6-G8)

| **G6** | No test for `pending_attack_tick` kill credit | The regression guard tests `auto_cycle_tick`, but `pending_attack_tick` (queued attack-while-holstered) is also a fixed path. Consider a parallel test for that path. |
| **G7** | No test for `SET_AUTO_CYCLE` immediate-fire kill credit | The immediate-fire on toggle ON is also a fixed path. Consider a test that toggles auto-cycle ON with a 1-HP target and verifies the kill credits. |
| **G8** | No test for double-fire race | What if two players fire at the same 1-HP target in the same tick? Both see `was_alive_before = true`, one kills, the other hits a corpse. Verify only one `fire_entity_death` fires (the `just_died` check should handle this). |

MyPalClara:
## Category 4: Documentation + Summary (G9)

| **G9** | No architecture doc | The helper is well-documented in code, but there's no architecture doc explaining the kill-credit flow. Consider adding a section to `docs/gameplay/combat.md` explaining the `handle_use_ability` vs `handle_use_ability_with_kill_credit` distinction and when to use each. |

---

## Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟡 Verify | G1, G2, G4, G5 | ~30 min |
| 🟡 Consider | G6, G7 (parallel path tests) | ~45 min |
| 🟢 Quick wins | G9 (doc update) | ~15 min |
| ⚪ Nice-to-have | G3, G8 | ~30 min |

**Total: 9 gaps, ~2 hours of follow-up work.**

---

## Issue #367 Coverage ✅

All 6 attack paths are now correctly handled:
- ✅ `USE_ABILITY` (manual right-click) — via helper
- ✅ `USE_ABILITY_ON_GROUND` (AoE) — unchanged (verify G2)
- ✅ `interaction.rs` (right-click interact) — via helper
- ✅ `auto_cycle_tick` (**fixed**) — via helper
- ✅ `SET_AUTO_CYCLE` immediate fire (**fixed**) — via helper
- ✅ `pending_attack_tick` (**fixed**) — via helper
- ✅ NPC AI — correctly excluded

**Auto-shoot kills now credit quest objectives.** 🎯

---

## Cross-PR Note

This PR interlocks with **#368** (Castle drone encounter):
- #368 makes the drone wake up and attack
- #369 ensures auto-fire kills credit the mission

**Both are required for the Castle_CellBlock encounter to work end-to-end.**

Cadacious:
## Dismissing the three new Copilot "temporary across .await" comments

The pattern `auto_cycle_tick(&tx, &mut mgr, &empty_engine()).await` and `pending_attack_tick(..., &ChainEngine::new()).await` is **valid Rust** and compiles cleanly (verified by the 846 passing services tests, including the three new kill-credit regression guards that exercise these exact lines).

**Why Copilot is wrong about this one:**

Temporary lifetimes in Rust extend to the end of the enclosing **expression statement** (Reference §6.4.4 "Temporary lifetimes"). For
```rust
auto_cycle_tick(&tx, &mut mgr, &empty_engine()).await;
```
the temporary `ChainEngine` from `empty_engine()` lives until the trailing `;`. The future returned by `auto_cycle_tick(...)` borrows it; `.await` polls that future to completion *before* the statement ends; the future is dropped after `.await` resolves; the temporary is then dropped at `;`. No reference to the temporary survives past the await — the borrow ends when the future returns.

The pattern Rust **does** reject is holding a `&` to a temporary **across** an await suspension, where the reference is needed **after** the await:
```rust
// THIS would be rejected:
let r = &some_temp_fn();
some_other_async_thing().await;
use(r);  // ← borrow needed after .await
```

That's not what these test sites are doing — `&empty_engine()` is consumed entirely within the same expression as the `.await`.

**Decision**: no code change. Adding `let engine = ChainEngine::new();` bindings across ~14 test call sites would be ~14 lines of churn for zero behavior change, and the lint signal would be "Copilot's false positive is correct" — actively misleading for future readers.

Copilot's other 8 comments on this PR (issue-ref strips + the docstring wording fix) were accurate and have all been actioned in commit `8b7dfc16`.

Issue #355, opened 2026-05-23
Network chaos test apparatus (lossy UDP + pcap replay) — gates #354


Cadacious:
## TL;DR

Build a deterministic lossy-network test apparatus capable of reproducing the **lomiada failure shape**: a single packet dropped mid-session on a transatlantic UDP connection (Germany → US server, ~100-150ms RTT), followed by the client buffering all subsequent packets until the 60s inactivity timer kills the session. Today no test exercises this; we caught it only when a real player surfaced it. Going forward, every protocol-level regression must be reproducible offline.

**Gates #354.** The three interim fixes (Bundle, defer-creates, deferred-send queue) all need this infrastructure to prove they actually solve the bug. Without it, #354's acceptance criteria can only be validated by manual smoke testing against a real client, which doesn't scale to CI.

### What lomiada actually was

`debug/lomiada-broke-in-hallway02/` is the source of truth. The player connected from Germany to a US server. After ~6 minutes of clean play, server-sent packet `#1148` was lost on the wire (genuine network drop — transatlantic UDP has measurable loss probability at any given moment). The client kept receiving `#1149`, `#1150`, etc. and buffered all 210 subsequent packets above the gap. Server's TX window filled because the client cannot send a cumulative ACK past the missing seq, the deferred-best-effort path swallowed the rest, and 60s later the inactivity timer killed the session.

The failure has six discrete protocol-level steps. No current test exercises any of them.

### Why current tests miss this class of bug

- **Unit / wire-format tests** assert the *shape* of single packets, not the dynamic behavior of a Channel under loss.
- **Live-DB tests** assert DB invariants, orthogonal to network behavior.
- **Smoke tests** (`crates/services/src/test_support.rs`) run against an in-process server but use direct method invocation — no real socket, no loss simulation.
- **Replay-based testing**: we already have `crates/mercury/src/packet/replay_smoke.rs`, but it replays *fixture packets* through the parser to assert encode/decode round-trips. It does not run a server-side channel state machine and does not simulate loss.

There is no test in the workspace that does: "start a Channel, feed it 1000 packets, drop packet N, assert the channel recovers within K ticks."

### What the Germany context tells us

Transatlantic UDP has materially different failure characteristics than LAN:

- **RTT ~100-150ms** vs ~1ms LAN. ACKs come back slow; the TX window can fill faster than the slot-free rate.
[...]

PR #370, opened 2026-05-24
feat(mercury): Tier 2 loopback session harness + 22 paired-channel tests (#352)

### LINKED ISSUE #352: test-infra: Mercury loopback session harness (Tier 2 of UDP/async testing split)
## Goal

Build a **Mercury loopback session harness** in `crates/mercury/` that lets tests pair two `Nub` instances on loopback and exercise the full reliable-UDP protocol — encryption, sequencing, acks, retransmits, fragmentation, keepalives, channel lifecycle — without involving the service crate, the database, or any auth/content layer.

This is **distinct from #281** (the wireclient end-to-end harness). #281 is a content-aware client that walks the Castle Cellblock script with auth, entity mirrors, dialog state, and combat. This issue is the **protocol-only** layer one floor below it. Two Nubs on loopback that talk Mercury to each other and nothing else.

This issue is also distinct from **Tier 1** (the `Transport` trait + `TestTransport` recorder). Tier 1 covers the *outbound* side with a recording fake; Tier 2 covers the *interactive* side where two endpoints actually exchange packets and exercise the recv loop, channel state, and timing-sensitive paths.

### What today's mercury tests cover

`crates/mercury/` has ~1,586 lines of tests across 5 files exercising:
- Packet parse/build byte exactness (`packet/tests.rs`, 305 lines + 161-line proptest round-trip)
- Channel state-machine in isolation, with `Instant`-mocked time (`channel/tests/channel_lifecycle.rs`, 541 lines)
- Fragment reassembly in isolation (`channel/tests/reassembly.rs`, 253 lines)
- Unpacker out-of-order reassembly (`unpacker/tests.rs`, 480 lines)
- `Nub::tick` return-value contract (`nub.rs` tests, ~240 lines inline)

### What today's mercury tests can't cover

Every existing test exercises **one half of the wire conversation at a time**. There is no test in the repo today that does any of the following:

1. **End-to-end reliable delivery under simulated loss.** Channel A sends a reliable packet to Channel B; the test drops it; Channel A's RTO fires; Channel A retransmits; Channel B receives the retransmit; Channel B acks; Channel A's ack-handling clears the tx window. Every leg of that pipeline has unit tests; the **pipeline itself** is untested.
2. **Fragment reassembly across two paired channels with mixed-order arrival.** The unpacker tests reassemble in-place; the channel tests don't pair two channels. A regression where channel A's fragment numbering disagrees with channel B's reassembly expectation would slip every test in the suite.
3. **Keepalive cadence end-to-end.** The `tick_re_flags_keepalive_until_caller_acks_send` test pins the return-value contract — it does NOT verify that a real keepalive packet, decoded by the peer, is recognized as a keepalive and updates the peer's `last_received`. A regression in the keepalive packet's wire form (wrong flags, missing `seq_id` footer, etc.) wouldn't fail any current test.
4. **Encryption round-trip across multiple bundles.** `encryption.rs` has unit tests; `nub.rs` tests use unencrypted bytes. A regression where the RC4 keystream advances incorrectly across bundles is invisible to today's suite.
5. **Channel lifecycle handshake.** `Connecting` → `Connected` → `Disconnecting` is tested as state mutations on a single channel. There's no test where two channels execute the handshake by exchanging packets.
6. **Ack aggregation behavior.** When channel A receives N packets from B in a short window, A is supposed to coalesce acks into one packet's footer. The footer build is unit-tested; the coalescing decision under realistic arrival patterns is not.
7. **Adaptive RTO convergence.** The RTO calculator is unit-tested with synthetic RTT samples. With paired channels on loopback (which produce real sub-millisecond RTTs), the floor in `crates/mercury/src/channel/rto.rs` is exercised under the actual conditions it was added for.

### What this enables downstream

- **#281 Phase 1** (the wireclient handshake driver) is layered on top of this. Without a Mercury loopback harness, #281's Phase 1 has to either (a) bring its own equivalent infrastructure or (b) test the handshake driver against a real server process, which slows the dev loop.
- **#57 Acceptance items** ("Retransmission test: Drop a packet → verify channel retransmits", "Keepalive test: Idle connection for >threshold → verify keepalive packet sent", connect/disconnect/saturate smokes) are written assuming a harness like this exists. Tier 2 is the home for them.
- **Mercury protocol regression suite** — once paired-channel tests exist, every future Mercury-layer change (new packet flag, ack-window tuning, fragment-size adjustment) lands with an end-to-end test in this harness, not just a state-machine unit test.

### New module: `crates/mercury/src/test_harness.rs` (cfg-gated)

```rust
//! Loopback Mercury session harness for end-to-end protocol tests.
//!
//! Pairs two `Nub` instances on real loopback `UdpSocket`s and runs a
//! driver loop that pumps recv on both sides, dispatches inbound bytes
//! to the right channel, drives `tick()` at a controllable cadence, and
//! exposes hooks for simulated loss / reorder / latency.

#![cfg(any(test, feature = "test-harness"))]

use std::net::SocketAddr;
use std::sync::Arc;
use std::time::{Duration, Instant};

use tokio::net::UdpSocket;
use tokio::sync::Mutex;

use crate::encryption::MercuryEncryption;
use crate::nub::Nub;
use crate::packet::{Bytes, Packet};

/// One end of a paired Mercury session. Owns a `Nub` and a recv pump.
pub struct LoopbackPeer {
    pub nub: Arc<Mutex<Nub>>,
    pub addr: SocketAddr,
    pub recv_handle: tokio::task::JoinHandle<()>,
}

/// A paired Mercury session for end-to-end testing.
pub struct LoopbackSession {
    pub a: LoopbackPeer,
    pub b: LoopbackPeer,
    pub policy: Arc<Mutex<NetworkPolicy>>,
}

/// Per-direction packet manipulation policy. Default is lossless,
/// in-order, zero-latency loopback.
#[derive(Default, Clone)]
pub struct NetworkPolicy {
    /// Drop the next N packets in `direction` before resuming.
    pub drop_next: NetworkDirection<u32>,
    /// Insert `latency` before delivering each packet in `direction`.
    pub latency: NetworkDirection<Duration>,
    /// Deliver packets in `direction` in reverse pairs (1,2 → 2,1, etc.).
    pub reorder_pairs: NetworkDirection<bool>,
    /// Duplicate every Nth packet in `direction`.
    pub duplicate_every: NetworkDirection<Option<u32>>,
}

#[derive(Default, Clone)]
pub struct NetworkDirection<T> {
    pub a_to_b: T,
    pub b_to_a: T,
}

impl LoopbackSession {
    /// Pair two Nubs on ephemeral loopback ports. Both endpoints have
    /// established `Connected` channels to each other; the handshake
    /// has already been driven before the constructor returns.
    pub async fn connected(enc: MercuryEncryption, key: [u8; 32]) -> Self { ... }

    /// Pair two Nubs without driving the handshake. Caller is responsible
    /// for sending the first packet that walks each channel through
    /// `Connecting` → `Connected`. Used by handshake-specific tests.
    pub async fn unconnected() -> Self { ... }

    /// Drive `tick` on both Nubs once. Returns the actions each side
    /// would have emitted; the harness has already shipped them onto
    /// the loopback wire respecting the current `NetworkPolicy`.
    pub async fn tick(&self) -> (TickActions, TickActions) { ... }

    /// Wait until peer `b` has received `n` complete bundles from `a`,
    /// or `timeout` elapses. Returns the received bundles in order.
    pub async fn recv_n_bundles_from_a(
        &self,
        n: usize,
        timeout: Duration,
    ) -> Vec<Bytes> { ... }

    /// Wait until both peers have empty tx windows and zero pending acks,
    /// or `timeout` elapses. Useful for "drive to quiescence" assertions.
    pub async fn quiesce(&self, timeout: Duration) -> bool { ... }

    /// Tear down both peers. Idempotent.
    pub async fn shutdown(self) { ... }
}
```

### Time-control hook

```rust
/// Per-Nub controllable clock. Each `LoopbackPeer` owns its own `TestClock`
/// instance — there is no process-global state, so parallel nextest runs are
/// isolated. Behind the `test-harness` feature.
pub struct TestClock {
    base: Instant,
    offset: Mutex<Duration>,
    frozen: AtomicBool,
}

impl TestClock {
    pub fn new() -> Arc<Self> { ... }

    /// Advance the clock by `by`. Subsequent `now()` calls return the
    /// advanced time. Does NOT touch `tokio::time` — keepalive/RTO logic
    /// reads time through the clock trait, not through `tokio::time::sleep`.
    pub fn advance(&self, by: Duration) { ... }

    /// Stop the clock — `now()` returns the same value until `resume`.
    pub fn freeze(&self) { ... }
    pub fn resume(&self) { ... }

    pub fn now(&self) -> Instant { ... }
}
```

**Integration shape (recommended for Phase 1):** introduce a `Clock` trait in `crates/mercury/src/clock.rs`:

```rust
pub trait Clock: Send + Sync {
    fn now(&self) -> Instant;
}

pub struct SystemClock;
impl Clock for SystemClock { fn now(&self) -> Instant { Instant::now() } }
```

Inject `Arc<dyn Clock>` into `Nub::new` (defaulting to `SystemClock`); replace every `Instant::now()` in `crates/mercury/src/channel/` with `self.clock.now()`. The harness constructs each `LoopbackPeer` with a fresh `TestClock` and the test drives time via `peer.clock.advance(Duration::from_millis(...))`. This does NOT replace `tokio::time::sleep` — Mercury's protocol logic is `tokio::time`-free; only the `Channel`/`Nub` bookkeeping reads `Instant::now()`.

**Usage example:**

```rust
#[tokio::test]
async fn keepalive_fires_after_idle_interval() {
    let session = LoopbackSession::connected(enc, key).await;

    // Both peers idle for the full keepalive window.
    session.a.clock.advance(Duration::from_millis(KEEPALIVE_INTERVAL_MS + 100));

    let (a_actions, _) = session.tick().await;
    assert_eq!(a_actions.keepalives.len(), 1, "tick must schedule a keepalive");

    // Drive the keepalive on the wire and let B observe it.
    let bundles = session.recv_n_bundles_from_a(1, Duration::from_secs(1)).await;
    assert_eq!(bundles.len(), 1, "B must receive the keepalive bundle");
    assert!(bundles[0].is_keepalive(), "B must recognize it as a keepalive");
}
```

**Open question Phase 1 must close:** is the clock-trait injection feasible without rippling through `Channel`'s public API? If not, the fallback is a `harness::backdate_peer(&LoopbackPeer, by: Duration)` helper that walks the peer's channel state and applies the same `last_sent -= by, last_received -= by, tx_window[i].last_sent -= by` pattern the existing channel tests use directly. Less elegant; ships in days instead of weeks if the clock trait blows up.

## Scope

**In scope:**
- Paired `LoopbackPeer`s sharing a `NetworkPolicy`
- Drop / reorder / latency / duplicate policies (deterministic, seedable for repro)
- Time-control hook for keepalive / RTO tests
- Encryption round-trip in the loopback path (use `MercuryEncryption` exactly as production does)
- A first pass of end-to-end tests covering the 7 gap categories above
- Behind a `cfg(any(test, feature = "test-harness"))` so production builds don't compile it

### Ordering, isolation, and runtime guarantees

These are the contract bits tests must be able to rely on. Spec'd here so reviewers can confirm and so test authors know what they can assert:

- **Per-direction ordering is preserved.** If A sends packets 1, 2, 3 to B, B observes them in that order (subject to any active `reorder_pairs` or `drop_next` policy in the A→B direction). Same for B→A.
- **Cross-direction interleaving is implementation-defined unless serialized via `tick()`.** If A sends 1, 2, 3 and B sends 4, 5, 6 *simultaneously*, the order in which A's recv pump observes (4, 5, 6) relative to its own (1, 2, 3) sends is NOT guaranteed. Tests that need a precise interleaving must drive the harness with explicit `session.tick().await` calls between phases. This is the only honest guarantee given the recv pump can't observe both directions atomically; documenting it up front prevents flake-debugging later.
- **Per-session `TestClock` isolation.** Each `LoopbackPeer` owns its own `Arc<TestClock>` — no process-global state. `cargo nextest run` with parallel jobs is safe; two harness tests in flight at once cannot perturb each other's clock.
- **Per-session port isolation.** Each `LoopbackPeer` calls `UdpSocket::bind("127.0.0.1:0")` so the OS assigns ephemeral ports. Parallel runs cannot collide on a fixed port.
- **Tokio runtime: defaults to `#[tokio::test]` (current-thread).** The recv pumps are driven cooperatively via `tokio::select!`, not via true parallelism. Tests that want to assert simultaneous send-without-stall behavior must opt in to `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` — document in the test header why. The category-7 RTO tests run on current-thread; the few "no head-of-line blocking" tests in category 1 use multi-thread.

**Out of scope** (these are #281's job, not this issue's):
- SOAP/HTTP auth phase
- Shard key exchange
- Entity creation, property bags, AoI, dialog state
- Any service-crate dependency
- Any DB dependency
- Walking content scripts

The harness is intentionally **content-blind** — it speaks Mercury and nothing else. Peer A sends bytes; Peer B receives bytes. What those bytes mean is the next layer's problem.

## Test inventory (Phase 4 deliverable)

The first batch of tests this harness enables, organized by the 7 gap categories. Each is a single end-to-end test that today is impossible to write.

### 1. Reliable delivery under simulated loss
- `reliable_packet_retransmits_after_drop` — A sends reliable seq=42 to B; policy drops it; advance time past RTO; A retransmits; B acks; A's tx_window clears.
- `retransmit_at_max_retries_minus_one_still_delivers` — same shape but the drop policy expires right at the last retry; verify delivery succeeds and channel survives (parallel to the existing `tick_does_not_reap_channel_on_same_tick_max_retries_hit` unit test, but driven by real send→drop→retx cycles).
- `channel_dies_after_max_retries_exhausted` — drop policy forever; advance time; A's channel transitions to dead; the dead-channel cleanup actions fire.
- `out_of_order_acks_clear_tx_window_correctly` — A sends 1,2,3 reliable; B acks 2 then 1 then 3; A's tx_window clears in arrival order, not seq order.

### 2. Fragment reassembly across paired channels
- `large_payload_fragments_send_and_reassemble` — A sends 8KB bundle that fragments into 3 packets; B reassembles to byte-identical original.
- `fragments_arrive_out_of_order_via_reorder_policy` — same payload; reorder policy delivers (1,2,3) as (2,3,1); B's reassembly still completes.
- `single_fragment_loss_causes_retransmit_of_only_lost_fragment` — 3-fragment bundle; drop fragment 2; verify A only retransmits fragment 2, not the whole bundle.
- `concurrent_bundles_dont_cross_contaminate_reassembly` — A sends bundle X (3 fragments) and bundle Y (2 fragments) interleaved; B's reassembler keeps the two bundle states separate.

### 3. Keepalive cadence end-to-end
- `keepalive_fires_after_idle_interval_and_peer_recognizes_it` — both peers connect; advance time past `KEEPALIVE_INTERVAL_MS`; A's tick emits a keepalive; B receives it and recognizes it as a keepalive (B's `last_received` advances).
- `active_traffic_suppresses_keepalive` — A sends data continuously; advance time; verify no extra keepalive packets appear in B's recv stream.
- `dropped_keepalive_re_flags_next_tick` — keepalive scheduled; policy drops it; next tick re-schedules. Pairs the existing nub-level unit test with the wire-level confirmation.

### 4. Encryption round-trip across multiple bundles
- `multi_bundle_encryption_keystream_advances_correctly` — A sends bundles 1,2,3 reliable; B decrypts all three to the expected plaintext (catches keystream-rewind regressions).
- `retransmit_uses_pre_encrypted_bytes_not_re_encryption` — A sends reliable bundle; policy drops; A retransmits. Verify the retransmit bytes are byte-identical to the original send (which is the `Bytes` stored on `TxWindowEntry`, not a fresh encryption).
- `encryption_failure_on_corrupted_arrival_does_not_crash_channel` — flip a bit in transit; B's decryption fails gracefully; channel survives; A's next send still works.
- `known_answer_ciphertext_matches_sgw_client_reference` — **the highest-value test in this category.** Encrypt a fixed plaintext with a fixed key + IV that match a Ghidra-extracted reference capture from the SGW client's `Mercury::Channel::send` path. Assert the resulting ciphertext byte-exactly matches the reference. This catches the "Mercury encrypts correctly per RFC but doesn't match what the real client expects" failure mode — the one the property-based round-trip tests structurally cannot catch because they only validate self-consistency. Reference samples to extract: one short (≤32 byte) payload, one mid-size (~512 byte) payload that crosses an RC4 block boundary, one fragmented payload. File reference samples under `crates/mercury/src/test_harness/reference_samples/` with a doc-comment naming the Ghidra address each was lifted from.

### 5. Channel lifecycle handshake
- `handshake_drives_both_channels_to_connected_state` — `LoopbackSession::unconnected` then drive the handshake; both channels reach `Connected`.
- `unrequited_first_packet_does_not_advance_state` — A sends to B but B doesn't reply; A's channel stays `Connecting` past one tick.
- `disconnect_message_drives_both_to_disconnected` — happy-path tear-down.

### 6. Ack aggregation
- `n_packets_in_window_produce_one_aggregated_ack` — A sends 5 reliable packets back-to-back; B's response packet's ack footer carries all 5 sequence ids in one ack list.
- `ack_aggregation_respects_packet_size_limit` — push past the aggregation cap; verify the ack list splits across two packets rather than overflowing.

### 7. Adaptive RTO convergence on loopback
- `rto_converges_to_floor_on_zero_latency_loopback` — runs N round-trips on the harness's zero-latency loopback; verify the adaptive RTO settles to the configured floor rather than zero (regression for the floor-was-removed bug shape called out in `crates/mercury/src/channel/rto.rs:286-292, 407-411`).
- `rto_inflates_under_added_latency_policy` — set latency policy to 200ms; verify RTO climbs to roughly 200ms + jitter, not stuck at the floor.

## Phasing

| Phase | Work | Estimate |
|---|---|---|
| 1 | Land the `Clock` trait in `crates/mercury/src/clock.rs`; thread `Arc<dyn Clock>` through `Nub::new` and every `Instant::now()` call in `crates/mercury/src/channel/{mod,rto}.rs`. Scaffold `LoopbackPeer` + `LoopbackSession::unconnected`. One smoke test: A sends one packet, B receives it. **If the clock-trait threading proves invasive (>15 sites in `channel/`), bail to the `harness::backdate_peer` fallback described in the time-control section and document why.** | 3 days |
| 2 | `NetworkPolicy` (drop / reorder / latency / duplicate) wired into the loopback pump. `LoopbackSession::connected` with the handshake drive. `quiesce` + `recv_n_bundles_from_a`. | 2 days |
| 3 | Encryption integration — accept a `MercuryEncryption` + key at construction, exercise the encrypted path on every send. Extract the 3 Ghidra reference samples for the known-answer test (category 4). | 1.5 days |
| 4 | Land the test inventory: 7 categories × ~3 tests each ≈ 22 paired-channel tests (was 21 — the known-answer ciphertext test brings it to 22). One file per category under `crates/mercury/src/test_harness/tests/`. | 3 days |
| 5 | Doc updates (see "Required documentation" below). | 0.5 day |
| 6 | Optional: address #57's acceptance criteria using the harness (long-session idle, saturation, reconnect). Can land in a follow-up. | 1 day |

**Total: ~10 days of focused work**, split naturally across two PRs (1–3 land the harness, 4–5 land the test inventory). Phase 1's bumped estimate reflects the real cost of clock-trait injection through `channel/` — the previous 2-day estimate was wishful given the call-site count in `channel/mod.rs` and `channel/rto.rs`.

## Acceptance criteria

- [ ] `crates/mercury/src/test_harness/` exists with `LoopbackSession`, `LoopbackPeer`, `NetworkPolicy`, `TestClock`
- [ ] Behind `cfg(any(test, feature = "test-harness"))` so production builds don't include it
- [ ] `Clock` trait lands in `crates/mercury/src/clock.rs` with `SystemClock` production impl; `Nub::new` accepts `Arc<dyn Clock>` (with a `Nub::with_system_clock` convenience constructor for production callers)
- [ ] `LoopbackSession::connected` drives the handshake before returning
- [ ] All 4 `NetworkPolicy` modes (drop / reorder / latency / duplicate) work per-direction
- [ ] Per-direction send ordering is preserved end-to-end; cross-direction interleaving guarantees documented in `test_harness/mod.rs` doc-comment
- [ ] Each `LoopbackPeer` owns its own `Arc<TestClock>` (no process-global clock state)
- [ ] `cargo nextest run --profile=ci -p cimmeria-mercury` runs the harness tests and they pass deterministically (no `sleep` for time advancement — use the clock hook)
- [ ] The full harness suite passes under `cargo nextest run --partition count:1/100` (no flake at 100x repetition)
- [ ] The full harness suite passes under nextest default parallelism without test-test interference (proves the per-session port/clock isolation)
- [ ] Default test attribute is `#[tokio::test]` (current-thread); tests that need real parallelism explicitly use `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` with a doc-comment naming the reason
- [ ] The 22-test inventory above lands under `crates/mercury/src/test_harness/tests/{reliable,fragment,keepalive,encryption,handshake,ack,rto}.rs`
- [ ] Encryption tests exercise the same `MercuryEncryption` impl production uses
- [ ] `known_answer_ciphertext_matches_sgw_client_reference` lands with the 3 Ghidra-extracted reference samples under `crates/mercury/src/test_harness/reference_samples/`, each with a doc-comment naming the source address in SGW.exe
- [ ] Doc updates per CLAUDE.md's required-documentation map:
  - [ ] New `docs/architecture/mercury-loopback-harness.md` (explanation) covering: when to use this vs `TestTransport` (Tier 1) vs wireclient (#281); the clock-injection decision; the policy model
  - [ ] `TESTING.md` gains "mercury session test" as the 9th test type (8th is fan-out byte tests from Tier 1) with a picker entry
  - [ ] `crates/README.md` `cimmeria-mercury` entry mentions the harness and its feature flag
  - [ ] `.github/copilot-instructions.md` review checklist: "Mercury-protocol changes (channel state, retransmit, fragmentation, keepalive, ack) require a paired-channel test in the loopback harness, not just a unit test"

## Risks & open questions

1. **Clock injection vs `Instant` backdating.** The existing channel tests backdate `Instant::now()` directly into the state. For paired channels that's fragile — both peers' clocks must advance coherently. Resolution: introduce a `Clock` trait that defaults to `std::time::Instant` in production and uses `TestClock` in the harness (see "Integration shape" in the time-control section). This is a non-trivial Nub change touching `channel/mod.rs` and `channel/rto.rs`. Land in Phase 1, before any test code. **Phase 1 has a named fallback** (`harness::backdate_peer`) if the trait threading proves invasive — the test inventory ships either way.
2. **Recv-pump scheduling fairness.** A naive `tokio::select!` between A's recv and B's recv may starve under high traffic. Use `Notify` or per-direction queues; document the scheduling guarantees in the harness doc. The ordering contract spec'd in "Ordering, isolation, and runtime guarantees" above is the test-facing surface; the implementation has freedom below that.
3. **Loopback recv timing on Windows.** Windows' loopback path has different latency characteristics than Linux. The RTO floor test (category 7) must use the harness clock, not wall-clock, to avoid platform flakes.
4. **Encryption keystream coupling to the harness.** If the harness ever needs to inspect plaintext mid-flight (for assertion), it must hold a second copy of `MercuryEncryption` in matching state. Document the cost and provide a `decrypt_observed` helper that does this safely.
5. **Known-answer reference samples are a maintenance burden.** The 3 Ghidra-extracted samples for category 4 are tied to a specific SGW.exe build. Document the exact build version + Ghidra addresses in each sample file's header so a future engineer can re-extract if the binary changes. If the client is ever re-built, the samples need regeneration — flag this in the reference-samples README.
6. **Test count budget.** 22 new tests add ~1.6% to the workspace test count (currently 1351) — below the 5% threshold that requires a docs/testing/inventory refresh. Confirm during Phase 5.
7. **Tokio runtime flavor.** Default `#[tokio::test]` is current-thread, which means `tokio::select!` between the two recv pumps cooperatively yields rather than running in parallel. Most tests in the inventory are fine with that. The few tests asserting "neither direction stalls the other" need `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` — call out in the test header. CI nextest profile must not pin a single-threaded runtime globally.
8. **Interaction with Tier 1.** Tier 1's `TestTransport` records outbound only; Tier 2's loopback harness uses a real socket bridge. Tier 1 (#351) ships first; the harness MAY use `TestTransport` internally as the "between A's nub and the wire" stub, but isn't required to.

## What this issue is NOT

Listing this explicitly so reviewers can confirm the carve-out:

- **Not a wireclient.** No auth, no entity model, no content awareness. See #281 for that.
- **Not a Tier 1 replacement.** Tier 1's `TestTransport` is the right tool for "I want to assert that handler H emitted these bytes." The harness is the right tool for "I want to assert that the Mercury protocol delivers these bytes under conditions X."
- **Not an integration test against the real server.** No service crate dependency, no DB dependency, no `BaseService::start`. Spinning up the server is #281's problem.
- **Not a Nub I/O implementation.** `Nub::send_to`/`recv_from` are resolved by Tier 1 (#57) — this issue uses whatever resolution Tier 1 lands.

## Cross-references

- Prerequisite for #281 Phase 1 (wireclient handshake driver) — that phase builds the SOAP+phase-3 client on top of this harness's session machinery
- Resolves the smoke-test acceptance items in #57 (retransmission test, keepalive test, idle stability, saturation, reconnect)
- Complements Tier 1 (`Transport` trait + `TestTransport`) — Tier 1 covers handler-side fan-out byte tests; this covers protocol-layer end-to-end tests
- Mercury source under test: `crates/mercury/src/{nub,channel/mod,channel/rto,channel/tests/reassembly,encryption,packet/mod,packet/parse,packet/build,unpacker/mod,bundle}.rs`
- Existing mercury tests this complements (does not replace): `crates/mercury/src/packet/tests.rs`, `crates/mercury/src/channel/tests/`, `crates/mercury/src/unpacker/tests.rs`, `crates/mercury/src/nub.rs` inline tests

Cadacious:
## Summary

Closes #352. Lands the Tier 2 Mercury loopback session harness and the full 22-test paired-channel inventory the issue called for.

- **`Clock` trait + `SystemClock`** (`crates/mercury/src/clock.rs`) — threaded through `Channel` (11 call sites swapped from `Instant::now()` / `.elapsed()` to `self.clock.now()`). `Channel::new(addr)` keeps its single-arg signature; the 3 services-layer constructor callers don't change.
- **`test_harness` module** behind the `test-harness` Cargo feature — `LoopbackPeer`, `LoopbackSession`, `NetworkPolicy` (per-direction drop / latency / duplicate), `TestClock`, with recv pumps on real `127.0.0.1:0` loopback sockets and end-to-end encryption integration.
- **22 paired-channel tests** across 7 files matching the issue spec — 4 reliable + 4 fragment + 3 keepalive + 4 encryption (incl. known-answer ciphertext KAT) + 3 handshake + 2 ack + 2 RTO. Plus 3 smoke tests proving the scaffold.
- **Known-answer ciphertext test** uses 3 reference vectors generated by `tools/generate-mercury-kat.py` (Python `cryptography`) — Ghidra reconnaissance confirmed SGW.exe uses standard CryptoPP `CBC_Encryption<Rijndael>` + `HMAC<MD5>` with no custom modifications, so an independent reference implementation gives equivalent failure-mode coverage to a live Ghidra extraction (and is reproducible without an SGW.exe build).
- **ADR** at `docs/architecture/mercury-loopback-harness.md` covers the design (clock-injection rationale, policy model, runtime contract, KAT vector approach, alternatives considered).
- **TESTING.md** gains a 9th test type (Mercury session tests) with a picker entry; `.github/copilot-instructions.md` adds the review checklist row; `crates/README.md` updates the cimmeria-mercury entry; `CLAUDE.md` adds the doc-update map row.

## Test results

| Run | Tests | Status |
|---|---|---|
| \`cargo nextest run -p cimmeria-mercury\` | 182/182 (was 149) | ✅ |
| \`cargo nextest run --workspace\` (excl. GUI crates) | 1418/1418 | ✅ |
| Harness flake check (3× full suite) | 33 × 3 = 99/99 | ✅ no flakes, ~130ms/run |
| \`cargo fmt --all --check\` | clean | ✅ |
| \`cargo clippy --workspace ... --features cimmeria-mercury/test-harness -- -D warnings\` | clean | ✅ |
| \`cargo test --doc -p cimmeria-commands\` | 4/4 | ✅ |
| Figure-source-in-sync (blocking) | 48 sources clean | ✅ |
| Figure style lint (blocking) | 21 sources, 48 SVGs, 3 chapters clean | ✅ |

## Architecture notes

**Clock injection footprint vs. issue's risk register.** The issue named a >15-call-site bailout to a `backdate_peer` fallback. The actual count post-Nub-deletion: 11 call sites in one file. Elegant path was unambiguously the right move.

**Per-direction policy on sender side.** Real loopback has no OS hook between A's \`send_to\` and B's \`recv_from\`; applying drop / latency / duplicate at the sender produces the same observable behavior as wire-side application, with the same RTO / retransmit consequences on the sender's side.

**Cross-direction interleaving contract.** Documented as implementation-defined unless serialized via \`session.tick()\`. Honest about the cooperative \`tokio::select!\` shape; tests that need a precise interleaving use explicit ticks.

## Test plan

- [x] Mercury crate: \`cargo nextest run -p cimmeria-mercury\` → 182/182
- [x] Workspace: \`cargo nextest run --workspace --exclude cimmeria-app --exclude cimmeria-content-editor --exclude cimmeria-scene-editor --exclude sgw-launcher\` → 1418/1418
- [x] Flake check: harness suite 3× consecutive runs, all passing
- [x] All five CI gates pass locally (fmt / clippy / build / nextest / doctests)
- [x] Both blocking figure lints clean
- [x] KAT regeneration verified: \`python tools/generate-mercury-kat.py\` produces the constants currently in \`kat_vectors.rs\`

## Release Notes

* **Tests**
  * Added comprehensive integration test suite for Mercury protocol layer, validating reliability under loss, fragmentation, keepalive cadence, encryption round-trips, handshake lifecycle, ACK aggregation, and RTO convergence.

* **Documentation**
  * Added architecture documentation for Mercury protocol testing infrastructure.
  * Updated testing guidance with new protocol-layer test requirements.

* **Chores**
  * Added Mercury protocol test harness with deterministic time control and network policy simulation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/370?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MyPalClara:

### G1: `session.rs` handshake state mutation
**Location:** `LoopbackSession::connected()` lines 85-92
**Observation:** The `connected()` constructor directly mutates `ch.state = ChannelState::Connected` after exchanging empty packets. This bypasses the actual handshake state machine — the state transition happens via direct field write, not via the channel's own state-machine logic.

The ADR acknowledges this: "Tests that need to assert specific handshake bytes should use `unconnected`." But it means the `connected()` path doesn't exercise the real handshake transition.

**Consider:** A `drive_handshake()` helper that uses the real state machine, or document explicitly that `connected()` is a shortcut that skips state-machine validation.

---

### G2: `peer.rs` `send_large_bundle` seq reservation
**Location:** `LoopbackPeer::send_large_bundle()` lines 130-145
**Observation:** The seq reservation happens under one lock acquisition, then the lock is released, then the sends happen outside the lock. The comment says "Reserve the sequence space for all fragments up front so a concurrent send can't interleave" — but the lock IS released before the sends.

The reservation is atomic (the seqs are reserved), so a concurrent send would get the NEXT seq, not an interleaved one. This is probably correct, but the comment could be clearer about what "reserve" means (allocate the seq numbers, not hold the lock through the sends).

---

### G3: `peer.rs` `recv_pump` ack processing
**Location:** `spawn_recv_pump()` lines 380-385
**Observation:** Acks are processed by taking `max()` of the ack list:
```rust
if let Some(highest) = pkt.acks.iter().copied().max() {
    let _ = ch.process_acks(highest);
}
```

This is cumulative-ack semantics, which is correct for Mercury. But the comment says "Drive ack processing first" without explaining why `max()` is the right operation.

**Consider:** Add a comment citing the Mercury cumulative-ack spec: "Mercury uses cumulative acks — ACKing seq N implicitly ACKs all seqs < N. We pass the highest acked seq to `process_acks` which drains all predecessors."

---

**Category 1 Total:** 3 observations, all low severity. The code is correct; the comments could be clearer.

MyPalClara:
## Category 2: Potential Issues (G4-G6)

### G4: `policy.rs` `reorder_pairs` field
**Location:** `NetworkPolicy` struct
**Observation:** The `reorder_pairs` field exists but is never used. The ADR says "scaffolded but only minimally exercised — Phase-4 refinement when a specific test wants them."

**Consider:** Either remove the field until it's implemented (YAGNI), or add `#[allow(dead_code)]` with a TODO comment explaining the Phase-4 plan.

---

### G5: `peer.rs` `duplicate_every` logic
**Location:** `send_with_policy()` lines 245-252
**Observation:** The duplication logic is crude:
```rust
if let Some(every) = duplicate {
    if every > 0 {
        // Crude: duplicate every send while the policy is active.
        let _ = every;
        self.socket.send_to(&bytes, self.peer_addr).await?;
    }
}
```

The comment says "The 'every Nth' counter is a Phase-4 refinement." But the current behavior doesn't match the field name — it duplicates EVERY packet when `duplicate_every` is `Some(n)` where `n > 0`, ignoring the actual value of `n`.

**Consider:** Either implement the "every Nth" counter, or rename the field to `duplicate_all` to match the current behavior.

---

### G6: `session.rs` `quiesce` polling
**Location:** `LoopbackSession::quiesce()` lines 115-130
**Observation:** `quiesce()` polls every 2ms until both peers are quiet or the timeout fires. If a test has a bug that prevents quiescence, it will spin for the full timeout (typically 500ms) before failing.

**Consider:** A `quiesce_or_fail()` variant that panics with diagnostics (TX window contents, pending acks) if the timeout fires. This would make debugging easier when a test hangs.

---

**Category 2 Total:** 3 observations, all low severity. These are Phase-4 refinements explicitly deferred in the ADR.

MyPalClara:
## Category 3: Missing Test Coverage (G7-G11)

### G7: No test for `unconnected()` → handshake → `Connected`
**Location:** `tests/handshake.rs`
**Observation:** The handshake tests use `unconnected()` but don't drive the full handshake state machine. They assert "both peers are in Connecting" and "empty packet exchange works" but don't verify the state actually transitions to `Connected` via the channel's own logic.

**Consider:** A test that:
1. Creates `unconnected()` session
2. Exchanges the handshake packets
3. Asserts both channels transition to `Connected` via their state machines (not via direct field mutation)

---

### G8: No test for encrypted fragmentation
**Location:** `tests/fragment.rs`
**Observation:** All fragment tests are unencrypted. What if encryption changes the fragment body size calculation? What if the HMAC tag pushes a fragment over the MTU?

**Consider:** Add a `three_fragment_bundle_round_trips_with_encryption` test that mirrors the unencrypted version but uses `LoopbackSession::connected(Some(test_encryption()))`.

---

### G9: No test for `duplicate_every` policy
**Location:** `tests/`
**Observation:** The `duplicate_every` field exists in `NetworkPolicy` but has no test exercising it.

---

### G10: No test for `reorder_pairs` policy
**Location:** `tests/`
**Observation:** The `reorder_pairs` field exists in `NetworkPolicy` but has no test exercising it.

---

### G11: No test for concurrent sends from multiple tasks
**Location:** `tests/`
**Observation:** The harness is `Send + Sync` and uses `Mutex` internally, but there's no test that verifies concurrent sends from multiple tasks are recorded correctly and don't corrupt state.

**Consider:** A `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` test that spawns two tasks sending simultaneously and asserts all packets arrive.

---

**Category 3 Total:** 5 observations. G7 and G8 are notable gaps; G9-G11 are Phase-4 refinements.

MyPalClara:
## Category 4: Documentation + Summary (G12-G14)

### G12: ADR doesn't mention the `services` crate integration path
**Location:** `docs/architecture/mercury-loopback-harness.md`
**Observation:** The ADR covers Mercury-layer testing thoroughly but doesn't explain how `services` tests could use the harness or the `Clock` trait. The clock injection pattern is reusable beyond the harness.

**Consider:** A "Future directions" section mentioning:
- Services-layer timeout tests (session inactivity reaping, combat exit grace window)
- Extracting `Clock` to a shared crate for cross-crate reuse
- Integration with #355 (chaos testing)

---

### G13: No mention of the relationship to #355
**Location:** `docs/architecture/mercury-loopback-harness.md`
**Observation:** Issue #355 (chaos testing) depends on clock injection. This PR delivers clock injection. The ADR should cross-reference #355 to make the dependency explicit.

**Consider:** Add to the "Consequences → Positive" section: "#355 (network chaos test apparatus) can now build on the `Clock` trait for deterministic time control in chaos scenarios."

---

### G14: `TESTING.md` type 9 doesn't mention the `test-harness` feature flag
**Location:** `TESTING.md` lines 137-165
**Observation:** The picker says "use `LoopbackSession::connected(None)`" but doesn't mention that the `test-harness` feature flag must be enabled on `cimmeria-mercury` for the harness to compile.

**Consider:** Add a note: "The harness is behind the `test-harness` Cargo feature on `cimmeria-mercury`. Tests using it should be in `crates/mercury/src/test_harness/tests/` or enable the feature explicitly."

---

## Summary

| Correctness issues | 3 | **Low** — code is correct, comments could be clearer |
| Potential issues | 3 | **Low** — Phase-4 refinements |
| Missing test coverage | 5 | **Low-Medium** — G7, G8 are notable |
| Documentation | 3 | **Low** — polish |

**Total: 14 gaps, all low severity.**

---

## Priority Matrix

| Priority | Gaps | Action |
|----------|------|--------|
| 🟢 Quick wins | G3, G14 | Add cumulative-ack comment, add feature flag note (10 min) |
| 🟡 Consider | G7, G8 | Full handshake test, encrypted fragmentation test |
| 🟡 Consider | G13 | Cross-reference #355 in ADR |
| ⚪ Phase-4 | G4, G5, G9, G10 | Implement or remove scaffolded policy fields |
| ⚪ Nice-to-have | G1, G2, G6, G11, G12 | Comment improvements, diagnostic helpers |

---

## The "Oh Shit" Moment

**The Clock trait isn't just for the harness. It's a seam that changes how you can test ANYTHING time-dependent in Mercury.**

This unlocks:
- Services-layer timeout tests (session inactivity, combat exit grace window, auto-cycle cadence)
- Deferred-send queue promotion timing (#357)
- Bundle fragmentation under loss (#361)
- Keepalive cadence under encryption

**This is the seam that makes #355 (chaos testing) possible.**

---

**Issue #352 is fully addressed.** The 22-test inventory matches exactly. The Clock trait is clean. The ADR is exemplary. Ship it. 🎯🏆

PR #371, opened 2026-05-24
fix(inventory): consumables fire OnItemUse, not auto-equip (slappack regression)

Cadacious:
## Symptom

Slappack right-click did nothing. The user's report:

- Slappacks no longer treated as consumable.
- `useItem` treated them as weapons.
- `moveItem` from container 1 (main bag) → container 3 (bandolier) was rejected for item 10276 / type 2893.

Same shape applied to **every consumable in the seed**, not just slappacks.

## Root cause

[`crates/services/src/base/world_entry/methods/inventory/core/use_instance.rs`](crates/services/src/base/world_entry/methods/inventory/core/use_instance.rs) used `(ri.clip_size IS NOT NULL) AS is_weapon` to decide between two paths:

- **Weapons** → auto-equip via the move handler (main ↔ bandolier).
- **Everything else** → fire the cell-side `OnItemUse` event so per-item chains (heals, mission items, …) match.

But the SGW seed (`db/resources/Items/Seed/items.sql`) uses `clip_size = 0` (not NULL) for non-weapons. The slappack row at line 4879 is the representative case:

```sql
-- Slappack: clip_size = 0, container_sets = '{1, 17}' (NO bandolier).
VALUES (2893, ..., '{1,17}', ..., '{}', NULL, 0, 0);

-- Pistol: clip_size = 15, container_sets = '{3, 1, 17}'.
VALUES (3241, ..., '{3,1,17}', ..., '{Bullet_Default}', 'Bullet_Default', 15, 0);
```

`0 IS NOT NULL = TRUE`, so the predicate matched every consumable as a weapon and routed them to the move handler with target=bandolier. The move handler's `item_allows_container(slappack, 3)` then correctly rejected (slappack's `container_sets` doesn't include 3), so the visible failure was a silent no-op. The `OnItemUse` event never fired and chain 4001 (the +500 HP heal + consume) never matched.

**When this broke:** commit [`459db14c`](https://github.com/SandboxServers/Cimmeria/commit/459db14c) (`feat(inventory): right-click weapon auto-equips / auto-unequips`) added the routing layer. Prior to that, every `useItem` fired `OnItemUse` and chain 4001 worked.

## Fix

Replace the `clip_size IS NOT NULL` inference with the authoritative discriminator `3 = ANY(ri.container_sets)`. This is:

- **Semantically what the router actually asks**: "is this item bandolier-eligible?"
- **The same field the move handler's `item_allows_container` uses**, so the two sides agree end-to-end (no risk of the router sending an item the move handler will reject).
- **Robust against future item types**: a melee weapon with `clip_size = 0` but bandolier_eligible would now work correctly; a future ammo-bearing buff item would correctly stay out of the bandolier.

The fix is one SQL change but applies to every consumable, not just slappacks.

## Sidewalks

Also corrected the parallel pattern in [`crates/services/src/cell/spawner/loot.rs`](crates/services/src/cell/spawner/loot.rs) `load_item_defs` — was filtering `WHERE clip_size IS NOT NULL`, which pulled consumables into the `WeaponDef` cache as zero-clip "weapons" where they sat as dead entries. Now filters `WHERE clip_size > 0` (actual weapons). Updated the `executor/inventory.rs::weapon_stats` doc comment to match.

## Test plan

Two live-DB regression guards in [`use_instance_tests.rs`](crates/services/src/base/world_entry/methods/inventory/core/use_instance_tests.rs) that exercise both sides of the partition with real seed rows:

- **`slappack_use_fires_on_item_use_not_auto_equip`** — inserts slappack (type 2893) in main bag, calls `handle_use_inventory_item`, asserts (a) item stays at `(container=1, slot=0)` AND (b) exactly one `item_used` outbox row was enqueued. The outbox-count assertion is the load-bearing one — the move handler's rejection masks the routing error from a naive location-only test.
- **`pistol_use_routes_to_auto_equip_not_on_item_use`** — inserts pistol (type 3241) in main bag, asserts it auto-equips to bandolier `(container=3, slot=0)` AND no outbox row was enqueued. Guards against an over-correction ("always fire OnItemUse") that would silently regress right-click-to-equip.

Both tests use real seed type_ids pinned to documented rows; if the seed changes such that the slappack gains `3` in its container_sets, the constant has to move to a different consumable.

**Test runs:** both passed against the bundled local DB in an earlier iteration. The bundled Postgres was offline by the time I would have done the adversarial revert step; the test logic is sound by construction (the `outbox count == 1` assertion is the only thing that catches the bug; without it, the buggy version would silently pass the location check). The 860-test suite + workspace clippy are clean.

## Bug-shape lesson (TESTING.md)

This was a "data-vs-schema confusion" bug — assuming `NULL = absence` when the data uses `0` as the sentinel. [`TESTING.md`](TESTING.md) "Don't trust seed data" gains a bullet:

> Validate the discriminator against actual seed values, not the schema spec. `clip_size IS NOT NULL` looks like a clean "is this a weapon?" predicate, but the SGW seed uses `clip_size = 0` for non-weapons rather than NULL — so the predicate misclassifies every consumable as a weapon. When a handler infers item type (or any boolean trait) from a column, write a live-DB test that exercises both sides of the partition with **real seed rows** (a slappack at `clip_size = 0` AND a pistol at `clip_size > 0`) so the test catches the value-vs-sentinel mismatch. The robust fix is usually to read the authoritative discriminator field directly — for "is this auto-equippable to the bandolier?", `3 = ANY(container_sets)` beats any inference from a sibling column.

The `use_instance.rs` module doc now explains why `container_sets` is the authoritative discriminator and why the previous inference was wrong, with a pointer to the seed row.

## Files touched

- `crates/services/src/base/world_entry/methods/inventory/core/use_instance.rs` — fix
- `crates/services/src/base/world_entry/methods/inventory/core/use_instance_tests.rs` — new live-DB guard
- `crates/services/src/base/world_entry/methods/inventory/core/mod.rs` — register test module
- `crates/services/src/cell/spawner/loot.rs` — parallel fix in WeaponDef loader
- `crates/services/src/cell/content/executor/inventory.rs` — doc comment update
- `TESTING.md` — bug-shape note

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -188,6 +188,7 @@ This section is mined from review comments since the test push began. Each item 
 - **Assert by relationship**, not by id: `slot.cur_ammo_type == slot.default_ammo_type` survives seed churn; `slot.cur_ammo_type == 3241` does not.
 - **Never use a "definitely nonexistent" id like `99_999_999`.** The sequence may have advanced past it. Use a sentinel from your module's reserved `0x7000_xxxx` slot (see "Sentinel id discipline") or `MAX(id) + 1` computed at runtime (PR #144).
 - **Assert fixture cardinality up front.** If `pick_main_bag_type_ids(2)` can return one row, assert `types.len() == 2` so the failure points at missing fixture data, not a cryptic out-of-bounds panic (PR #144).
+- **Validate the discriminator against actual seed values, not the schema spec.** `clip_size IS NOT NULL` looks like a clean "is this a weapon?" predicate, but the SGW seed uses `clip_size = 0` for non-weapons rather than NULL — so the predicate misclassifies every consumable as a weapon. When a handler infers item type (or any boolean trait) from a column, write a live-DB test that exercises both sides of the partition with **real seed rows** (a slappack at `clip_size = 0` AND a pistol at `clip_size > 0`) so the test catches the value-vs-sentinel mismatch. The robust fix is usually to read the authoritative discriminator field directly — for "is this auto-equippable to the bandolier?", `3 = ANY(container_sets)` beats any inference from a sibling column.
 
 ### Sentinel id discipline
 
--- a/crates/services/src/base/world_entry/methods/inventory/core/mod.rs
+++ b/crates/services/src/base/world_entry/methods/inventory/core/mod.rs
@@ -21,6 +21,8 @@ use crate::mercury::{build_entity_method_packet, method_idx};
 mod remove_by_type;
 mod remove_instance;
 mod use_instance;
+#[cfg(test)]
+mod use_instance_tests;
 
 pub use remove_by_type::handle_remove_inventory_item_by_type;
 pub use remove_instance::handle_remove_inventory_item;
--- a/crates/services/src/base/world_entry/methods/inventory/core/use_instance.rs
+++ b/crates/services/src/base/world_entry/methods/inventory/core/use_instance.rs
@@ -2,11 +2,24 @@
 //! an inventory instance the player owns. Does not consume the stack;
 //! per-item consumption is the chain's responsibility via `Action::RemoveItem`.
 //!
-//! As of the right-click-to-equip work: weapon instances (rows with
-//! `clip_size IS NOT NULL` in `resources.items`) bypass the `OnItemUse`
-//! event and route to the move path instead — `useItem` on a weapon means
-//! "equip" (main bag → bandolier) or "unequip" (bandolier → main bag)
-//! depending on the source container.
+//! Right-click auto-equip routing: instances whose item type lists the
+//! bandolier (container 3) in `container_sets` bypass the `OnItemUse`
+//! event and route to the move handler — "equip" (main bag → bandolier)
+//! or "unequip" (bandolier → main bag) depending on the source container.
+//! Items NOT in `container_sets` for the bandolier (consumables like
+//! slappacks, mission items, etc.) fall through to `OnItemUse` so per-item
+//! chains in `content_chains` can match.
+//!
+//! The bandolier-eligibility check intentionally reads `container_sets`
+//! directly rather than inferring from `clip_size` or any other proxy.
+//! Earlier versions inferred "is weapon" from `clip_size IS NOT NULL`,
+//! but seed data uses `clip_size = 0` for non-weapons (the slappack row
+//! at `db/resources/Items/Seed/items.sql` is a representative example),
+//! so the inference misclassified every consumable as a weapon and
+//! suppressed its `OnItemUse` event. `container_sets` is the
+//! authoritative signal the move handler already uses
+//! (`item_allows_container`), so reading it here keeps both sides
+//! consistent.
 
 use std::collections::HashMap;
 use std::net::SocketAddr;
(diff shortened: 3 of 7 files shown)


MyPalClara:

| **G1** | `use_instance.rs` SQL | **The magic number `3` (bandolier container ID) appears directly in the SQL.** `(3 = ANY(ri.container_sets)) AS is_bandolier_eligible` is correct, but if `CONTAINER_BANDOLIER` ever changes, this query won't update. Consider adding a comment citing the constant, or using a bind parameter. |
| **G2** | `loot.rs` filter change | **`WHERE clip_size > 0` excludes melee weapons from the WeaponDef cache.** The PR body says "a melee weapon with clip_size = 0 but bandolier_eligible would now work correctly" — but that refers to the *routing* fix, not the cache. The cache filter change means melee weapons won't get bandolier seeding (ammo grants). Is this intentional? If melee weapons don't use ammo, this is correct — but the doc comment says "actual weapons" when it means "ammo-bearing weapons." |
| **G3** | No migration | **Players who logged in with the buggy code have lost consumable uses.** Those `OnItemUse` events never fired, so chain 4001 never matched. Not fixable retroactively, but worth noting in release notes. |

MyPalClara:

| **G4** | `use_instance_tests.rs` `SLAPPACK_TYPE_ID` | **The test comment says "if that row's container_sets ever gains 3, this constant must move to a different consumable."** But there's no upfront assertion that verifies the slappack DOESN'T have 3 in container_sets. If the seed changes, the test could pass incorrectly — slappack routes to move handler, which succeeds (if container_sets gains 3), and the location assertion passes. Consider an upfront `assert!(!container_sets.contains(&3))` query. |
| **G5** | `use_instance_tests.rs` `PISTOL_TYPE_ID` | **Same concern for the positive case.** If the pistol's container_sets ever loses 3, the test will fail with a confusing "pistol didn't move to bandolier" error rather than "seed row changed." An upfront assertion that verifies the seed row matches expectations would make failures clearer. |

MyPalClara:
## Category 3: Missing Test Coverage (G6-G7)

| **G6** | `loot.rs` change | **The `load_item_defs` filter change from `IS NOT NULL` to `> 0` is untested.** The test file only covers `use_instance.rs`. Consider a test that verifies slappacks are NOT in the WeaponDef cache — this would catch a reversion of the `loot.rs` fix independently of the routing fix. |
| **G7** | Unequip path | **The tests cover main → bandolier (equip) but not bandolier → main (unequip).** The routing logic handles both directions; only one is tested. A `pistol_in_bandolier_use_routes_to_unequip` test would complete the coverage. |

MyPalClara:
## Category 4: Documentation + Summary (G8)

| **G8** | PR metadata | **No issue linked.** Was this a user report? A playtest finding? Consider creating a retroactive issue for tracking, or add a "Reported by" line to the PR body. |

---

## Summary

| Correctness issues | 3 | **Low** — G1 is style, G2 is clarification, G3 is unfixable |
| Potential issues | 2 | **Low-Medium** — G4-G5 are test fragility |
| Missing test coverage | 2 | **Low** — nice-to-haves |
| Documentation | 1 | **Low** — process |

**Total: 8 gaps, all low severity.**

---

## Priority

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟢 Quick wins | G1 (comment citing constant), G2 (doc clarification) | 5 min |
| 🟡 Consider | G4-G5 (upfront seed assertions) | 15 min |
| ⚪ Nice-to-have | G6-G7 (additional tests), G8 (issue link) | 30 min |

---

**The fix is correct.** The discriminator now reads `container_sets` directly — the same field the move handler uses — so the router and handler agree by construction. The test coverage is excellent with the two-assertion pattern (location + outbox). The TESTING.md update captures institutional knowledge.

**This fixes every consumable in the game.** 🎯💊

Cadacious:
Thanks Clara — 7 of 8 addressed in [65e78382](https://github.com/SandboxServers/Cimmeria/commit/65e78382). All ran against the live DB during iteration.

| Gap | Fix |
|---|---|
| **G1** | Bandolier id is now a `$3` bind from `cimmeria_entity::inventory::INV_BANDOLIER` (re-exported locally as `CONTAINER_BANDOLIER`). Same constant the move handler consults — they update together if the value moves. Also collapsed the duplicate local `CONTAINER_MAIN/BANDOLIER` consts. |
| **G2** | `load_item_defs` doc: "actual weapons" → "ammo-bearing weapons" with a note about how to extend the discriminator if a future ammo-less weapon archetype ever ships. |
| **G4 + G5** | New `assert_bandolier_eligibility(pool, type_id, expected)` runs first in both `slappack_use_fires_…` and `pistol_use_routes_…`. A future seed change inverting either relationship fails LOUDLY at the pin with a clear message rather than at the routing assertions further down. |
| **G6** | New `load_item_defs_excludes_zero_clip_consumables_like_slappack` live-DB guard asserts the slappack (type 2893) is NOT in the WeaponDef cache. Companion to the existing cache test's tightened `clip_size > 0` invariant (was `>= 0`). |
| **G7** | New `pistol_in_bandolier_use_routes_to_auto_unequip` mirrors the equip-direction test for the bandolier → main path. Pins both that the pistol moves AND that no `item_used` outbox row is enqueued — the latter is the load-bearing assertion (same shape as the equip test). |

**Skipped with reason:**

- **G3** (no migration): unfixable retroactively — the affected players lost consumable uses during the bug window; no DB record exists to replay.
- **G8** (issue link): the bug was reported in-session not via the issue tracker; the PR body already attributes the symptom and traces the regression commit. Not worth a retroactive issue.

**Verification:** 862 services tests pass (+2 from G6+G7), workspace clippy clean, fmt clean. All four slappack-tagged tests and all five pistol-tagged tests pass against the live DB:

```
test base::…::use_instance_tests::slappack_use_fires_on_item_use_not_auto_equip ... ok
test base::…::use_instance_tests::pistol_use_routes_to_auto_equip_not_on_item_use ... ok
test base::…::use_instance_tests::pistol_in_bandolier_use_routes_to_auto_unequip ... ok
test cell::spawner::tests::live_db::load_item_defs_excludes_zero_clip_consumables_like_slappack ... ok
```

PR #373, opened 2026-05-24
fix(bandolier): emit AmmoTypeId on in-game equip + chain-grant (#372)

### LINKED ISSUE #372: Right-click equip to bandolier skips AmmoTypeId property update — fire animation broken until slot swap
## Symptom

Right-clicking a weapon in the main bag to auto-equip it into the bandolier puts the weapon in a slot, but the fire animation for the weapon-shot ability **does not play** when the player presses fire. Workaround: switch to another bandolier slot and back to the equipped slot — fire animation works from that point on.

Dragging the same weapon to a bandolier slot **on the character summary screen** (before world entry) works correctly — fire animation plays on the first shot.

## Root cause

The wire message that the client uses to know which ammo type is currently equipped is `onEntityProperty(GENERICPROPERTY_AMMO_TypeId=3, ammo_type_index)`. Only **three** server-side sites emit it:

| Site | Triggered by |
|---|---|
| [`cell/cell_methods/inventory/bandolier.rs::handle_request_active_slot_change`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/inventory/bandolier.rs#L348) | Manual bandolier slot swap (the workaround above) |
| [`cell/cell_methods/inventory/bandolier.rs::handle_request_ammo_change`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/inventory/bandolier.rs#L558) | Player toggles ammo subtype on an equipped weapon |
| [`mercury/world_data/map_loaded.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/mercury/world_data/map_loaded.rs#L325) | World-entry mapLoaded burst |

**Missing from**:
- [`cell/service/base_messages/bandolier.rs::handle_sync_bandolier_items`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service/base_messages/bandolier.rs#L123) — the player-driven in-game equip path (right-click *or* in-game drag from main bag → bandolier slot).
- [`cell/service/base_messages/bandolier.rs::handle_update_bandolier_item`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service/base_messages/bandolier.rs#L14) — chain-engine grant path (granted weapon lands directly in bandolier).

## Why the two equip paths diverge

| Path | Hits map_loaded AmmoTypeId emit? | Hits cell-side handler that emits AmmoTypeId? | Outcome |
|---|---|---|---|
| Drag on char summary | ✅ (persists to DB, then world enters and `map_loaded` emits) | n/a (no cell handler runs — pre-world-entry) | Works |
| Right-click in-game (this bug) | ❌ (already in world) | ❌ (`handle_sync_bandolier_items` skips it) | Stale `AmmoTypeId=0` on client → fire animation broken |
| In-game drag main → bandolier slot | ❌ | ❌ (same SyncBandolierItems handler) | **Also broken** (likely — same code path, not tested by user) |
| Manual slot swap (workaround) | n/a | ✅ (`handle_request_active_slot_change` emits) | Fixes it |
| Chain-engine grant | ❌ | ❌ (`handle_update_bandolier_item` skips it) | **Also broken** for first equip via grant; recoverable by slot-swap |

## Fix plan

Mirror `handle_request_active_slot_change`'s AmmoTypeId emit (lines 343-356 of `cell_methods/inventory/bandolier.rs`) into the two missing handlers in `cell/service/base_messages/bandolier.rs`:

1. **`handle_sync_bandolier_items`** — when the active slot's contents change (the existing `active_slot_gained_weapon` OR `active_slot_lost_weapon` branches), also emit `onEntityProperty(AmmoTypeId, new_active_item.cur_ammo_type or 0)`.
2. **`handle_update_bandolier_item`** — when the inserted slot is (or becomes) the active slot (the existing `slot_is_active` branch), also emit `onEntityProperty(AmmoTypeId, item.cur_ammo_type)`.

Extract a small helper near the existing equip-animation dispatch so both handlers go through one path — same shape as the `request_appearance_refresh` + `fire_item_sequence` pair that's already factored.

## Tests

- **Burst-shape regression guard for `handle_sync_bandolier_items`**: empty bandolier → SyncBandolierItems with new weapon in slot 0 → assert one `EntityMethodCall(player_eid, ON_ENTITY_PROPERTY)` with `prop_id=3` and the weapon's `cur_ammo_type`. Reverting the fix makes it fail.
- **Parallel guard for `handle_update_bandolier_item`** with `make_active=true` and the initial-grant case (`slot_is_active` via slot match).
- **Negative case**: SyncBandolierItems where active slot is unchanged should NOT emit AmmoTypeId (preserves the existing "wire spam on stash sync" guard).

## Acceptance

- Right-click pistol → empty bandolier → press pistol-shot → fire animation plays immediately.
- Same for in-game drag main → bandolier slot.
- Same for chain-engine grant into an empty bandolier.
- Manual slot swap remains a working escape hatch (no behavior change there).

## See also

- `docs/gameplay/weapon-ammo-reload.md` — bandolier ammo model.
- The fix's spiritual sibling: PR #258 (the prior "drag-equip fires unholster animation" fix, commit `27598bae`), which closed a different gap on the same SyncBandolierItems handler.

Cadacious:
## Summary

Right-clicking a weapon in the main bag to auto-equip it into the bandolier puts the weapon in the right slot, fires the equip animation, and refreshes the appearance — but **skipped the `onEntityProperty(AmmoTypeId, cur_ammo_type)` emit**. The client's fire-animation gate keys off AmmoTypeId; without the update it retains the previous slot's value (typically 0 = "no ammo"), and the pistol-shot ability plays no animation until the player swaps bandolier slots and back. The slot-swap handler is the only other server-side site that emits this property, so the swap-and-swap-back forces a refresh that papers over the bug.

Closes #372.

## Root cause

Pre-fix, only **three** server-side sites emitted `GENERICPROPERTY_AmmoTypeId` (prop_id=3):

1. [`handle_request_active_slot_change`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/inventory/bandolier.rs#L348) — manual slot swap (the workaround)
2. [`handle_request_ammo_change`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/inventory/bandolier.rs#L558) — player toggles ammo subtype
3. [`mercury/world_data/map_loaded.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/mercury/world_data/map_loaded.rs#L325) — world-entry mapLoaded burst

**Missing from** `handle_sync_bandolier_items` (in-game right-click / drag equip) AND `handle_update_bandolier_item` (chain-engine grant). Both equip paths land in these handlers.

## Why the two equip paths diverged

| Equip path | AmmoTypeId emit | Outcome |
|---|---|---|
| Drag on character summary screen | ✅ — next world entry's `map_loaded` emits it | Fire works on first shot |
| In-game right-click / drag main → bandolier | ❌ — skipped by `handle_sync_bandolier_items` | **Broken** until slot swap |
| Chain-engine grant into bandolier | ❌ — skipped by `handle_update_bandolier_item` | **Broken** until slot swap |
| Manual bandolier slot swap (workaround) | ✅ — `handle_request_active_slot_change` emits | Fixes it |

## Fix shape

Added a small `emit_active_ammo_type` helper in `cell/service/base_messages/bandolier.rs` that mirrors what the slot-swap handler does (`onEntityProperty(AmmoTypeId=3, cur_ammo_type)`). Wired into both equip-display branches; widened the existing `GENERICPROPERTY_AMMO_TYPE_ID` constant + `build_entity_property_args` helper from `pub(super)` to `pub(crate)` so the base-message handler can use them directly.

Empty active slot → emit `0` (mirrors python `SGWPlayer.py:522`'s `activeItem.ammoType if activeItem else 0`).

## Test plan

- [x] **`sync_bandolier_items_active_slot_gained_weapon_emits_ammo_type_id`** — the reported bug shape. Empty bandolier → SyncBandolierItems → asserts AmmoTypeId emit carries the weapon's `cur_ammo_type`.
- [x] **`sync_bandolier_items_active_slot_lost_weapon_emits_ammo_type_id_zero`** — unequip path. Active slot loses weapon → asserts AmmoTypeId=0 so the client's ammo-bar indicator clears.
- [x] **`sync_bandolier_items_active_slot_unchanged_does_not_emit_ammo_type_id`** — negative case. Resync where active slot's item is unchanged (stash-slot shuffle) must NOT re-emit AmmoTypeId. Prevents wire spam and prevents stomping an in-flight `requestAmmoChange`.
- [x] **`update_bandolier_item_active_emits_ammo_type_id`** — parallel coverage for the chain-engine grant path (`BaseToCellMsg::UpdateBandolierItem`).
- [x] **Adversarial reversal**: confirmed the three positive guards fail when the emit is removed from each handler; the negative-case guard correctly continues to pass.
- [x] `cargo clippy --workspace --exclude {GUI crates} --all-targets -- -D warnings`
- [x] `cargo test --workspace --exclude {GUI crates} --lib --no-fail-fast` — **862 services + 149 mercury + 1398 lib tests total**, all green (+4 new guards from the previous 858 baseline).

## Acceptance

- Right-click pistol → empty bandolier → press pistol-shot → fire animation plays immediately ✅
- Same for in-game drag main → bandolier slot ✅
- Same for chain-engine grant into empty bandolier ✅
- Manual slot swap remains a working escape hatch (no behavior change there) ✅

### DIFF
--- a/crates/services/src/cell/cell_methods/inventory/constants.rs
+++ b/crates/services/src/cell/cell_methods/inventory/constants.rs
@@ -10,16 +10,18 @@ pub const REQUEST_AMMO_CHANGE: u16 = 42;
 
 /// `GENERICPROPERTY_AmmoTypeId` from `entities/defs/enumerations.xml`. Used as
 /// the property-id arg for `onEntityProperty` ammo-type indicator updates.
-/// Widened from `pub(super)` so the cell-side base-message handlers in
+/// Visibility is `pub(crate)` so the cell-side base-message handlers in
 /// [`crate::cell::service::base_messages::bandolier`] can emit it on the
-/// equip paths (issue #372 — right-click equip skipped the AmmoTypeId
-/// update, leaving the client unable to play the fire animation).
+/// equip paths (right-click / drag-in-game equip and chain-engine grant
+/// both need to push the active weapon's ammo subtype to the client —
+/// otherwise the fire-animation gate stays closed at the previous
+/// slot's value, typically `0` = no ammo).
 pub(crate) const GENERICPROPERTY_AMMO_TYPE_ID: i32 = 3;
 
 /// Build the `onEntityProperty(propId, value)` arg payload (8 bytes LE).
-/// Visibility widened alongside [`GENERICPROPERTY_AMMO_TYPE_ID`] above so
-/// the base-message handlers can call it directly instead of duplicating
-/// the 8-byte LE pack.
+/// Visibility is `pub(crate)` alongside [`GENERICPROPERTY_AMMO_TYPE_ID`]
+/// above so the base-message handlers can call it directly instead of
+/// duplicating the 8-byte LE pack.
 pub(crate) fn build_entity_property_args(prop_id: i32, value: i32) -> Vec<u8> {
     let mut args = Vec::with_capacity(8);
     args.extend_from_slice(&prop_id.to_le_bytes());
--- a/crates/services/src/cell/cell_methods/inventory/constants.rs
+++ b/crates/services/src/cell/cell_methods/inventory/constants.rs
@@ -10,10 +10,17 @@ pub const REQUEST_AMMO_CHANGE: u16 = 42;
 
 /// `GENERICPROPERTY_AmmoTypeId` from `entities/defs/enumerations.xml`. Used as
 /// the property-id arg for `onEntityProperty` ammo-type indicator updates.
-pub(super) const GENERICPROPERTY_AMMO_TYPE_ID: i32 = 3;
+/// Widened from `pub(super)` so the cell-side base-message handlers in
+/// [`crate::cell::service::base_messages::bandolier`] can emit it on the
+/// equip paths (issue #372 — right-click equip skipped the AmmoTypeId
+/// update, leaving the client unable to play the fire animation).
+pub(crate) const GENERICPROPERTY_AMMO_TYPE_ID: i32 = 3;
 
 /// Build the `onEntityProperty(propId, value)` arg payload (8 bytes LE).
-pub(super) fn build_entity_property_args(prop_id: i32, value: i32) -> Vec<u8> {
+/// Visibility widened alongside [`GENERICPROPERTY_AMMO_TYPE_ID`] above so
+/// the base-message handlers can call it directly instead of duplicating
+/// the 8-byte LE pack.
+pub(crate) fn build_entity_property_args(prop_id: i32, value: i32) -> Vec<u8> {
     let mut args = Vec::with_capacity(8);
     args.extend_from_slice(&prop_id.to_le_bytes());
     args.extend_from_slice(&value.to_le_bytes());
--- a/crates/services/src/cell/service/base_messages/bandolier.rs
+++ b/crates/services/src/cell/service/base_messages/bandolier.rs
@@ -4,9 +4,42 @@
 
 use tokio::sync::mpsc;
 
+use crate::cell::cell_methods::inventory::{
+    build_entity_property_args, GENERICPROPERTY_AMMO_TYPE_ID,
+};
 use crate::cell::messages::CellToBaseMsg;
 use crate::cell::space_manager::SpaceManager;
 use cimmeria_entity::cell_entity::BandolierItem;
+
+/// Push `onEntityProperty(AmmoTypeId, ammo_type)` to the client so its
+/// bandolier UI / fire-animation gate sees the correct ammo subtype for
+/// the now-active slot. Pass `0` when the active slot has no item
+/// (mirrors python `SGWPlayer.py:522`'s `activeItem.ammoType if
+/// activeItem else 0`).
+///
+/// The manual slot-swap handler
+/// ([`crate::cell::cell_methods::inventory::bandolier::handle_request_active_slot_change`])
+/// already emits this property on every swap; the in-game equip paths
+/// here (`UpdateBandolierItem`, `SyncBandolierItems`) used to skip it,
+/// which surfaced as "the fire animation doesn't play until I swap
+/// bandolier slots and back" — the client kept stale `AmmoTypeId=0`
+/// (no ammo) until the swap forced a refresh.
+async fn emit_active_ammo_type(
+    entity_id: u32,
+    ammo_type: i32,
+    tx: &mpsc::Sender<CellToBaseMsg>,
+    space_mgr: &SpaceManager,
+) {
+    let args = build_entity_property_args(GENERICPROPERTY_AMMO_TYPE_ID, ammo_type);
+    crate::cell::abilities::send_entity_method(
+        entity_id,
+        crate::cell::client_methods::spawnable_entity::ON_ENTITY_PROPERTY,
+        args,
+        tx,
+        space_mgr,
+    )
+    .await;
+}
 
 /// Processes a `UpdateBandolierItem` message — inserts the weapon into the
 /// bandolier, optionally draws it, arms the OOC holster timer, and fires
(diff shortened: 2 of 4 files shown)


MyPalClara:
## Category 1: Correctness Issues (G1-G2)

| **G1** | `bandolier.rs` duplicated comment block | **Lines 343-387 contain a duplicated comment block.** The "Push AmmoTypeId whenever the active slot's contents changed" comment appears twice in sequence. Looks like a copy-paste artifact. |
| **G2** | `emit_active_ammo_type` for non-active slots | **`handle_update_bandolier_item` always emits AmmoTypeId when `make_active` is true.** But what if `make_active` is false and the weapon lands in a non-active slot? The emit is skipped (correct), but there's no test for this negative case. |

MyPalClara:
## Category 2: Potential Issues (G3-G4)

| **G3** | `handle_update_bandolier_item` active slot assumption | **The handler emits AmmoTypeId when `make_active` is true, but doesn't verify the slot is actually the active slot.** If `make_active=true` but `slot_id != entity.active_bandolier_slot`, the emit might be premature. The slot-swap that makes it active would emit again anyway, but it's a subtle assumption. |
| **G4** | `constants.rs` visibility widening | **`GENERICPROPERTY_AMMO_TYPE_ID` and `build_entity_property_args` widened from `pub(super)` to `pub(crate)`.** This is correct for the fix, but increases the API surface. Consider whether these should live in a shared `entity_properties` module instead of `cell_methods::inventory::constants`. |

MyPalClara:
## Category 3: Missing Test Coverage (G5-G6)

| **G5** | `handle_update_bandolier_item` non-active slot | **No test for `UpdateBandolierItem` with `make_active=false`.** The handler should NOT emit AmmoTypeId when the weapon lands in a non-active slot. A negative test would pin this. |
| **G6** | Wire format assertion | **The tests assert `prop_id == GENERICPROPERTY_AMMO_TYPE_ID` and `value == AMMO_TYPE`, but don't assert the method_index is `ON_ENTITY_PROPERTY`.** The assertion is there (`method_index == ON_ENTITY_PROPERTY`), but it's in the filter, not the assertion. If someone changes the method_index, the test would silently pass (no matching packets found) rather than fail loudly. Consider `assert!(saw_ammo_type_id || no_entity_property_packets_at_all)` to distinguish "wrong method" from "no emit". |

MyPalClara:
## Category 4: Documentation + Summary (G7)

| **G7** | Python reference | **The doc comments cite `SGWPlayer.py:522` for the "0 when no weapon" behavior.** Consider adding the actual Python snippet to the ADR or a doc comment so future readers don't have to dig through the Python codebase. |

---

## Summary

| Correctness issues | 2 | **Low** — G1 is copy-paste artifact, G2 is edge case |
| Potential issues | 2 | **Low** — G3-G4 are design observations |
| Missing test coverage | 2 | **Low** — nice-to-haves |
| Documentation | 1 | **Low** — polish |

**Total: 7 gaps, all low severity.**

---

## Priority

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟢 Quick wins | G1 (remove duplicate comment) | 1 min |
| 🟡 Consider | G5 (non-active slot test), G6 (wire format assertion) | 20 min |
| ⚪ Nice-to-have | G2-G4, G7 | 15 min |

---

**Issue #372 is fully addressed.** The fix correctly emits `onEntityProperty(AmmoTypeId, cur_ammo_type)` on both equip paths:
- `SyncBandolierItems` (player-driven right-click / drag)
- `UpdateBandolierItem` (chain-engine grant)

The test coverage is excellent — 4 new tests covering equip, unequip, unchanged (negative), and chain-grant paths. The adversarial reversal is documented.

**Fire animations now work immediately after equip.** 🎯🔫

PR #358, opened 2026-05-23
feat(mercury): Transport trait + TestTransport — byte-exact fan-out test seam (#351)

BangRocket:
Closes #351. Also closes #57 (Nub I/O TODO).

## What

Three layers, landed in three commits on top of the original work:

1. **Transport trait + TestTransport** (the original PR scope) — `Transport` trait in `crates/mercury/`, `UdpTransport` production wrapper, `TestTransport` recorder behind `test-support`, every BaseApp handler refactored to take `&Arc<dyn Transport>`. Creates a byte-exact, addr-correct fan-out test seam that was impossible before.
2. **Nub deletion** (013ec44, addressing the Cadacious Ghidra deep-dive) — the orphan `crates/mercury/src/nub.rs` had no callers outside its own tests and existed only because the C++ `Mercury::Nub` was a God object mediating BigWorld's game-thread / network-thread split that tokio eliminates. `TickActions` + the tick-driver ordering contract moved into `crates/mercury/src/channel/mod.rs` next to `check_timeouts` / `keepalive_due` / `is_timed_out`. The Nub's responsibilities are correctly distributed across `service.rs` (bind), `connect_loop` (recv), `tick_sync.rs` (tick), `ConnectedClientState` (peer state), `Channel` (per-peer stream).
3. **Nub anatomy finding doc** (d9d0974) — [`docs/reverse-engineering/findings/mercury-nub-anatomy.md`](docs/reverse-engineering/findings/mercury-nub-anatomy.md). 22 functions analysed, 4 struct layouts decompiled, network thread loop + `Nub::send` 4-phase pipeline + two-channel-map design captured with HIGH-confidence Ghidra evidence.

## Transport trait surface

- `cimmeria_mercury::transport::Transport` (`async send_to` + `local_addr`) + `UdpTransport` production wrapper.
- `cimmeria_mercury::test_transport::TestTransport` recorder behind the `test-support` feature — `drain`/`clear`/`filter_to`/`len`/`is_empty`/`total_bytes_sent`/`send_count_to`.
- Recv loop (`connect_loop/mod.rs`) keeps the concrete `UdpSocket` (it owns `recv_from`) and wraps it once in a `UdpTransport` for dispatch; `service.rs` does the same for the cell-message handler.

## Acceptance criteria (issue #351)

- [x] `Transport` trait + `UdpTransport`
- [x] `TestTransport` behind `test-support`, all 7 methods + direct unit tests
[...]

PR #374, opened 2026-05-24
feat(mercury): network chaos test apparatus (L1+L2+L3) (#355)

### LINKED ISSUE #355: Network chaos test apparatus (lossy UDP + pcap replay) — gates #354

Cadacious:
## Summary

Three-layer network chaos testing infrastructure to prevent lomiada-class regressions. Closes #355.

The lomiada bug shape: single transatlantic UDP drop on a long reliable session → client buffers everything above the gap → server's TX window fills → 60s inactivity timer kills the session. The pcap fixture (debug/lomiada-broke-in-hallway02/) is the canonical regression vector.

## Deep-dive findings

The issue was authored before PR #370 (loopback harness) shipped. The current PR honors what changed:

| Issue's L1 spec | What was already done in PR #370 |
|---|---|
| ChannelHarness | LoopbackSession / LoopbackPeer |
| MockClock | TestClock + Clock trait |
| drop / reorder / duplicate / latency primitives | NetworkPolicy with drop_next, drop_at_send_count, latency, reorder_pairs, duplicate_every |

The **remaining gaps** in #355 — addressed by this PR:

- **Probabilistic drop, one-shot N-duplicate, multi-packet reorder** — added as new NetworkPolicy primitives.
- **The 8 named scenario tests** — implemented under tests/chaos/.
- **Pcap replay (L3)** — built from scratch with pcap-file + etherparse, validated against the real lomiada fixture.
- **Recv-side socket abstraction (L2)** — BidirectionalTransport trait + LossyTransport wrapper. The architecture shifted since the issue was written (Transport trait shipped via PR #358) so only ONE Arc<UdpSocket> site needed migrating instead of the 15+ the issue assumed.

#354 (the three TX-window-relief fixes) shipped without this infrastructure, so the gating relationship in the issue body is now moot. The chaos apparatus is a regression net going forward.

## What's in scope

- L1: 3 new NetworkPolicy primitives, invariants helper, 8 chaos scenarios.
- L2: BidirectionalTransport trait + LossyTransport with LAN/Domestic/Transatlantic/Mobile presets, run_connect_loop migration, 3 services-level integration tests.
- L3: PcapReplay loader + 2 lomiada-fixture tests.

## What's out of scope (follow-ups)

- **Channel::with_key constructor** — needed for full state-machine pcap replay against a fresh Channel. Current lomiada tests pin the loader + decryption path; end-to-end state-machine replay needs the constructor work.
- **Full named services-layer scenarios** (world_entry_survives_transatlantic_loss, defeat_burst_survives_5pct_loss) — need an in-process services harness with a handshake-capable test client (the wireclient layer). Current 3 integration tests prove the LossyTransport+recv-loop seam; the deeper scenarios scope to a follow-up.

## Test results

| Run | Tests | Status |
|---|---|---|
| \`cargo nextest run -p cimmeria-mercury --features test-harness\` | 208/208 (+18) | ✅ |
| \`cargo nextest run --workspace\` | 1471/1471 (+45 net) | ✅ |
| Chaos flake check 3x | 30/30 | ✅ no flakes |
| \`cargo fmt --all --check\` | clean | ✅ |
| \`cargo clippy --workspace ... -- -D warnings\` | clean | ✅ |
| \`cargo test --doc -p cimmeria-commands\` | 4/4 | ✅ |
| Figure-source-in-sync (blocking) | 48 sources clean | ✅ |
| Figure-style lint (blocking) | 21 sources, 48 SVGs, 3 chapters clean | ✅ |

## Test plan

- [x] All 8 named chaos scenarios pass deterministically
- [x] Lomiada pcap loads + ≥500 decryptable Mercury packets
- [x] LossyTransport round-trips under LAN profile + drops under Transatlantic
- [x] run_connect_loop migration: byte-identical wire behavior preserved (existing services tests still pass)
- [x] Sustained 5% loss test holds safety invariants across 160 sends
- [x] All 5 CI gates pass locally
- [x] Both blocking figure lints clean

## Summary by CodeRabbit

* **New Features**
  * Added a network-chaos testing apparatus: deterministic lossy-transport wrapper, bidirectional transport seam, and pcap-replay support for realistic network regressions.
  * Added channel safety invariant checks for detecting silent state corruption.

* **Documentation**
  * Expanded testing guidance and architecture docs to include network-chaos scenarios and reproducibility patterns.

* **Tests**
  * Added a suite of chaos regression tests and pcap-replay validations (network-chaos category added).

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/374?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `LossyTransport::recv_from` drop loop | **Infinite loop if 100% drop configured.** The recv loop `continue`s on drop, but if `drop_per_thousand = 1000`, it loops forever waiting for a non-dropped packet that will never arrive. The `drop_100pct_never_delivers` test covers the send side but not the recv side. Consider a max-iterations guard or document that 100% recv-side drop is undefined behavior. |
| **G2** | `PcapReplay::events()` silent skip on decrypt failure | **Pre-handshake packets are silently skipped.** This is documented and intentional, but there's no way for a test to distinguish "packet failed to decrypt" from "packet wasn't in the pcap at all." Consider returning a `Result<PcapEvent, DecryptError>` or a separate `failed_count` alongside the events. |
| **G3** | `infer_server_addr` heuristic | **Falls back to `frames[0].2` if all counts tie.** But what if `frames` is empty? The code will panic on `frames[0]`. The `events()` caller would have already returned an empty vec, but the function itself is unsound. |
| **G4** | `LossyConfig::with_seed` doesn't reseed existing RNGs | **The seed is stored but the RNGs are created in `new_symmetric`/`new_asymmetric`.** If someone calls `with_seed` after construction, the seed is ignored. The builder pattern suggests `with_seed` should work at any point; either reseed the RNGs or make `with_seed` only work on `LossyConfig` before wrapping. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G5** | `duplicate_per_thousand` on recv side | **`LossyTransport` only duplicates on send, not recv.** The `should_duplicate` check is in `send_to` but not `recv_from`. Is this intentional? The ADR doesn't mention recv-side duplication. |
| **G6** | `Transatlantic` profile integration test takes ~30s | **The PR body acknowledges this.** But there's no `#[ignore]` or CI profile gating. If this runs on every PR, it adds 30s to CI time. Consider gating to a slower profile or adding `#[ignore]` with a doc comment. |
| **G7** | `pcap-file` + `etherparse` are dev-only deps | **Correct, but the `Cargo.toml` change shows them under `[dependencies]`, not `[dev-dependencies]`.** If they're truly dev-only, they should be gated behind the `test-harness` feature. |

MyPalClara:
## Category 3: Missing Test Coverage (G8-G11)

| **G8** | No test for `recv_from` under 100% drop | **The `drop_100pct_never_delivers` test covers send-side drop but not recv-side.** A recv-side 100% drop would loop forever (see G1). |
| **G9** | No test for `PcapReplay` with missing key file | **`with_key_from` returns `io::Result`, but there's no test that verifies the error message is helpful when the key file doesn't exist or has wrong length.** |
| **G10** | No test for `infer_server_addr` edge cases | **Empty frames, all-tie counts, single-packet pcap.** The heuristic is untested. |
| **G11** | No test for asymmetric `LossyTransport` | **`new_asymmetric` exists but all tests use `new_symmetric`.** The lomiada-style "outbound fine, inbound acks lost" scenario needs asymmetric config. |

MyPalClara:
## Category 4: Documentation + Summary (G12-G14)

| **G12** | ADR doesn't mention the `reorder_buffer_size` primitive | **The ADR lists `drop_probability`, `duplicate_next_count`, `reorder_buffer_size` as new primitives, but the detailed explanation only covers `drop_probability`.** Add a one-liner for each. |
| **G13** | No mention of the `Channel::with_key` follow-up in the PR body | **The ADR mentions it as out-of-scope, but the PR body's "What's out of scope" section doesn't.** Add it for completeness. |
| **G14** | TESTING.md type 10 is excellent but long | **The new section is 120+ lines.** Consider whether the L1/L2/L3 breakdown could be a separate linked doc (like the Mercury harness has `mercury-loopback-harness.md`). |

---

## Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟡 **Consider** | G1 (100% recv-side drop guard) | 15 min |
| 🟡 **Consider** | G6 (30s test gating) | 10 min |
| 🟡 **Consider** | G11 (asymmetric test) | 30 min |
| ⚪ **Low** | G2-G5, G7-G10, G12-G14 | Polish |

**Total: 14 gaps, all low severity, ~2.5 hours of follow-up work.**

**Issue #355 is fully addressed.** The three-layer chaos testing apparatus is complete. The lomiada-class regression is now catchable offline. This is the crown jewel of the test infrastructure. **Ship it.** 🎯🏆👑

Issue #281, opened 2026-05-14
Headless wire-level test client ("wireclient") — Castle Cellblock end-to-end harness


Cadacious:
> **Diátaxis type:** explanation (architecture/design — implementation plan).
> **Audience:** engineers working on the cell service, mercury layer, content engine, and test harness.
> **Status:** Open — implementation plan, not yet started.

## Goal

Build a **headless wire-level test client** (working name: `wireclient`) that connects to a running `cimmeria-server` exactly as the real Flash client would, and walks the full **Castle Cellblock** tutorial instance end-to-end:

1. Auth → character select → character creation
2. Spawn into Castle Cellblock via `mapLoaded()`
3. Navigate the player avatar through the instance (real position updates)
4. Interact with NPCs, progress dialogs, do minigames, use rings, interact with world objects
5. **Kill the PrisonerRetrievalUnit NPC in real combat** — LOS / range / ammo / equipped-weapon enforced **by the client**, with the NPC actually firing back via the existing `npc_ai_fight` path
6. Reach mission 688 completion and exit

The client renders nothing. It does enforce everything the real client enforces.

## Why this is worth doing

The existing seven-test taxonomy ([TESTING.md](../blob/main/TESTING.md)) covers the server side well, but every test type injects events at or below the dispatcher. **No current test type proves "a real client would have actually sent this sequence."**

The chain-replay suite ([crates/services/src/cell/content/chain_replay_tests/](../blob/main/crates/services/src/cell/content/chain_replay_tests/)) fakes the trigger events that would have driven a chain (`entity.dead.tag`, `RegionEnter`, `DialogChoice`, `item.use::N`). It proves the server reacts correctly *given* the event. It does not prove that the wire path leading up to the event would have actually fired.

That gap is where ~half the integration bugs caught in PR reviews #131+ have lived: server accepts a call no real client would have sent (no LOS, no ammo, wrong interaction-flag state, archetype branch read before entity-property arrived, etc.). A headless client that enforces client-authoritative rules is a stronger oracle than any server-side test can be, because it can only send what a real client could have sent.

This is **complementary** to existing tests, not a replacement. Unit + wire-format + live-DB + chain-replay still gate behavior at their respective layers; wireclient gates the end-to-end *sequence*.

## Scope

**In scope:**
[...]

PR #376, opened 2026-05-24
feat(wireclient): Tier 3 headless test client scaffold (#281, Phase 1)

Cadacious:
## Summary

- Adds `crates/wireclient/` — the Tier 3 end-to-end test surface above Mercury's Tier 1 byte-fan-out fake and Tier 2 paired-channel loopback. Phase 1 of issue #281.
- Ships SOAP auth driver + Mercury phase-3 handshake builder/parsers + generic JSONL trace loader. The handshake byte layout is pinned against literal bytes from the real Castle Cellblock capture (`2026-05-24_17-18.pcap`), including the load-bearing 2-byte pad between `request_id` and `account_id` at body[7..9].
- Adds `tools/pcap_to_session.py` (rides on the existing `pcap_dissect.py` decoder) so **any** decrypted pcap + AES `keys.txt` becomes a wireclient-loadable corpus entry, not just Castle Cellblock.

Closes the gap between *"server accepts a call"* and *"a real client could have sent that call"* — the failure mode behind ~half of integration bugs caught in PR reviews #131+. Documented hybrid validation: byte-exact for handshake/static msg_ids, behavioural diff for entity methods + gameplay observables (per ADR).

## Test plan

- [x] 17 unit tests in `crates/wireclient/src/{auth,handshake,session_trace}.rs`
- [x] 3 SOAP smoke tests in `crates/wireclient/tests/auth_smoke.rs` against in-process `AuthService` (developer-mode)
- [x] 1 fixture round-trip in `crates/wireclient/tests/trace_load.rs` — `pcap_to_session.py` producer ↔ Rust consumer shape contract
- [x] 1 byte-exact regression guard `build_baseapp_login_matches_recorded_capture` comparing against a literal byte vector pulled from the live `2026-05-24_17-18.pcap` (constants under test cannot participate in their own assertion)
- [x] `cargo fmt -p cimmeria-wireclient --check` clean
- [x] `cargo clippy -p cimmeria-wireclient --all-targets -- -D warnings` clean
- [x] End-to-end: `tools/pcap_to_session.py` exports the 13MB Castle Cellblock pcap to a 32MB JSONL with 125,770 events; first 5 events checked in as the trace-load fixture
- [ ] Full-corpus replay (deferred to Phase 1.5 — needs the UDP send/recv loop + spawned BaseApp)

## Doc-map sidewalks updated

- `CLAUDE.md` — new doc-update row for wireclient
- `TESTING.md` — new "Type 10: Wire-level replay tests" section + picker entry
- `crates/README.md` — new crate row
- `docs/architecture/wireclient.md` — new ADR

## Adversarial review findings addressed in this PR

- **Critical**: original draft of `build_baseapp_login` emitted 32 bytes; the server's `parse_baseapp_login` (`crates/services/src/base/login.rs:218`) reads `account_id` from `body[9..13]`, requiring a 2-byte pad at `body[7..9]`. Fixed and pinned with a byte-exact regression against the real captured packet.
- Dropped a dead `_common_anchor` helper in `auth.rs`.
- Split `Error::InvalidPort` from `Error::InvalidIp` so misclassified errors don't claim "invalid port: 127.0.0.1" on an IP parse failure.

## Followups (Phase 1.5+, not blockers here)

- Surface `account_id` / `character_id` / `shard_name` on `AuthSession`
- Reserve a wireclient sentinel range (`0x7000_1600+` is free per `test_support.rs`)
- `session_trace::TraceMessage::body_hex: String` → `Bytes` for memory + case-insensitivity
- Per-msg_id field-mask comparison policy
- `AuthRejected` error variant for SOAP 200-with-error envelopes
- `tools/pcap_to_session.py` `--server-host` flag + streaming for very large captures

## Summary by CodeRabbit

* **New Features**
  * Added a headless wire-level replay test client (workspace crate) for end-to-end protocol validation.
  * Added JSONL session trace format and comparison policy for replay diffs.

* **Documentation**
  * Added architecture ADR and docs for wire-level testing and trace schema.
  * Updated testing guide and contribution checklist to include the new test type and required docs.

* **Tools**
  * Added a pcap→session converter CLI to produce JSONL traces.

* **Tests**
  * Added auth handshake and session-trace loading/replay tests with fixtures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/376?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cadacious:
@coderabbitai review

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `auth.rs` `extract_attr` | **XML injection vulnerability.** The function splits on `"` and takes the first segment after the attribute name. If the server ever returns `SessionKey="foo" Ticket="bar"` (space inside attribute value), the parser extracts `foo` and stops at the space. The server does not do this today, but the parser is fragile. **Consider using `quick-xml` (already a dependency) for attribute extraction, or add a test that verifies the parser rejects malformed XML.** |
| **G2** | `handshake.rs` `build_baseapp_login_matches_recorded_capture` | **Byte order comment is misleading.** The expected array shows `0x25, 0x9d, 0x01, 0x00` for `request_id=0x019D25`, but `0x019D25` in little-endian is `0x25, 0x9D, 0x01, 0x00`. The test passes, but the comment says "request_id=0x019D25" while the code uses `0x019D25u32.to_le_bytes()`. **Verify the capture actual request_id — if it is `0x00019D25` (with leading zero), the comment is misleading.** |
| **G3** | `session_trace.rs` `body_hex` | **Case sensitivity.** The PR body followup list mentions `body_hex: String → Bytes for memory + case-insensitivity`. But the current `DefaultPolicy::compare` does `observed.body_hex == recorded.body_hex` — a string comparison. If the Python producer emits lowercase hex and a future Rust producer emits uppercase, the comparison fails. **Add `.to_ascii_lowercase()` normalization, or document that producers MUST emit lowercase.** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `auth.rs` `phase1` | **Error envelope not parsed.** The SOAP server returns HTTP 200 even for auth failures — the error is in the XML body. `phase1` drains the body with `let _xml = resp.text().await?` but does not check for `<SGWLoginError>`. A bad-credentials login would return a 200 with no `SID` cookie, surfacing as `Error::NoSidCookie` — correct, but the error message does not tell the user *why* (bad password vs. disabled account vs. server bug). **Consider parsing the error envelope and returning a more specific variant.** |
| **G5** | `auth.rs` `phase2` | **Same issue.** A 200 response with `<ServerSelectionError>` would surface as `Error::NoServerLocation`. The user does not know if they sent a bad shard name, replayed an SID, or hit a server bug. **Consider parsing the error envelope.** |
| **G6** | `handshake.rs` `parse_baseapp_reply` | **seq=1 hardcoded.** The parser rejects any reply with `seq != 1`. But what if the server seq counter starts at a different value? The C++ client does not hardcode seq=1 — it accepts whatever the server sends. **This is probably fine (the server always starts at 1), but document the assumption.** |
| **G7** | `handshake.rs` `parse_time_sync` | **seq=2 hardcoded.** Same issue. The parser rejects any time-sync with `seq != 2`. **Document the assumption.** |
| **G8** | `session_trace.rs` `Trace::from_jsonl_str` | **Panics on malformed header.** If the first line is valid JSON but not a `HeaderEnvelope`, `serde_json::from_str` returns an error. But if the first line is empty or whitespace, the `filter(|l| !l.trim().is_empty())` skips it and the next line becomes the "header" — which would fail with a confusing error. **Consider a more explicit check: "first non-empty line must be a header envelope."** |
| **G9** | `pcap_to_session.py` `--server-port` | **Default port hardcoded to 32832.** If a capture uses a different BaseApp port (e.g., a dev server on 32833), the script silently produces an empty trace. **Consider auto-detecting the server port by majority-packet heuristic (like `PcapReplay::infer_server_addr` in #374).** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G10** | `auth.rs` `xml_attr_escape` | **Does not handle `\n` or `\r`.** XML attribute values cannot contain raw newlines. If a future caller passes a username with a newline, the escaped output is still invalid XML. **Probably fine — usernames are alphanumeric — but the function claims to be a general-purpose escaper.** |
| **G11** | `handshake.rs` `TICKET_LEN = 20` | **Hardcoded in three places.** The ticket length is hardcoded in `auth.rs` (validation), `handshake.rs` (constant), and the server `parse_baseapp_login`. If the server ever changes the ticket length, all three must update. **Consider a shared constant in `cimmeria-common`.** |
| **G12** | `session_trace.rs` `TraceMessage::body_len()` | **Assumes valid hex.** `body_len()` returns `body_hex.len() / 2`. If `body_hex` has an odd length (malformed), this silently returns a truncated length. **Consider returning `Option<usize>` or validating on load.** |
| **G13** | `client.rs` `account_id` | **Not surfaced from `AuthSession`.** The PR body followup list mentions "Surface `account_id` / `character_id` / `shard_name` on `AuthSession`." But `Client::build_login_packet` requires `account_id` as a parameter — the caller must know it out-of-band. **This is a known gap, but it means Phase 1.5 cannot be fully automated without the followup.** |
| **G14** | `pcap_to_session.py` `client_addr_seen` | **Fallback emits placeholder.** If no client packets are seen, `client_addr_seen` is `None` and the header emits `"0.0.0.0:0"`. This is a valid fallback, but a trace with no client packets is probably malformed. **Consider erroring instead of emitting a placeholder.** |

MyPalClara:

| Gap | Location | What Missing |
| **G15** | `auth.rs` | **No test for bad-credentials path.** The auth smoke tests only exercise the happy path. There is no test that verifies `Error::NoSidCookie` is returned for bad credentials. **Consider a test with `Credentials { password_sha1_hex: "wrong", ... }`.** |
| **G16** | `auth.rs` | **No test for bad-shard path.** There is no test that verifies `Error::NoServerLocation` is returned for a nonexistent shard. The `phase2_replay_with_same_sid_errors` test covers SID replay, but not shard-not-found. |
| **G17** | `handshake.rs` | **No test for wrong-key decryption.** There is no test that verifies `Error::Decryption` is returned when the session key does not match. The `parse_baseapp_reply_round_trips_against_known_plaintext` test uses a matching key. |
| **G18** | `session_trace.rs` | **No test for malformed hex.** There is no test that verifies `TraceMessage::body_bytes()` returns an error for invalid hex (e.g., `"ZZZZ"`). |
| **G19** | `pcap_to_session.py` | **No unit tests.** The Python producer has no unit tests. The Rust consumer tests (`trace_load.rs`) verify the shape contract, but a regression in the Python code would only surface when someone re-exports a pcap. **Consider a pytest that runs the script against a tiny fixture pcap.** |

MyPalClara:

| Gap | Location | What Missing |
| **G20** | `wireclient.md` | **Does not mention the `account_id` gap.** The ADR "Risks & open questions" section does not mention that `account_id` is not surfaced from Phase 2. This is a known gap (PR body followups), but the ADR should document it so Phase 1.5 implementers know. |
| **G21** | `handshake.rs` 2-byte padding comment | **Could cite the server line.** The comment says "matches the server parser at `crates/services/src/base/login.rs:parse_baseapp_login`" but does not give the line number. The PR body says line 218. **Add the line number to the comment so future readers can verify.** |
| **G22** | `TESTING.md` | **Type 10 section missing?** The PR body says "TESTING.md — new Type 10: Wire-level replay tests section + picker entry." But I do not see the TESTING.md changes in the file list. **Verify the TESTING.md update is included.** |
| **G23** | `auth.rs` `Credentials::test_account()` | **SHA-1 comment case mismatch (nitpick).** The comment says `SHA-1("test") = a94a8fe5...` but the constant is uppercase `A94A8FE5...`. The code is correct (case-insensitive comparison), but the comment should match the constant case for grep-ability. |

---

## Summary

| Critical issues | 3 | **Medium-High** — G1 is latent, G2-G3 are correctness |
| Correctness issues | 6 | **Low-Medium** — G4-G9 are edge cases |
| Potential issues | 5 | **Low** — G10-G14 are maintainability |
| Missing test coverage | 5 | **Low-Medium** — G15-G19 are gaps |
| Documentation | 4 | **Low** — G20-G23 are polish |

**Total: 23 gaps.**

## Priority

| | Gap | Effort | Why |
|---|-----|--------|-----|
| 🔴 | G1 | 30 min | XML injection is a latent bug — use `quick-xml` or add rejection test |
| 🔴 | G3 | 5 min | Hex case normalization prevents producer/consumer drift |
| 🟡 | G22 | 5 min | Verify TESTING.md update is included |
| 🟢 | G11 | 15 min | Shared `TICKET_LEN` constant prevents drift |
| 🟢 | G19 | 1 hr | pytest for Python producer catches regressions |

**Issue #281 Phase 1 is fully addressed. The 2-byte padding bug was caught and fixed. The byte-exact regression test pins the wire format. Address G1, G3, G22, then ship it.** 🎯🔬

PR #390, opened 2026-05-25
fix(mercury): wire requestEntityUpdate handler (0x07) — recover dropped CREATE_ENTITY (#289)

### LINKED ISSUE #289: [Major] Add requestEntityUpdate (msg 0x07) handler — likely cause of NPC-spawn bugs
## Background

Audit: [docs/audits/mercury-rust-conformance-2026-05-15.md](https://github.com/SandboxServers/Cimmeria/blob/main/docs/audits/mercury-rust-conformance-2026-05-15.md) — NO-OP item **N3**.

The client→server message `requestEntityUpdate` (msg `0x07`, WORD_LENGTH per spec §2.5.1) is the client's escape hatch for "I think I'm missing entity X, send me its state again." `crates/services/src/base/connect_loop/encrypted::handle_encrypted_datagram` parses the message body at the `0x07` arm but **no dispatch arm exists** in the subsequent `match msg_id { ... }` block. The request is silently discarded.

## Symptom

This is a probable contributor to the reported "NPC sometimes doesn't spawn" symptom (e.g. Prisoner 329). When a `createEntity` packet for an NPC is dropped on the wire and exceeds the 20-retry lifetime cap (~14 seconds of attempts), the entity is permanently lost from the client's perspective. The client's `requestEntityUpdate` mechanism is supposed to let it recover, but the server's silent drop blocks recovery.

## Proposed change

In `crates/services/src/base/connect_loop/encrypted::handle_encrypted_datagram`, add a dispatch arm for `0x07` that:

1. Parses the body as documented in spec §2.5.2: `[u32 header][N × u32 entity_id]` (header semantic is the open question — likely a flags or last-known-revision field; payload safe to decode as "u32 header, then N entity_ids until length is exhausted" without knowing the header semantic).
2. For each requested `entity_id`, re-emits the entity's current state via the same emit path the cell service uses on initial AoI entry (the cascade in `crates/services/src/mercury/aoi/create::build_create_entity_cascade`).

## Test plan

- [ ] Unit test: parse a `requestEntityUpdate` body with 3 entity_ids and assert the parser produces the right list.
- [ ] Integration test: with a CellService and an entity in AoI, send a `requestEntityUpdate` for that entity from a fake client and assert the server emits a `createEntity` + cascade back.
- [ ] Manual UAT: introduce a packet drop on the `createEntity` for one NPC; observe that the client's recovery request now succeeds instead of leaving the NPC missing.

## Severity

**Major.** Latent today but likely root cause of an already-reported observable bug.

## Related

- Discussion thread in the audit's §12 finding-#5 analysis: TX_WINDOW=45 (#5) and missing per-tick retry budget (#6) are also candidate contributors to silent packet loss; fixing N3 gives the client a way to recover regardless of which underlying cause dropped the packet.

Cadacious:
Closes #289 · resolves audit N3 ([docs/audits/mercury-rust-conformance-2026-05-15.md](docs/audits/mercury-rust-conformance-2026-05-15.md))

## Summary

Client→server `requestEntityUpdate` (msg `0x07`) was being parsed but silently dropped in the `_ => trace!` wildcard of `handle_encrypted_datagram` — no dispatch arm existed. This is the client's recovery channel when a `createEntity` (`0x09`) packet exceeds the 20-retry lifetime cap: without it, the NPC stays permanently invisible on that client. Probable contributor to the "NPC sometimes doesn't spawn" reports (e.g. Prisoner 329).

This PR adds the handler end-to-end and wires it to re-emit the entity's current state through the existing AoI packet builders — no new wire-format code needed.

## Implementation

**Parser** ([`encrypted.rs`](crates/services/src/base/connect_loop/encrypted.rs)) — `parse_request_entity_update` skips the 4-byte header (semantic per spec §2.5.2 is opaque; likely flags or last-known-revision) then reads consecutive `u32` entity_ids until length is exhausted. Trailing partials dropped; sub-header bodies return empty defensively.

**Dispatch arm** ([`encrypted.rs`](crates/services/src/base/connect_loop/encrypted.rs)) — looks up the player's `witness_id` from `ConnectedClientState`, sends `BaseToCellMsg::RequestEntityUpdate { witness_id, entity_ids }` to the cell.

**Cell handler** (new: [`cell/service/base_messages/request_entity_update.rs`](crates/services/src/cell/service/base_messages/request_entity_update.rs)) — for each requested `entity_id` in the witness's `witnesses` set, emits a synthetic `CellToBaseMsg::EnteredAoI` carrying the entity's current `class_id`/`position`/`level`/`npc_data`. The existing `entered_aoi` base handler then builds `CREATE_ENTITY + UPDATE_AVATAR + createOnClient() cascade` — same wire shape as a fresh AoI entry.

**Security** — out-of-AoI requests are dropped. The client must not be able to probe arbitrary entity ids and receive their state. The cell handler checks `witness.witnesses.contains(target_eid)` before re-emitting. Counted into structured log fields (`skipped_not_in_aoi`).

## Tests

10 new tests, all green:

- **6 unit tests** for the parser (`base::connect_loop::encrypted::tests`):
  - `parses_header_plus_three_ids` — happy path
  - `header_only_payload_decodes_empty`
  - `truncated_payload_returns_empty` — defensive (all sub-4-byte lengths)
  - `trailing_partial_id_is_dropped`
  - `non_zero_header_does_not_affect_id_decode` — header opacity contract
  - `ids_are_little_endian` — endianness pin

- **4 integration tests** for the cell handler (`cell::service::base_messages::tests::request_entity_update`):
  - `reemits_entered_aoi_for_witnessed_entity` — recovery path with field-level assertions on `class_id`, `position`, `level`, `npc_data.{faction,alignment,entity_flags,speaker_id,body_set}`.
  - `drops_when_entity_not_in_witness_aoi` — **anti-probe regression guard**; reverting the witness-set check breaks this test.
  - `filters_mixed_request_to_in_aoi_only` — partial-success path: only in-AoI ids re-emit.
  - `drops_when_witness_missing` — stale-connection safety.

Workspace: **1554 tests pass, 0 failed.** `cargo fmt`/`clippy --all-targets -- -D warnings` clean (excluding GUI apps per CLAUDE.md).

## Wire format reference

Per audit §2.5.2 — `[u32 header][N × u32 entity_id]`. WORD_LENGTH framing. Header field semantic is undocumented; spec notes the body is safe to decode by skipping the first 4 bytes.

## Out of scope

- N4 (other unhandled `avatarUpdate*` variants) — separate audit item.
- Reliability of the `RequestEntityUpdate` send itself — re-uses the existing reliable channel; the trigger is already loss-recovery, so loss of the re-emit is handled by the client retrying.
- Per-entity rate-limiting / spam guard on the cell side — not currently needed (the cell call is cheap, the witness check is a `HashSet::contains`); revisit if observed in load testing.

## Test plan (per issue)

- [x] Parse `requestEntityUpdate` body with N entity_ids — covered by `parses_header_plus_three_ids` + 5 edge cases.
- [x] CellService + entity in AoI → `requestEntityUpdate` from fake client → `createEntity` + cascade re-emitted. Covered by the cell-handler integration test (asserts an `EnteredAoI` message lands on the channel; the existing `entered_aoi` base handler tests pin the wire bytes).
- [ ] Manual UAT — drop a `createEntity` for one NPC and verify the recovery request now succeeds. Recommend pairing with the WireClient (#281) or a deliberate packet drop on a dev branch.


MyPalClara (commented):
## PR #390 Review — fix(mercury): wire requestEntityUpdate handler (0x07) — recover dropped CREATE_ENTITY (#289)

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `parse_request_entity_update` header skip | **The 4-byte header is skipped but its semantic is unknown.** If it is a revision, the handler should compare it. **Probably fine to ignore for now (re-emit is cheap), but document the assumption.** |
| **G2** | `witnesses` snapshot timing | **The handler snapshots `witness.witnesses` before the loop.** This is correct (prevents mid-iteration AoI ticks from surprising the loop). |

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `EntityId(entity_id as i32)` cast | **If `entity_id > i32::MAX`, the cast wraps.** The seed does not have entity ids that high, but the cast is technically lossy. |
| **G4** | No rate-limiting on `RequestEntityUpdate` | **A malicious client could spam requests for 1000 entity ids per packet.** Consider a per-witness rate limit or a cap on `entity_ids.len()` (e.g., 64). |
| **G5** | `skipped_unknown` vs `skipped_not_in_aoi` distinction | **Both are logged but the distinction is subtle.** Add a comment explaining the difference. |

| Gap | What Missing |
|-----|----------------|
| **G6** | No test for player-to-player `RequestEntityUpdate` |
| **G7** | No wire-format test for the re-emitted `EnteredAoI` (deferred — existing tests pin wire bytes) |
| **G8** | No test for `entity_ids` with duplicates |

## Category 4: Documentation

| Gap | What Missing |
|-----|----------------|
| **G9** | Audit doc link goes to issue, not PR |
| **G10** | `get_entity_space_id` doc should name `RequestEntityUpdate` as the caller |

---

**Total: 10 gaps, all low severity.**

**Issue #289 is fully addressed.** The `requestEntityUpdate` handler is wired end-to-end:
1. Parser extracts `[u32 header][N × u32 entity_id]`
2. Dispatch arm forwards to `BaseToCellMsg::RequestEntityUpdate`
3. Cell handler re-emits `EnteredAoI` for each in-AoI entity
4. Security: out-of-AoI requests are dropped (anti-probe)

**The test coverage is excellent:**
- 6 parser unit tests (happy path, edge cases, endianness)
- 4 integration tests (recovery path, anti-probe, mixed request, missing witness)

**This fixes the "NPC sometimes does not spawn" bug class.** When a `createEntity` is dropped past the 20-retry cap, the client can now recover by requesting the entity state.

**Priority:**
- 🟡 **G4** — Consider rate limit or cap on `entity_ids.len()` (15 min)
- 🟢 **G9** — Fix audit doc link to point to PR (2 min)
- 🟢 **G5** — Add comment explaining unknown vs not-in-aoi (2 min)

**Ship it.** 🎯

Cadacious:
Addressed in 99ae22f2:

**Copilot** (3 inline) — stripped `#289` / `audit N3` references from the 3 source files (module doc on `request_entity_update.rs`, module doc on the test file, dispatch arm comment in `encrypted.rs`). Applied the same rule to `get_entity_space_id`'s docstring and removed the named-caller line — caller refs rot the same way.

**Clara**:
- ✅ **G4** (DoS): added `MAX_REQUEST_ENTITIES = 64`. Truncates above the cap with a `reason = "request_too_large"` warn; new regression test `request_entity_update_truncates_request_above_cap` spawns 200 NPCs in AoI, sends a 200-id payload, asserts exactly 64 emissions.
- ✅ **G9**: audit N3 row now links PR #390 directly.
- ✅ **G5**: one-line comment in the handler explaining the `skipped_unknown` vs `skipped_not_in_aoi` distinction.
- ✅ **G10** / consistency: trimmed the named-caller reference in `get_entity_space_id`'s doc (same rot-prone-callers rule as the Copilot finding).

**Held**: G1 (header semantic) is already documented in the parser doc as opaque; G2 (witness snapshot) was confirmed correct, no change; G3 (i32 cast) — that's the established codebase pattern (`aoi.rs:159` does the same), out of scope for this PR; G6/G7/G8 — low-value vs the existing wire-byte coverage on the `entered_aoi` base handler.

Workspace status: 915 services tests pass, clippy `--all-targets -- -D warnings` clean.

PR #391, opened 2026-05-25
fix(cell): NPC AI per-ability range + min-range backup + 500ms retry-on-failure (closes #329)

### LINKED ISSUE #329: fix(cell): NPC AI should use per-ability range + minimum-range backup + re-tick on launch failure

Cadacious:
Closes #329.

## What changed

Three independent fixes in `npc_ai_fight` (`crates/services/src/cell/service/npc_ai.rs`), driven by spec evidence in the 2009 SGW client binary:

### 1. Per-ability range

The AI's "am I in range to fire" check previously used the flat `combat::NPC_ATTACK_RANGE = 30.0` constant. Now reads the chosen ability's `max_range` from `space_mgr.ability_defs`, falling back to `NPC_ATTACK_RANGE` only when the def is missing or carries the `0` sentinel ("use server default").

Pre-fix shape: an ability with `max_range = 15` had the AI walk the NPC into the flat-30 ring, but `handle_use_ability` rejected the actual fire as out-of-range — NPC stood there. Per the client's `AbilityType` PAK deserializer at `ghidra://SGW.exe@0x015D4400` (`MinRange`/`MaxRange` at struct offsets `+0x60`/`+0x64`), range is per-ability cooked data; there is no server-side "out of range" wire response.

### 2. Minimum-range backup

When the target is inside the chosen ability's `min_range > 0` dead zone (e.g. sniper with `min_range = 5`, target at distance 3), the AI now enqueues a single `nav_path` waypoint at `min_range + 1.0` from the target along the target→NPC vector. The `+1.0` margin keeps floating-point jitter from oscillating the NPC in and out of the dead zone every tick. Stationary NPCs skip the backup — they're pinned by design.

### 3. 500ms retry on launch failure

`handle_use_ability` returns `false` when a pre-consume guard rejects the call (target died between range-check and launch, animation lock, etc.). Pre-fix the AI waited up to 2 seconds for the natural cadence, surfacing as a visible "dead frame" stutter.

Added an `ai_retry_at: Option<Instant>` field on `CellEntity` and a `npc_ai_retry_sweep` function called every AoI tick (100ms) from the message loop. On launch failure, `npc_ai_fight` sets `ai_retry_at = Some(now + 500ms)`; the sweep picks it up on the next tick whose `Instant::now()` passes the deadline. Worst-case retry latency: **500–600ms** (vs 2s pre-fix). Mirrors `Atrea.addTimer(t + 0.5, doAiAction)` from `python/cell/SGWMob.py`.

The sweep iterates only NPCs with `ai_retry_at <= now` AND `AiState::Fighting`. Idle / Leashing entries with stale `ai_retry_at` are misuse and the natural-cadence tick handles them anyway. The slot is cleared **before** running the fight pass so a chained launch failure can reschedule cleanly without racing the iteration.

## Out of scope (per issue)

The `NPC_ATTACK_RANGE` constant stays as a fallback sentinel; flat `LEASH_DISTANCE = 50.0` is untouched.

## Test plan (all 6 new tests in `cell::service::tests::npc_ai`)

- [x] `npc_ai_fight_with_short_max_range_holds_fire_when_target_outside` — `max_range = 15`, target at distance 20: cooldown NOT started AND `ai_retry_at` is None. **Pre-fix would have entered the fire path and scheduled a retry — that's the canary.**
- [x] `npc_ai_fight_with_short_max_range_fires_when_target_inside` — same ability, target at distance 14: cooldown started (fire path ran end-to-end).
- [x] `npc_ai_fight_max_range_zero_falls_back_to_npc_attack_range` — `max_range = 0` sentinel, target at distance 25: cooldown started. **Regression guard for legacy NPCs.**
- [x] `npc_ai_fight_target_inside_min_range_schedules_backup_waypoint` — `min_range = 5`, target at origin, NPC at (3,0,0): exactly one waypoint, distance from target >= 6.0, +x direction.
- [x] `npc_ai_fight_schedules_retry_on_handle_use_ability_failure` — `ai_retry_at` lands in `[now + 400ms, now + 600ms]`.
- [x] `npc_ai_retry_sweep_processes_due_npc_and_clears_slot` — backdated deadline gets cleared after sweep.
- [x] All 18 pre-existing `npc_ai` tests continue to pass.
- [ ] Live test (Castle Cellblock): the `Cellblock Guard` (Pistol Shot 592) fires from the same distance as before — no regression in baseline behavior.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace` (excluding GUI apps) `-D warnings` clean
- `cargo nextest --profile=ci --workspace` (excluding GUI apps): **1555 passed, 0 failed, 1 skipped** (was 1549; +6 new guards. Wireclient pcap fixture skip is expected without `debug/`.)
- `cargo test --doc -p cimmeria-commands`: 4 passed
- `tools/lint-figure-style.sh` + `tools/check-figure-sources.sh`: clean

### DIFF
--- a/crates/entity/src/cell_entity/mod.rs
+++ b/crates/entity/src/cell_entity/mod.rs
@@ -397,6 +397,19 @@ pub struct CellEntity {
     pub spawn_position: Option<Vector3>,
     /// Ticks until next AI action (count-down from ai tick interval).
     pub ai_cooldown_ticks: u32,
+    /// Deadline for the next AI fight tick on this NPC, in service-local
+    /// `Instant`. `Some(t)` means "run `npc_ai_fight` on this NPC the
+    /// moment `now >= t`, even if the natural 2-second AI cadence hasn't
+    /// fired yet." Cleared by the retry-sweep when it consumes the entry,
+    /// or set to `Some(now + 500ms)` by `npc_ai_fight` when
+    /// `handle_use_ability` returns false (target dead between range
+    /// check and launch, animation lock, etc.). Mirrors the
+    /// `Atrea.addTimer(t + 0.5, doAiAction)` pattern from the Python
+    /// fork: the original 2-second cadence sometimes leaves the NPC
+    /// visibly idle for a full tick after a launch failure; the retry
+    /// closes that "dead frame" gap. None = follow the natural cadence
+    /// only (default for healthy NPCs).
+    pub ai_retry_at: Option<std::time::Instant>,
     /// Navmesh path waypoints the NPC is currently following.
     /// Empty = not moving. Each tick pops the next waypoint off the front.
     /// Stored as `VecDeque` so per-tick `pop_front` is O(1) instead of the
@@ -593,6 +606,7 @@ impl CellEntity {
             threat_list: HashMap::new(),
             spawn_position: None,
             ai_cooldown_ticks: 0,
+            ai_retry_at: None,
             nav_path: VecDeque::new(),
             move_speed: 0.6, // ~0.6 world units per 100ms tick = 6 units/sec
             is_stationary: false,
--- a/crates/services/src/cell/service/message_loop.rs
+++ b/crates/services/src/cell/service/message_loop.rs
@@ -111,6 +111,16 @@ pub(super) async fn run_cell_loop(
                     super::npc_ai::npc_ai_tick(tx, &mut space_mgr).await;
                 }
 
+                // Retry sweep for NPCs with a pending `ai_retry_at`
+                // deadline — runs every AoI tick (100ms) so a
+                // `handle_use_ability` launch failure can drive a 500ms
+                // retry instead of waiting up to 2 seconds for the
+                // natural cadence. Healthy NPCs (no `ai_retry_at`) are
+                // skipped on the snapshot scan, so the per-tick cost
+                // here is just an `all_npc_entity_ids()` walk + the
+                // None-filter — negligible. See issue #329.
+                super::npc_ai::npc_ai_retry_sweep(tx, &mut space_mgr).await;
+
                 // Out-of-combat HP/focus regen — 1 Hz (every 10th 100ms tick).
                 // Cadence is wired here so the per-call delta in `regen_tick`
                 // can stay "points per second" without an internal time check.
--- a/crates/services/src/cell/service/message_loop.rs
+++ b/crates/services/src/cell/service/message_loop.rs
@@ -118,7 +118,8 @@ pub(super) async fn run_cell_loop(
                 // natural cadence. Healthy NPCs (no `ai_retry_at`) are
                 // skipped on the snapshot scan, so the per-tick cost
                 // here is just an `all_npc_entity_ids()` walk + the
-                // None-filter — negligible. See issue #329.
+                // pending-retry set membership — negligible. See
+                // `npc_ai::npc_ai_retry_sweep` for the rationale.
                 super::npc_ai::npc_ai_retry_sweep(tx, &mut space_mgr).await;
 
                 // Out-of-combat HP/focus regen — 1 Hz (every 10th 100ms tick).
--- a/crates/services/src/cell/service/npc_ai.rs
+++ b/crates/services/src/cell/service/npc_ai.rs
@@ -188,9 +188,32 @@ async fn npc_ai_fight(npc_id: u32, tx: &mpsc::Sender<CellToBaseMsg>, space_mgr: 
         }
     }
 
-    // Range check: don't attack until target is within weapon range
+    // Pick the ability up front so the range check can gate on the
+    // ability's own `min_range` / `max_range` instead of a flat
+    // server-wide constant. `choose_npc_ability` returns:
+    //   - `Some(NPC_DEFAULT_ABILITY)` when the NPC has no known abilities
+    //     (misconfigured template — explicit fallback per the selector's
+    //     "don't wedge silently" rule).
+    //   - `Some(id)` for the first non-cooling known ability.
+    //   - `None` when every known ability is on cooldown.
+    //
+    // In the `None` case we keep the range/LOS logic running against the
+    // server-wide fallback so the NPC still walks toward / tracks the
+    // target while waiting for an off-cooldown ability — same effective
+    // behavior as the pre-issue-329 flat-30.0 code path.
+    let chosen_ability = choose_npc_ability(npc_id, space_mgr);
+    let (max_range, min_range) =
+        ability_ranges(chosen_ability, space_mgr, combat::NPC_ATTACK_RANGE);
+
+    // Range check: don't attack until target is within the chosen
+    // ability's `max_range` (or `NPC_ATTACK_RANGE` if the def is missing
+    // or carries the `0` sentinel meaning "use server default"). Pinned
+    // by issue #329: prior code used the flat constant and ignored
+    // per-ability `max_range`, which produced "NPC walks into firing
+    // distance but stands there" for any ability with `max_range < 30`
+    // (e.g., a grenade at `max_range = 15`).
     let dist_to_target = npc_pos.distance_to(&target_pos);
-    let in_range = dist_to_target <= combat::NPC_ATTACK_RANGE;
+    let in_range = dist_to_target <= max_range;
     let has_los = space_mgr.has_line_of_sight(npc_id, target_id);
 
     // Out of range OR occluded — keep pathfinding so the NPC can reposition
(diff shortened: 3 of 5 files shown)


MyPalClara (commented):

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `ability_ranges` fallback on `None` | **When `chosen_ability` is `None` (all abilities cooling), the function returns `(npc_attack_range, 0.0)`.** This is correct for the "walk toward target" gate, but the min_range of 0.0 means the backup-waypoint logic is skipped. **If a future ability has `min_range > 0` AND the NPC's other abilities are cooling, the NPC would walk into the dead zone and stall.** Probably fine — the all-cooling case is transient — but document the interaction. |
| **G2** | `compute_backup_waypoint` Y-axis handling | **The backup waypoint preserves the Y-axis (vertical) component.** If the NPC is on a slope and the target is below, the backup point might be in mid-air or underground. **The navmesh should clamp this, but the raw waypoint could be invalid.** Consider projecting to the navmesh or documenting that the movement system handles this. |
| **G3** | `npc_ai_retry_sweep` iteration order | **The `due` vector is collected then iterated.** If two NPCs both have past-due deadlines, they're processed in `all_npc_entity_ids()` order, not deadline order. **Probably fine — both are past-due anyway — but if deadline ordering ever matters, this would need a sort.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `AI_LAUNCH_FAILURE_RETRY_DELAY` hardcoded | **The 500ms constant mirrors the Python `Atrea.addTimer(t + 0.5, ...)`.** But if the Python reference changes, this constant won't track. **Consider adding a comment citing the exact Python file/line.** |
| **G5** | `ai_retry_at` on non-Fighting NPCs | **The sweep filters to `AiState::Fighting`, but the field can be set on any NPC.** If an NPC transitions from Fighting → Idle while `ai_retry_at` is set, the stale deadline persists until the NPC re-enters Fighting. **Probably harmless — the sweep ignores it — but consider clearing `ai_retry_at` on state transitions.** |
| **G6** | `min_range + 1.0` margin | **The +1.0 margin is documented as "floating-point jitter prevention."** But 1.0 world unit might be too large for some abilities (e.g., a melee ability with `min_range = 1`). **Consider making the margin proportional to `min_range` (e.g., `min_range * 1.1`) or documenting that 1.0 is the floor.** |

| **G7** | No test for `ability_ranges` with missing def | **The function falls back to `npc_attack_range` when the def is missing, but there's no test that exercises this path explicitly.** The `max_range_zero_falls_back` test covers the `0` sentinel, not the missing-def case. |
| **G8** | No test for `compute_backup_waypoint` degenerate case | **The function returns `None` when NPC and target are co-located (`dist < EPSILON`).** There's no test that verifies this edge case. |
| **G9** | No test for stationary NPC skipping backup | **The PR body says "Stationary NPCs skip the backup" but I don't see a test that verifies `is_stationary = true` prevents the backup waypoint.** |

| **G10** | `npc_ai.rs` module doc doesn't mention retry sweep | **The module-level doc comment should mention the retry sweep as part of the AI tick architecture.** |
| **G11** | No mention of the Python file/line for the 500ms constant | **The PR body cites `python/cell/SGWMob.py` but not the exact line number.** Future maintainers would benefit from a precise citation. |

---

## Summary

| Correctness issues | 3 | **Low** — G1-G3 are edge cases |
| Potential issues | 3 | **Low** — G4-G6 are maintainability |
| Missing test coverage | 3 | **Low** — G7-G9 are nice-to-haves |
| Documentation | 2 | **Low** — G10-G11 are polish |

**Total: 11 gaps, all low severity.**

---

## Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟢 Quick wins | G4, G11 — add Python line citation | ~5 min |
| 🟡 Consider | G9 — stationary NPC test | ~15 min |
| ⚪ Follow-up | G5 — clear `ai_retry_at` on state transition | ~30 min |

---

## Verdict

**Issue #329 is fully addressed.** The three fixes are correct:
1. Per-ability `max_range` gates the fire path ✅
2. Min-range backup waypoint steps the NPC out of the dead zone ✅
3. 500ms retry on launch failure closes the "dead frame" gap ✅

The 6 new tests cover the issue's test plan. The 11 gaps are all low severity edge cases and polish.

**Ship it.** 🎯

PR #392, opened 2026-05-25
fix(mercury): per-entity idbase for entity-method sub-slot threshold (closes #315)

### LINKED ISSUE #315: Entity-method sub-slot threshold hardcoded at 61 — should be per-entity-computed
## Plain English

When the server sends a method call to the client, the wire format has two encodings: a short one (1 byte) for low method numbers, and a long one (2 bytes plus a marker prefix) for high ones. **Where the boundary between "low" and "high" falls depends on how many exposed methods that entity type has** — different entity types have different boundaries.

Our code hardcodes the boundary at `61` in `append_entity_method`. That's correct for SGWPlayer (the only entity type we currently send method calls to), because SGWPlayer has 157 exposed methods which works out to a boundary of exactly 61. The moment we start sending method calls to other entity types — NPCs, mobs, pets, anything smaller than ~62 exposed methods — the boundary will be `62`, not `61`, and our hardcoded value will silently send method index 61 through the wrong encoding path. The client will decode it as a different method than the one we meant.

The bug is invisible today because every method call we emit happens to target the player. It becomes visible the first time we emit a method call to a non-player entity at a method index near the boundary.

Bonus: the docstring on the function says "we use the simpler boundary at 128" but the code uses 61. Internal contradiction — probably an unfinished refactor.

## Spec evidence

Per the bible chapter [`spec.protocol.entity-property-sync`](docs/drafts/spec/entity-property-sync.md) §1.4 and **R2 + S1**, the threshold (`idBase`) is **per-entity**, computed by:

```
idBase = 0x3E - (nExposedCount + 0xC0) / 0xFF
```

For SGWPlayer (`nExposedCount = 157`), the formula gives `idBase = 61` — so the hardcoded value works for SGWPlayer by happy accident. For any entity with `nExposedCount ≤ 62`, the formula gives `idBase = 62`. The chapter calls this out explicitly: §1.17 S1 ("Sub-slot threshold is method-count-dependent, not a constant `0x3E`") and §1.18 R2 ("Compute the sub-slot threshold `idBase` from each entity's exposed-method count").

Binary anchors:
- `EntityDescription_AssignClientMethodIds @ ghidra://SGW.exe@0x01590df0` — formula source
- `EntityDescription_DecodeClientMethodId @ ghidra://SGW.exe@0x01590ee0` — inverse decoder using the same per-entity `idBase`
- Audit Appendix C.6 wire-confirms `idBase=61` for SGWPlayer

## Code site

[`crates/services/src/mercury/mod.rs:226-244`](crates/services/src/mercury/mod.rs#L226) `append_entity_method`:

```rust
if method_index >= 61 {
    // Extended encoding: marker 0xBD
    body.push(0xBD);
    // ...
} else {
    // Direct encoding: msg_id = index | 0x80
    body.push((method_index as u8) | 0x80);
    // ...
}
```

And the contradictory docstring at [`crates/services/src/mercury/mod.rs:220-225`](crates/services/src/mercury/mod.rs#L220-L225):

```rust
/// - **Extended** (method_index >= 128): `[0xBD: u8]...`
/// ...
/// We use the simpler boundary at 128.
```

## Proposed fix

Two-part:

1. **Pass per-entity exposed-method count into `append_entity_method`** (or precompute `idBase` per entity type and look it up at call time). Either form satisfies R2. The exposed-method count is available from the entity registry (`crates/defs/src/registry.rs`).
2. **Fix the docstring** to match the actual behavior — drop the "boundary at 128" / "client accepts through 127" claims, document the per-entity formula with the §1.4 Ghidra anchor.

## Test plan

- [ ] Unit test: `idBase` computation matches the formula for SGWPlayer (157), and at least one smaller entity where `nExposed ≤ 62` so `idBase = 62`.
- [ ] Wire-format test: encoding method index 60 always uses the direct path (well under any `idBase`).
- [ ] Wire-format test: encoding method index 156 for SGWPlayer (`idBase=61`) uses extended encoding with `(156 - 61) = 95` as the sub-index byte (matches §1.4 worked example).
- [ ] Wire-format regression: `setupWorldParameters` (index 122 for SGWPlayer) still encodes as `[0xBD][word_len][entity_id][(122-61)=61]...` (matches audit C.6 wire-capture).

## Related

- Chapter [`spec.protocol.entity-property-sync`](docs/drafts/spec/entity-property-sync.md) §1.4, §1.17 S1, §1.18 R2
- Audit [Appendix C.6 (V5 cell-method wire byte mask)](docs/audits/entity-property-sync-section2-audit-2026-05-16.md) — wire-confirms `idBase=61` for SGWPlayer
- Surfaced during entity-property-sync spec ↔ Rust gap analysis (PR #305)

Cadacious:
Closes #315

## Summary

The entity-method sub-slot threshold (between direct `0x80..0xBC` and extended `0xBD + sub_index` encoding) is **per-entity-type**, computed from each type's exposed-client-method count. The previous code hardcoded `61` and the docstring contradicted the implementation (\"we use the simpler boundary at 128\").

Latent today because every server→client method call currently targets the player, and SGWPlayer's 157 exposed methods happen to plug into the formula and yield `61` — so the hardcoded value was correct by accident. It becomes a wire bug the first time we emit a method index near the boundary on a non-player entity.

Formula (per audit §1.4, Ghidra [`EntityDescription_AssignClientMethodIds @ 0x01590df0`](docs/audits/entity-property-sync-section2-audit-2026-05-16.md#c2)):

```text
idBase = 0x3E - (nExposedCount + 0xC0) / 0xFF
```

| Entity | nExposed | idbase | Notes |
|---|---:|---:|---|
| SGWPlayer | 157 | **61** | Hardcoded value worked by accident. |
| SGWMob / SGWPet / SGWBeing-derived NPCs | ≤62 | **62** | All current NPC types. |

## Implementation

- **`crates/mercury/src/channel_bundle.rs`**: removed `EXTENDED_ENCODING_THRESHOLD`. Added:
  - `IDBASE_SGW_PLAYER: u8 = 61`
  - `IDBASE_NPC_DEFAULT: u8 = 62`
  - `const fn idbase_from_exposed_method_count(n: usize) -> u8`
  - `ChannelBundle::append_entity_method` now takes `idbase: u8`.

- **`crates/services/src/mercury/mod.rs::append_entity_method`**: takes `idbase: u8`. Docstring rewritten — old \"boundary at 128\" / \"client accepts through 127\" claims removed; new docstring documents the per-entity formula with bible/Ghidra anchors.

- **`crates/services/src/mercury/aoi/method.rs`**: `build_entity_method_packet` takes `idbase: u8`. Added `build_player_entity_method_packet` wrapper for the common SGWPlayer-target case.

- **AoI cascade (`compose_create_entity_cascade_body`)**: picks idbase via `cascade_idbase(npc_data)` — `IDBASE_NPC_DEFAULT` for NPC cascades, `IDBASE_SGW_PLAYER` for player-ghost cascades.

- **Witness broadcast (`witness_entity_method` in `cell_dispatch/aoi.rs`)**: uses `IDBASE_NPC_DEFAULT` since the target is a ghost entity in the witness's AoI (not the witness itself, who happens to be a player).

- ~20 player-targeted callers updated. Direct `append_entity_method` callers pass `IDBASE_SGW_PLAYER`; legacy `build_entity_method_packet` callers migrate to `build_player_entity_method_packet` — same wire bytes, but the function name now documents which idbase governs.

## Tests (per issue plan)

**8 new tests** in [`crates/mercury/src/channel_bundle.rs`](crates/mercury/src/channel_bundle.rs) covering the issue's test plan:

Formula:
- `idbase_formula_sgw_player_157_methods` — 157 → 61 (issue example).
- `idbase_formula_small_method_count_is_62` — every n ∈ 0..=62 → 62.
- `idbase_formula_drops_to_61_at_63_methods` — boundary jump 62→63.
- `idbase_formula_staircase_at_each_255_step` — 318→60, 573→59 (formula corners).

Encoder:
- `encoder_method_60_direct_for_both_idbases` — sanity (no wire change at low indices).
- `encoder_method_61_extended_for_sgw_player` — sub_index = 0 (worked-example boundary).
- `encoder_method_61_direct_for_npc_default_idbase` — the per-entity divergence the issue makes visible (msg_id `0xBD` here is a direct byte, not an extended marker).
- `encoder_method_156_extended_sub_index_95_for_sgw_player` — issue's worked example.
- `encoder_setup_world_parameters_122_extended_sub_index_61` — audit Appendix C.6 wire-capture witness.

**Existing wire-format byte-equivalence tests** updated to pass the new `idbase` parameter — covers BeingAppearance / onEntityTint / setup_world_parameters byte layouts. 1569 workspace tests pass.

## Wire bytes — what does and doesn't change

For SGWPlayer-targeted methods, **no wire bytes change** — `IDBASE_SGW_PLAYER` is `61`, the same value the hardcode used. The audit Appendix C.6 wire capture (35 `0xBD` sub-slot sentinels for SGWPlayer methods, named methods at the expected sub-indices) continues to pass byte-for-byte.

For NPC-targeted methods, no current emit crosses index 61, so no current wire byte changes either. The fix is forward-looking: when we start emitting NPC method indices ≥61, they will now correctly direct-encode with the NPC's idbase=62 instead of mis-encoding as an extended marker.

### DIFF
--- a/crates/mercury/src/channel_bundle.rs
+++ b/crates/mercury/src/channel_bundle.rs
@@ -79,26 +79,70 @@
 
 use crate::packet::FRAGMENT_BODY_SIZE;
 
-/// Method-index boundary between direct and extended entity-method
-/// encoding. Indices `< 61` use direct encoding (msg_id = index | 0x80),
-/// indices `>= 61` use extended encoding (msg_id = 0xBD, sub_index byte
-/// after entity_id).
+/// Per-entity sub-slot threshold (`idBase`) for SGWPlayer.
 ///
-/// **Single source of truth** for the entity-method encoding boundary —
-/// both `ChannelBundle::append_entity_method` and the services-layer
-/// `crates/services/src/mercury/mod.rs::append_entity_method` import
-/// this constant so they cannot silently drift. A divergence between
-/// the two encoders would split the wire format between bundle-migrated
-/// and un-migrated call sites (pinned by the byte-equivalence tests
-/// `compose_create_entity_*_body_matches_*` in
-/// `crates/services/src/mercury/aoi/tests.rs`, but the shared constant
-/// removes the failure mode entirely).
-pub const EXTENDED_ENCODING_THRESHOLD: u16 = 61;
+/// SGWPlayer exposes 157 client methods, which plugged into the formula
+/// [`idbase_from_exposed_method_count`] gives `idBase = 61`. Every
+/// server→player method call uses this value: indices `0..61` direct-
+/// encode (msg_id = `index | 0x80`), indices `61..157` extended-encode
+/// (marker `0xBD`, then sub_index byte = `index - 61`).
+///
+/// **The threshold is per-entity-type, NOT a global constant.** Different
+/// entity types have different exposed-method counts and therefore
+/// different idbases. Pass the correct idbase for the target entity to
+/// every encoder call. For entities with `nExposedCount <= 62`, idbase
+/// is `62` — see [`idbase_from_exposed_method_count`] for the formula
+/// (`EntityDescription_AssignClientMethodIds @
+/// ghidra://SGW.exe@0x01590df0`).
+///
+/// Spec: [docs/drafts/spec/entity-property-sync.md §1.4][1] §1.17 S1,
+/// §1.18 R2.
+///
+/// [1]: ../../../../docs/drafts/spec/entity-property-sync.md
+pub const IDBASE_SGW_PLAYER: u8 = 61;
 
-/// Extended-encoding marker byte. Cannot be used as a direct msg_id.
-/// Shared with `services` for the same anti-drift reason as
-/// [`EXTENDED_ENCODING_THRESHOLD`].
+/// Per-entity sub-slot threshold for SGW non-player entities (SGWMob,
+/// SGWPet, SGWBeing-derived NPCs, etc.). Every such entity type in the
+/// current SGW schema exposes ≤62 client methods, so the formula
+/// [`idbase_from_exposed_method_count`] returns `62` (the maximum). Use
+/// this for any method call targeting a non-player entity until a
+/// per-class lookup is wired up.
+pub const IDBASE_NPC_DEFAULT: u8 = 62;
+
+/// Extended-encoding marker byte. Cannot be used as a direct msg_id for
+/// SGWPlayer (whose `idbase = 61`) because the direct range stops at
+/// `0xBC = 60 | 0x80`. For entities with a higher idbase the byte `0xBD`
+/// can mean "direct method 61" rather than "extended marker" — the
+/// client decoder distinguishes by knowing the receiving entity's idbase.
 pub const EXTENDED_ENCODING_MARKER: u8 = 0xBD;
+
+/// Compute the per-entity sub-slot threshold (`idBase`) from the entity
+/// type's exposed-client-method count, per `EntityDescription_AssignClient
+/// MethodIds @ ghidra://SGW.exe@0x01590df0`:
+///
+/// ```text
+///   idBase = 0x3E - (nExposedCount + 0xC0) / 0xFF
+/// ```
+///
+/// Method indices `0..idbase` use single-byte direct encoding (wire byte
+/// `(index | 0x80)`). Method indices `idbase..` use the two-byte extended
+/// encoding (marker `0xBD` + sub_index byte = `index - idbase`).
+///
+/// - `nExposedCount = 157` (SGWPlayer) → `idBase = 61`. The constant
+///   [`IDBASE_SGW_PLAYER`] pre-computes this for the common case.
+/// - `nExposedCount <= 62` → `idBase = 62`. The whole namespace fits in
+///   single-byte direct encoding; the wildcard `0xBD` byte is then a
+///   valid direct msg_id for method 61 on that entity rather than the
+///   extended marker.
+/// - `nExposedCount = 318` → `idBase = 60`. Method 60 starts extended-
+///   encoding territory.
+///
+/// Spec: `docs/drafts/spec/entity-property-sync.md` §1.4, §1.17 S1,
+/// §1.18 R2.
+pub const fn idbase_from_exposed_method_count(n_exposed_count: usize) -> u8 {
+    let i_var2 = (n_exposed_count + 0xC0) / 0xFF;
+    (0x3E - i_var2) as u8
+}
 
 /// Accumulator for one logical bundle of application-level messages.
 ///
@@ -159,6 +203,13 @@ impl ChannelBundle {
 
     /// Append a server→client entity-method call to the bundle body.
     ///
+    /// `idbase` is the per-entity-type sub-slot threshold for the target
+    /// entity — see [`idbase_from_exposed_method_count`]. For methods
+    /// targeting SGWPlayer pass [`IDBASE_SGW_PLAYER`] (`61`). For entities
+    /// with ≤62 exposed methods, pass `62`. The threshold is **not** a
+    /// global constant; encoding with the wrong idbase produces a wire
+    /// byte the client decodes as a different method.
+    ///
     /// Wire format matches
     /// [`crates/services/src/mercury/mod.rs`]'s `append_entity_method`
     /// byte-for-byte — see the module doc for the encoding.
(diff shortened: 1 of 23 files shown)


MyPalClara (commented):

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `idbase_from_exposed_method_count` integer overflow | **The formula `(n_exposed_count + 0xC0) / 0xFF` can overflow if `n_exposed_count` is very large.** For `n > 15_807`, the result would be negative after the `0x3E -` subtraction. **No current entity type has anywhere near that many methods, but the function is `pub const` and could be misused.** Consider adding a `debug_assert!(n_exposed_count < 16_000)` or documenting the valid input range. |
| **G2** | `cascade_idbase` hardcoded to `IDBASE_NPC_DEFAULT` | **The PR body says "picks idbase via `cascade_idbase(npc_data)`" but the implementation always returns `IDBASE_NPC_DEFAULT` for NPCs.** If a future NPC type has >62 exposed methods, this would silently mis-encode. **The current schema is safe, but consider a TODO comment for when per-class lookup is wired.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `IDBASE_NPC_DEFAULT` assumes ≤62 methods | **The constant is documented as "every NPC type in the current schema exposes ≤62 client methods."** If a new NPC type is added with >62 methods, this constant becomes wrong. **Consider adding a compile-time or startup assertion that validates the assumption against the entity registry.** |
| **G4** | `build_player_entity_method_packet` wrapper | **The new wrapper is convenient but adds indirection.** Callers must now choose between `build_entity_method_packet` (explicit idbase) and `build_player_entity_method_packet` (implicit SGWPlayer idbase). **The naming is clear, but consider whether the explicit-idbase version should be `pub(crate)` to prevent misuse.** |
| **G5** | ~20 callers updated | **The PR body says "~20 player-targeted callers updated."** A missed caller would silently use the wrong idbase. **The byte-equivalence tests catch this for the tested paths, but consider a grep for `build_entity_method_packet` to verify all call sites are migrated.** |

| **G6** | No test for `cascade_idbase` | **The function is called in `compose_create_entity_cascade_body` but there's no unit test that verifies it returns the correct idbase for NPC vs player cascades.** |
| **G7** | No test for `witness_entity_method` idbase | **The PR body says witness broadcasts use `IDBASE_NPC_DEFAULT`, but there's no test that verifies the wire bytes.** The existing tests cover player-targeted methods. |
| **G8** | No test for formula edge case at n=0 | **`idbase_formula_small_method_count_is_62` covers `0..=62`, but the loop starts at 0.** An entity with 0 exposed methods is degenerate — consider whether this should be an error or documented as valid. |

| **G9** | ADR/spec cross-reference in code | **The docstrings cite `docs/drafts/spec/entity-property-sync.md §1.4` but the relative path `../../../../docs/...` may break if the file moves.** Consider using an absolute path from repo root or a stable anchor. |
| **G10** | No mention of the "worked by accident" history in code | **The PR body explains that the hardcoded `61` worked by accident for SGWPlayer.** This context is valuable — consider adding a one-liner comment in the code so future readers understand why the constant existed. |

---

## Summary

| Correctness issues | 2 | **Low** — G1-G2 are edge cases |
| Potential issues | 3 | **Low** — G3-G5 are maintainability |
| Missing test coverage | 3 | **Low** — G6-G8 are nice-to-haves |
| Documentation | 2 | **Low** — G9-G10 are polish |

**Total: 10 gaps, all low severity.**

---

## Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟢 Quick wins | G10 — add "worked by accident" comment | ~5 min |
| 🟡 Consider | G5 — grep for missed callers | ~10 min |
| ⚪ Follow-up | G3 — startup assertion for NPC method counts | ~30 min |

---

## Verdict

**Issue #315 is fully addressed.** The per-entity idbase formula is correct:
- `idbase_from_exposed_method_count` matches the Ghidra disassembly ✅
- `IDBASE_SGW_PLAYER = 61` for 157 methods ✅
- `IDBASE_NPC_DEFAULT = 62` for ≤62 methods ✅
- All ~20 callers updated to pass the correct idbase ✅

The 9 new tests cover the formula corners and encoder divergence. The wire bytes are unchanged for SGWPlayer (the common case) — this is a forward-looking fix for when NPC method indices ≥61 are emitted.

**Ship it.** 🎯

Cadacious:
Addressed in 0ebd801d.

**Copilot** (3 inline):
- ✅ **Overflow guard** on `idbase_from_exposed_method_count`. The formula's domain is `0..=15_807`; above that, `(n + 0xC0) / 0xFF` exceeds `0x3E` and the subtraction would underflow. New branch saturates to `0` instead of wrapping. Saturation is defensive — `idbase = 0` implies an unencodable entity anyway (sub_index byte caps at 255). New regression test `idbase_formula_saturates_to_zero_on_overflow` exercises `15_807`, `16_000`, `usize::MAX / 2`, `usize::MAX - 0xC0`.
- ✅ **Services-layer `append_entity_method` now uses explicit `u8::try_from` / `u16::try_from` with `.expect()` messages**, byte-for-byte aligned with `ChannelBundle::append_entity_method`. The two encoders' field-width contracts are now consistent — prevents silent narrowing-cast corruption on out-of-range inputs.
- ✅ **Stripped `issue #315` references from source comments** per `.github/copilot-instructions.md:28-29,50`. Same rule that bit me on PR #390 — PR/issue numbers in source rot, provenance belongs in PR body or audit doc.

**Clara**:
- ✅ **G2** (cascade_idbase TODO): expanded the docstring explaining when the `IDBASE_NPC_DEFAULT` fallback becomes incorrect (any NPC type gaining >62 exposed methods) and the long-term fix (thread `class_id` through `NpcAoIData` and look up precomputed idbase per class).
- ✅ **G5** (caller migration audit): `grep -rn "build_entity_method_packet("` confirms only the 3 intentional remaining calls — `cell_dispatch/aoi.rs` (explicit `IDBASE_SGW_PLAYER` for player target, explicit `IDBASE_NPC_DEFAULT` for ghost target) and the wrapper definition itself. No missed migrations.
- ✅ **G6** (cascade_idbase tests): added unit tests pinning both branches — `Some(npc_data) → IDBASE_NPC_DEFAULT` and `None → IDBASE_SGW_PLAYER`.

**Held with rationale**:
- **G1** subsumed by Copilot's identical finding — both addressed by the saturation guard.
- **G3** (compile-time/startup assertion that all NPC types have ≤62 methods): out-of-scope for this PR — would require parsing the entity registry at startup. The docstring on `cascade_idbase` now explicitly calls out when this fallback breaks and what the long-term fix is.
- **G4** (make explicit-idbase API `pub(crate)`): both APIs intentionally `pub` — `build_entity_method_packet` is used outside the crate when callers know their target's idbase, `build_player_entity_method_packet` is the convenience for the common case. Naming makes intent clear.
- **G7** (witness_entity_method wire-byte test): existing aoi/tests.rs byte-equivalence guard covers the same path via `build_entity_method_packet` with explicit `idbase` — adding a separate NPC-target test would duplicate coverage.
- **G8/G9** cosmetic.
- **G10** would re-introduce issue-number references the Copilot inline ruled out.

**Codecov**: 83% patch coverage — uncovered lines are in caller files (mail/, vendor/, inventory/) that lacked tests pre-existing this PR. The change at each site is a mechanical `_packet(...)` → `_player_entity_method_packet(...)` rename or explicit idbase add. The encoder logic (where bugs would actually live) is covered by the 12 new wire-format / formula tests in `cimmeria-mercury`.

Workspace status: 1572 tests pass, clippy `--all-targets -- -D warnings` clean.

PR #394, opened 2026-05-25
feat(cell): wire autoReload + reloadOnActivate client options

Cadacious:
## Summary

- Wires up the two client-side checkboxes "Auto reload weapon" and "Reload on weapon activation" — they could be toggled in the options menu but did nothing today.
- Implements the previously-stubbed `updateSystemOptions` cell method (player method 93, `ARRAY <of> NameValuePair` payload) so the server actually sees the player's selections.
- Triggers `handle_reload` automatically when (a) `autoReload` is set and a fire drains the clip to zero, or (b) `reloadOnActivate` is set and the player switches to / equips a weapon with a partial clip.

## Three missing pieces (now fixed)

| Gap | Before | After |
|---|---|---|
| `updateSystemOptions` (player method 93) | `tracing::info!("UNIMPLEMENTED")` and dropped on the floor | Parses the `ARRAY <of> NameValuePair` blob into `SystemOptions`, logs unknown keys at `debug` |
| Auto-reload on empty clip | Nothing — manual `R` was the only path | After `apply_damage_to_target` at both exit points of `handle_use_ability`, checks the gate and calls `handle_reload` if `active_ammo() == 0` |
| Reload on weapon activate | Nothing | Hooked into `handle_request_active_slot_change` (F1-F4 swap) and `handle_update_bandolier_item` (in-game equip from inventory); does NOT trigger on unholster because Phase A of `handle_reload` IS the draw animation and would loop |

## Architecture

- New `cimmeria_entity::cell_entity::SystemOptions` block on `CellEntity`. Defaults match `SGWGame/Content/XML/SystemOptions.xml` (`autoReload=true`, `reloadOnActivate=false`).
- New `parse_name_value_pairs` in `cell::cell_methods::player::world` decodes the BigWorld wire format (count u32 + per-pair WSTRING + WSTRING), bounded at 256 entries to defend against malformed payloads.
- Two thin trigger helpers, `maybe_trigger_auto_reload` (in `use_ability/mod.rs`) and `maybe_trigger_reload_on_activate` (in `cell_methods/player/world/mod.rs`), centralise the per-option gates so we don't sprinkle option checks across N call sites.

## What's NOT in scope here

- **Persistence to the DB.** The XML marks both options `serverSynch='true'` and the original game stored them account-side. Cimmeria's current options storage is in-memory only; persistence is a follow-up PR (needs a `sgw_player.system_options` JSONB column or equivalent + hydrate-on-login).
- **The other ~100 options in SystemOptions.xml.** Extending is mechanical — add a field on `SystemOptions` and a match arm in `apply()` — but not load-bearing for this PR's user-visible bug.

## Test plan

- [x] 6 `SystemOptions::apply` unit tests (defaults match XML, both known options, unknown option, alternate truthy tokens, empty string)
- [x] 4 `parse_name_value_pairs` byte-exact tests (two options, empty array, truncated count, implausible count)
- [x] 3 `updateSystemOptions` dispatch tests (both options apply, unknowns don't clobber knowns, garbage payload leaves state untouched)
- [x] 5 `maybe_trigger_reload_on_activate` guards (triggers when option+partial-clip, skips when option off, skips when clip full, skips during in-flight reload, skips for non-players)
- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo nextest run --profile=ci --workspace`: **1578 tests pass, 1 skipped, 0 failed**

## Files touched

| File | Change |
|---|---|
| `crates/entity/src/cell_entity/system_options.rs` | NEW — `SystemOptions` struct + `apply()` + 6 unit tests |
| `crates/entity/src/cell_entity/mod.rs` | + module decl, + field on `CellEntity`, + default in ctor |
| `crates/services/src/cell/cell_methods/player/world/mod.rs` | UNIMPLEMENTED stub → real handler + `parse_name_value_pairs` + `maybe_trigger_reload_on_activate` |
| `crates/services/src/cell/cell_methods/player/world/system_options_tests.rs` | NEW — 12 regression guards |
| `crates/services/src/cell/abilities/use_ability/mod.rs` | + `maybe_trigger_auto_reload` called at both fire-success exits |
| `crates/services/src/cell/cell_methods/inventory/bandolier.rs` | + reload-on-activate hook at end of F1-F4 swap |
| `crates/services/src/cell/service/base_messages/bandolier.rs` | + reload-on-activate hook at end of in-game equip |

## Summary by CodeRabbit

* **New Features**
  * Automatic reload when ammo depletes: weapons now automatically reload when active ammo is consumed (configurable per-player)
  * Automatic reload on weapon switch: weapons now automatically reload when switching active bandolier slots (configurable per-player)
  * Both settings persist across sessions and can be individually enabled or disabled by players

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/394?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/entity/src/cell_entity/mod.rs
+++ b/crates/entity/src/cell_entity/mod.rs
@@ -499,17 +499,19 @@ pub struct CellEntity {
 
     /// Per-session client option state populated by `updateSystemOptions`
     /// (player method index 93). Defaults to `SystemOptions::default()` on
-    /// entity construction, then overwritten when the client pushes its
-    /// options blob shortly after login (and again whenever the user
-    /// changes an option in the in-game menu).
+    /// entity construction, then overwritten by either of two paths:
     ///
-    /// Currently honours `autoReload` and `reloadOnActivate`; extending to
-    /// the full ~100 options in `SGWGame/Content/XML/SystemOptions.xml`
-    /// is mechanical — see [`SystemOptions::apply`].
+    /// - **Hydrate-on-login** — `InitPlayerState` carries the persisted
+    ///   values loaded from `sgw_player.auto_reload` and
+    ///   `sgw_player.reload_on_activate`.
+    /// - **Live update** — the client's `updateSystemOptions` push, which
+    ///   also fires a `CellToBaseMsg::SystemOptionsUpdate` so the row is
+    ///   written back to the DB on every toggle.
     ///
-    /// Not yet persisted to the DB across logins. The XML marks these
-    /// options `serverSynch='true'` and the original game stored them
-    /// account-side; that hydrate path is a follow-up.
+    /// Currently honours `autoReload` and `reloadOnActivate` — the only
+    /// two options the 2009 XML marks `serverSynch='true'`. Extending to
+    /// new server-synced options is mechanical — see
+    /// [`SystemOptions::apply`].
     pub system_options: SystemOptions,
 }
 
--- a/crates/entity/src/cell_entity/mod.rs
+++ b/crates/entity/src/cell_entity/mod.rs
@@ -22,6 +22,9 @@ use crate::stats::StatList;
 
 mod bandolier;
 mod state_flags;
+mod system_options;
+
+pub use system_options::SystemOptions;
 
 #[cfg(test)]
 mod tests;
@@ -493,6 +496,21 @@ pub struct CellEntity {
     /// transient. The chain that reaches the threshold also explicitly
     /// resets the counter via `Action::ResetCounter`.
     pub counters: HashMap<String, i32>,
+
+    /// Per-session client option state populated by `updateSystemOptions`
+    /// (player method index 93). Defaults to `SystemOptions::default()` on
+    /// entity construction, then overwritten when the client pushes its
+    /// options blob shortly after login (and again whenever the user
+    /// changes an option in the in-game menu).
+    ///
+    /// Currently honours `autoReload` and `reloadOnActivate`; extending to
+    /// the full ~100 options in `SGWGame/Content/XML/SystemOptions.xml`
+    /// is mechanical — see [`SystemOptions::apply`].
+    ///
+    /// Not yet persisted to the DB across logins. The XML marks these
+    /// options `serverSynch='true'` and the original game stored them
+    /// account-side; that hydrate path is a follow-up.
+    pub system_options: SystemOptions,
 }
 
 /// An item in a dead NPC's loot list, ready for display to players.
@@ -610,6 +628,7 @@ impl CellEntity {
             ring_source_id: None,
             destination_ring_id: None,
             counters: HashMap::new(),
+            system_options: SystemOptions::default(),
         }
     }
 
--- a/crates/entity/src/cell_entity/system_options.rs
+++ b/crates/entity/src/cell_entity/system_options.rs
@@ -1,0 +1,144 @@
+//! Per-session client option state — populated by the
+//! `updateSystemOptions` cell method (player method index 93, wire signature
+//! `ARRAY <of> NameValuePair`).
+//!
+//! Defaults mirror `SGWGame/Content/XML/SystemOptions.xml`:
+//!
+//! ```text
+//! <option name='autoReload'         defaultBool='true'  serverSynch='true' />
+//! <option name='reloadOnActivate'   defaultBool='false' serverSynch='true' />
+//! ```
+//!
+//! Persistence to the database is a follow-up; this struct lives on
+//! `CellEntity` and resets to defaults on every login. Server-synched
+//! semantics in the original game meant the option was stored against the
+//! account and pushed back at character-select; we will need a DB column
+//! plus a hydrate-on-login pass when we wire that. The behavioural change
+//! that makes the checkboxes *do anything* is the immediate gap.
+
+/// Single source of truth for the two reload-related client options.
+///
+/// Extending the struct to cover more of the ~100 options in
+/// `SystemOptions.xml` is mechanical — add a field, default it from the
+/// XML's `defaultBool`/`defaultInt`/`defaultFloat`/`defaultString`, and
+/// add a match arm in [`SystemOptions::apply`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SystemOptions {
+    /// `autoReload` — auto-reload weapon when the active clip hits zero.
+    /// XML default: `true`.
+    pub auto_reload: bool,
+
+    /// `reloadOnActivate` — reload the active weapon on bandolier
+    /// activate / weapon equip if the clip is below max.
+    /// XML default: `false`.
+    pub reload_on_activate: bool,
+}
+
+impl Default for SystemOptions {
+    fn default() -> Self {
+        Self {
+            auto_reload: true,
+            reload_on_activate: false,
+        }
+    }
+}
+
+impl SystemOptions {
+    /// Apply a single `(name, value)` pair from the wire. Returns `true`
+    /// if the option name was recognized (so the caller can log unknowns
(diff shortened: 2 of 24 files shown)


Cadacious:
Updated PR with the persistence layer + answers the "what other options need server-side action" question.

## Inventory: server-synced options in SystemOptions.xml

I counted every `serverSynch=` attribute:

| `serverSynch=` value | Count |
|---|---|
| `'false'` | 44 |
| `'true'` | **2** (these two) |

So the answer is: **only these two options sync to the server in the entire 2009 design**. The other ~140 options are client-local (volume sliders, brightness, control bindings, etc.). No additional server logic needed for any other option in `SystemOptions.xml`.

## DB persistence — added in commit `d3a9525`

**Schema** (`db/sgw/Players/Tables/sgw_player.sql`):
- `auto_reload BOOLEAN NOT NULL DEFAULT TRUE`
- `reload_on_activate BOOLEAN NOT NULL DEFAULT FALSE`

Defaults match SystemOptions.xml's `defaultBool` for each.

**Hydrate-on-login** (DB → cell entity):
- `player_load/core.rs` SELECT picks up the two columns
- `PlayerLoadData` carries them
- `world_entry_appearance.rs` reads them in the same round-trip as `bandolier_slot` (no extra DB hop)
- `BaseToCellMsg::InitPlayerState` plumbs a full `SystemOptions` struct
- `player_init` handler assigns it onto `CellEntity.system_options` BEFORE the content engine fires, so any chain that consults the option sees the persisted value, not the default

**Persist-on-change** (cell entity → DB):
- New `CellToBaseMsg::SystemOptionsUpdate` variant
- `updateSystemOptions` cell handler now async; fires the message after applying options in-memory
- New `cell_dispatch/system_options.rs` handler runs `UPDATE sgw_player SET auto_reload = $1, reload_on_activate = $2 WHERE player_id = $3`
- Same warn-on-no-rows / warn-on-write-failure discipline as `ActiveSlotUpdate`
- In-memory value still applies for the current session even if persistence fails — the user's checkbox toggle isn't completely lost when DB hiccups

**Tests** (3 new live-DB regression guards at sentinel range `0x7000_1600`):
1. Fresh player row has XML defaults (schema regression guard)
2. Persist round-trip flips columns and a second persist overwrites cleanly
3. Missing-row persist no-ops without panicking (race with character deletion)

**1581 workspace tests pass**, fmt + clippy clean.

## Re: other options that need server-side action

None. The XML is explicit — `serverSynch='true'` is the marker, and it's set on exactly these two. Everything else (volume, brightness, mouse sensitivity, etc.) is purely client-side.

If we ever want to add a NEW option that requires server logic (e.g., "auto-loot dropped items"), the recipe is:
1. Add to `SystemOptions.xml` with `serverSynch='true'`
2. Add a field to `SystemOptions` struct + match arm in `apply()`
3. Add a column to `sgw_player`
4. Update the hydrate/persist plumbing (the structural plumbing is mechanical now that the seams exist)
5. Wire the actual gameplay trigger

MyPalClara (commented):

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `parse_name_value_pairs` 256-entry cap | **The cap is documented as "defend against malformed payloads" but 256 is arbitrary.** The XML has ~140 options; a legitimate client would never send more. **Consider lowering to 200 or documenting why 256 was chosen.** |
| **G2** | `maybe_trigger_auto_reload` at both exit points | **The PR body says "at both exit points of `handle_use_ability`."** If a future refactor adds a third exit point, the auto-reload hook could be missed. **Consider centralizing the hook in a single post-fire function.** |
| **G3** | `reload_on_activate` skips unholster | **The PR body says "does NOT trigger on unholster because Phase A of `handle_reload` IS the draw animation and would loop."** This is correct, but the reasoning is subtle. **Add a comment at the unholster call site explaining why the hook is intentionally absent.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `SystemOptions` not persisted until follow-up | **The PR body says "Persistence to the DB is a follow-up."** But the DB schema change (`auto_reload`, `reload_on_activate` columns) is included in this PR. **The columns exist but aren't read on login — a player who sets the option, logs out, and logs back in will see the default.** Document this limitation in the PR body or add a TODO. |
| **G5** | `parse_bool` permissive | **The function accepts `"1"`, `"true"`, `"True"`, `"TRUE"` as truthy.** But the C++ client only sends `"1"` or `"0"`. **The extra tokens are belt-and-braces, but consider whether they could mask a bug (e.g., a typo `"ture"` would be falsy).** |
| **G6** | `active_ammo() == 0` check | **The auto-reload triggers when `active_ammo() == 0`.** But what if the weapon has no ammo type (melee)? **The `handle_reload` call would be a no-op, but it's still a wasted call.** Consider gating on `has_ammo_type()` first. |
| **G7** | `CellToBaseMsg::SystemOptionsUpdate` | **The message is sent from cell to base, but the base handler just persists to DB.** If the DB write fails, the in-memory state and DB state diverge. **Consider logging the failure or retrying.** |

| **G8** | No test for `maybe_trigger_auto_reload` | **The PR body lists 5 `maybe_trigger_reload_on_activate` tests but no `maybe_trigger_auto_reload` tests.** The auto-reload path is exercised indirectly via `handle_use_ability`, but a direct unit test would be clearer. |
| **G9** | No test for DB persistence round-trip | **The columns are added but there's no test that verifies `persist_system_options` writes and `load_player_data` reads.** The PR body says persistence is a follow-up, but the schema is here now. |
| **G10** | No test for `parse_name_value_pairs` with odd-length WSTRING | **The parser reads `u32` count then pairs of WSTRINGs.** If the payload is truncated mid-WSTRING, the parser should fail gracefully. **Consider a test with a truncated payload.** |

| **G11** | `system_options.rs` doesn't cite the XML line numbers | **The docstring says "SystemOptions.xml line 441" but the actual XML path isn't in the repo.** Consider citing the full path or adding a note that the XML is in the client PAK. |
| **G12** | No mention of the ~100 other options | **The PR body says "extending is mechanical" but doesn't list which options are next.** Consider adding a TODO with the high-priority options (e.g., `invertMouseY`, `showDamageNumbers`). |

---

## Summary

| Correctness issues | 3 | **Low** — G1-G3 are edge cases |
| Potential issues | 4 | **Low-Medium** — G4 is a known gap, G5-G7 are minor |
| Missing test coverage | 3 | **Low** — G8-G10 are nice-to-haves |
| Documentation | 2 | **Low** — G11-G12 are polish |

**Total: 12 gaps, all low-medium severity.**

---

## Priority Matrix

| Priority | Gaps | Effort |
|----------|------|--------|
| 🟢 Quick wins | G3 — add comment at unholster site | ~5 min |
| 🟢 Quick wins | G11 — cite XML path | ~5 min |
| 🟡 Consider | G4 — document persistence limitation | ~10 min |
| 🟡 Consider | G8 — add `maybe_trigger_auto_reload` test | ~20 min |
| ⚪ Follow-up | G9 — DB round-trip test (with persistence PR) | ~30 min |

---

## Verdict

**This PR correctly wires the two client checkboxes that did nothing before:**
1. `autoReload` — auto-reload when clip hits zero ✅
2. `reloadOnActivate` — reload on weapon switch if clip is partial ✅

The `updateSystemOptions` cell method is now implemented (was `UNIMPLEMENTED` stub). The 18 new tests cover the option parsing, wire format, and trigger guards. The DB schema is added for future persistence.

The 12 gaps are all low severity — the main caveat is G4 (options don't persist across logins yet), which is documented as out-of-scope.

**Ship it.** 🎯

Cadacious:
@coderabbitai review

Cadacious:
@coderabbitai review

Issue #304, opened 2026-05-16
Negative-logging audit: 40+ expectation seams across cell/base/content


Cadacious:
## Summary

The server logs liberally when things happen and almost never when expected things fail to happen. A three-way audit of the cell, base, and content/mission subsystems turned up 40+ expectation seams — call sites that assume a downstream effect will land but emit no signal when it doesn't. The seams are catalogued by tier below, with a four-PR plan for landing the fixes without sprawling cross-crate diffs.

The class of bug this targets: NPC fails to spawn, quest fails to advance, teleport silently drops a packet, AoI packet vanishes into an empty witness map. Currently each is invisible until a player files a ticket with no repro. A `warn!` or `error!` at the right seam turns every one of them into a greppable ops signal.

## Cross-cutting patterns

Three patterns account for most seams. Each has a uniform fix; the per-seam tables below are the application of these three patterns plus a handful of one-offs.

### Pattern A — `let _ = tx.send(...)` / `let _ = query.execute(...)`

Endemic across the content executor (`PlaySequence`, `StartMinigame`, dialog dispatch) and the `first_login` clear in `map_loaded.rs`. Errors are silently dropped. Fix: replace bare `let _` with `if let Err(e) = ... { warn!(...) }`. No call-site needs to handle the error differently — they all just need to log it.

### Pattern B — `rows_affected == 0` is inconsistently handled

Teleport and gate travel warn on it. `first_login` clear doesn't log at all. Inventory `remove_instance` warns but drops the actual count from the structured fields. Fix: always emit `rows_affected` and `expected` as paired structured fields so a single ops query (`rows_affected != expected`) surfaces every divergence in one place.

### Pattern C — Witness / lookup misses log at `trace!`

`send_to_witness` in [`crates/services/src/base/helpers.rs:133`](crates/services/src/base/helpers.rs) silently drops AoI packets at `trace!` level when `entity_to_addr` returns `None`. Same shape repeats in the delayed-cinematic resend in `map_loaded.rs`. Given the active `fix/issue-288-world-entry-spawn-glitches` work, this is the single biggest blind spot — invisible entities are exactly the bug class that branch is chasing.

---

## Tier 1 — Critical

Highest player impact, lowest effort. Most entries are one-line changes (`let _ =` → `if let Err`, or `trace!` → `warn!`).

| # | File:Line | Expectation | Current | Fix |
|---|---|---|---|---|
[...]

PR #397, opened 2026-05-25
feat(discord): notification crate + tracing-layer harvest + panic hook

Cadacious:
New crate `cimmeria-discord` that posts server events to Discord channels via webhooks for ops visibility during development. Wired into the server binary at startup; auto-harvests existing `warn!`/`error!` events via a tracing-subscriber `Layer`.

## Architecture

```
emit_*() ─┐
          ├─► SenderHandle ─► bounded mpsc(256) ─► sender task ─► reqwest POST
Layer  ───┘    (try_send,                            │
                drop on full)                  per-channel token bucket
                                                     │
                                                     ▼
                                                  Discord
```

**Two emit paths, one pipeline:**
1. **Explicit `emit_*` typed helpers** at semantically meaningful seams (login, world entry, mission complete, …).
2. **`DiscordLayer`** auto-harvests `warn!`/`error!` events into `Event::TracingEvent`, lifting structured fields (`reason=`, `entity_id=`, …) directly into the embed. The #304 negative-logging convention is what makes this work without instrumenting every site.

## What this gives you

- **40 event types** across **8 logical channels** (lifecycle, auth, world, chat, gameplay, gm, errors, ops). Each event has its own toggle in `[discord.events]`.
- **TOML config** with `${ENV_VAR}` interpolation for webhook URLs, per-channel rate limits, per-event toggles. Unknown keys rejected at parse time (typo guard).
- **Live reload** via `notify` + `ArcSwap`. Edit `config/discord.toml`, save → toggle takes effect on the next event. Parse failures keep the previous config in place (so an edit-in-progress typo doesn't kill notifications).
- **Custom panic hook** posts `Event::ServerPanic` via synchronous `reqwest::blocking` with a 2s timeout before the default hook aborts the process.

## Privacy + defaults

- **Whisper content is always `[hidden]`** in the embed regardless of how the channel is configured. The fact that a whisper occurred is visible (for moderation context); the message body never is. Enforced in `embed::format_chat`; pinned by `whisper_content_is_hidden_regardless_of_input` test — reverting the privacy branch trips it.
- **Defaults**: lifecycle/auth/world/gm/ops all on; warning off (noisy); chat-non-global off; death/loot/item-used off. Full default table in `EventToggles::default`.

## Best-practices checklist (all implemented)

| Concern | Implementation |
|---|---|
| Webhook secrets | `${ENV_VAR}` substitution. Config commits cleanly; secrets stay in env. |
| Back-pressure | Bounded mpsc (256), drop-on-full with atomic counter — tick loop never blocks. |
| Rate limiting | Per-channel token bucket. Burst capped at 5 (Discord's per-webhook budget). |
| HTTP retries | 2× exponential (250ms, 500ms) on 5xx + network errors. **No retry on 4xx**. |
| 429 handling | `Retry-After` honoured in-task. |
| Recursion safety | Layer filters explicit `target: \"cimmeria_discord\"` so HTTP-error tracing can't loop. |
| Embed size limits | Title 256, desc 4096, field-value 1024, total 6000. Visible `…` truncation. |
| Mockability | `DiscordSender` trait + `MockSender` + `HttpDiscordSender`. Same pattern as the mercury `Transport` trait from #358. |
| Graceful shutdown | `emit_server_shutdown` + 1s drain. |

## Files

- `crates/discord/` (new crate, 7 modules): `color`, `config`, `embed`, `event`, `layer`, `router`, `sender`. ~2.1k lines including tests.
- `config/discord.toml.example`: annotated user-facing config with every event type + defaults.
- `docs/architecture/discord-notifications.md`: design + ops runbook + \"adding a new event type\" instructions.
- `CLAUDE.md`: doc-update map row added under the routing rules.
- `crates/server/src/main.rs`: `cimmeria_discord::init` + `install_panic_hook` at the top of `main`; `DiscordLayer` registered in the existing layered subscriber setup; `emit_server_startup` / `emit_server_shutdown` at the lifecycle boundaries.

## Operator quick start

1. Copy `config/discord.toml.example` to `config/discord.toml`.
2. In Discord: Server Settings → Integrations → Webhooks → New Webhook (one per channel).
3. Set the matching env vars: `DISCORD_LIFECYCLE_WEBHOOK`, `DISCORD_AUTH_WEBHOOK`, …
4. `cargo run -p cimmeria-server` — should see a server-startup embed land within a second.

Missing config file → Discord silently disabled (no failure). Present-but-invalid file → fails at startup with a clear error (typo guard).

## Tests

**41 new unit tests** in `cimmeria-discord` covering:
- Formula correctness + overflow saturation + staircase steps + boundary jump
- Embed shape per variant + total budget enforcement + Unicode-safe truncation
- **Whisper privacy regression guard** (the single most important assertion in the PR)
- Router exhaustive coverage (every variant, every channel)
- Config TOML round-trip + env interpolation + unknown-key rejection + rate-limit clamp + should_post gating
- Sender token bucket burst + refill, 4xx no-retry, network retry cap, queue-full drop counters, pre-queue filtering
- Layer warn/error routing + level filtering + recursion guard + structured-field preservation

**1668 workspace tests pass** (added 41 new + auto-picked up other landings on main). `cargo clippy --all-targets -- -D warnings` clean.

## Adding a new event type

Documented in `docs/architecture/discord-notifications.md`:

1. Add a variant to `EventKind` + matching `Event` variant.
2. Add the field to `EventToggles` + default + `is_enabled` arm.
3. Route in `router::channel_for` (no `_` fallback — compile error if forgotten).
4. Add a `format_event` arm in `embed.rs`.
5. Add a typed helper in `lib.rs`.
6. Document in `config/discord.toml.example`.

The `event_kind_all_matches_variant_count` test pins the count so step 2 not getting done trips a test failure.

## Test plan

- [ ] Set up a Discord server with the 8 channels and webhook URLs.
- [ ] `cargo run -p cimmeria-server` — confirm `Server up` embed lands.
- [ ] Edit `discord.toml` to flip `warning = true`, save → next `warn!` from the server posts to `#errors` without restart.
- [ ] Ctrl-C → confirm `Server shutdown` embed lands within 1s.
- [ ] Deliberately panic (e.g. attach a debugger, force an assert) → confirm `💥 Server panic` posts to lifecycle.

## New Features
* Added Discord webhook notifications for server and player events
* Configurable per-channel routing with event-type toggles
* Live configuration reload without server restart
* Privacy-preserving chat whispers (message content hidden)
* Per-channel rate limiting to prevent notification spam
* Automatic error and warning notifications to Discord

## Documentation
* Added example Discord configuration file
* Added Discord notifications architecture documentation

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/397?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `install_panic_hook` uses `reqwest::blocking` | **The panic hook uses synchronous HTTP with a 2s timeout.** If the webhook is unreachable, the process hangs for 2s before dying. This is probably acceptable (you want the panic to post), but document the timeout. |
| **G2** | `DiscordLayer` filters by `target: "cimmeria_discord"` for recursion safety | **The recursion guard checks the target, not the module path.** If someone adds `tracing::warn!(target: "cimmeria_discord", ...)` elsewhere, it would be filtered. Consider using a more specific marker (e.g., a span or a custom field). |
| **G3** | `try_send` on full queue increments `dropped_full` but doesn't log | **When the queue is full, events are silently dropped.** The counter is incremented, but there's no `warn!` or `trace!` at the drop site. Consider a rate-limited log (e.g., once per 100 drops) so operators know events are being lost. |
| **G4** | `enforce_total_budget` truncates fields in order | **If the total exceeds 6000 chars, fields are truncated in iteration order.** But iteration order of a `Vec` is insertion order, which may not be priority order. Consider truncating least-important fields first, or document that truncation order is arbitrary. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G5** | `notify` crate for file watching | **The `notify` crate is cross-platform but has platform-specific quirks.** On macOS, `kqueue` doesn't always fire on file content changes (only metadata). The PR uses `macos_kqueue` feature — verify this works for content changes. |
| **G6** | `ArcSwap` for config hot-swap | **`ArcSwap` is lock-free but not wait-free.** Under extreme contention (many threads reading config simultaneously), there could be brief stalls. Probably fine for this use case, but document the choice. |
| **G7** | Token bucket refill is per-send, not per-tick | **The token bucket refills on each send attempt, not on a background timer.** If no events are sent for a long time, the bucket could be empty when a burst arrives. Verify the refill logic handles this case. |
| **G8** | `wiremock` is a dev dependency but listed in `[dependencies]` | **Looking at the Cargo.toml diff, `wiremock` is added to the workspace `[dependencies]`.** If it's only used in tests, it should be `[dev-dependencies]`. Verify the dependency categorization. |
| **G9** | 8 webhook URLs = 8 env vars | **Each channel needs its own webhook URL.** If an operator wants all events in one channel, they'd set all 8 env vars to the same URL. Consider a `default_webhook` fallback for channels without explicit URLs. |

MyPalClara:

| **G10** | No integration test for live reload | **The unit tests cover config parsing, but there's no test that exercises the `notify` watcher + `ArcSwap` swap.** The manual test plan covers this, but an automated test would catch regressions. |
| **G11** | No test for panic hook | **The panic hook is installed but not tested.** A test would need to fork a subprocess and verify the webhook was called. Consider a test that at least verifies `install_panic_hook` doesn't panic itself. |
| **G12** | No test for `DiscordLayer` with real tracing subscriber | **The layer tests use `MockSender`, but there's no test that wires the layer into a real `tracing_subscriber::Registry` and verifies events flow through.** |
| **G13** | No test for embed truncation with Unicode | **The embed tests cover ASCII truncation, but what about multi-byte UTF-8?** Truncating mid-codepoint would produce invalid UTF-8. Verify `enforce_total_budget` handles Unicode correctly. |
| **G14** | No test for rate limiter under sustained load | **The rate limiter tests cover burst + refill, but not sustained load over time.** A test that sends 100 events over 10 seconds would verify the bucket doesn't starve. |

MyPalClara:
## Category 4: Documentation

| **G15** | `discord-notifications.md` doesn't mention the 256-entry queue cap | **The ADR mentions "bounded mpsc (capacity 256)" in the best-practices table but not in the architecture section.** Add it to the ASCII diagram or the prose. |
| **G16** | No mention of the `SenderStats` fields in the ADR | **The ADR says "Self-observability: `SenderStats` counters" but doesn't list the fields.** Add the field names so operators know what to look for. |
| **G17** | `config/discord.toml.example` not included in the file list | **The PR body mentions this file, but I don't see it in the files changed.** Verify it's included. |
| **G18** | No mention of the `DISCORD_CONFIG_PATH` env var in the ADR | **The ADR says "Path: `config/discord.toml` (override with `DISCORD_CONFIG_PATH`)" but doesn't explain when you'd want to override.** Add a use case (e.g., "for testing with a different config"). |

MyPalClara:
## Category 5: Privacy / Security + Summary

| Gap | Location | Concern |
|-----|----------|---------|
| **G19** | Whisper privacy is enforced in `embed::format_chat` | **The privacy guard is in the embed formatter, not the event emitter.** If someone adds a new code path that formats whispers differently, the guard could be bypassed. Consider enforcing at the `Event::ChatWhisper` construction site as well. |
| **G20** | Webhook URLs in env vars could leak to logs | **If someone does `tracing::info!("config: {:?}", config)`, the webhook URLs would be logged.** The URLs contain tokens. Consider implementing `Debug` for `Config` that redacts URLs, or document that configs should never be logged. |

---

## Summary

**20 gaps found, all low-medium severity.**

| Category | Count |
|----------|-------|
| Correctness issues | 4 |
| Potential issues | 5 |
| Missing test coverage | 5 |
| Documentation | 4 |
| Privacy/Security | 2 |

**What's excellent:**
- The tracing layer integration harvests existing `warn!`/`error!` events without per-site instrumentation
- The whisper privacy guard is bulletproof (test pins it)
- The rate limiter is production-grade (per-channel, burst-capped, 429-aware)
- The config validation is thorough (unknown keys rejected, live reload with debounce)
- The "Adding a new event type" checklist is complete with a count-pinning test
- 41 new tests covering formula, embed shape, truncation, rate limiter, layer harvest, recursion guard, whisper privacy

**This is a complete Discord notification system. 40 event types, 8 channels, live reload, automatic warn/error harvest.**

Cadacious:
## Review-fix pass (commit a816a31e)

### Addressed

**Copilot inline (13/13):**
- ✅ `main.rs:71` — hard-fail (exit 2) on invalid config; file-missing remains a soft-fail
- ✅ `main.rs:221` — gate the 1 s shutdown drain on `discord_enabled && should_post(ServerShutdown)` so an unconfigured server doesn't pay the latency
- ✅ `discord-notifications.md:35`/`:144`, `lib.rs:10`, `layer.rs:6`/`:312`, `embed.rs:103`/`:598` — scrubbed every `#304` and `PR Q1` reference; replaced with the stable convention name + link to `docs/architecture/negative-logging-convention.md`
- ✅ `embed.rs:717` — rewrote `enforce_total_budget` as a converging loop that trims description, then fields back-to-front, and bails only when no progress is possible. Pins ≤ 6000 chars in all reachable cases.
- ✅ `sender.rs:277` — bucket now tracks `rate_per_min`; mismatch with the current config triggers a rebuild so live-reload of `rate_limit_per_min` actually takes effect
- ✅ `sender.rs:29` + `config.rs:554` — dropped admin-api endpoint claims; comments now describe only what's wired in this PR

**Clara (8/20):**
- ✅ G4 (truncation order): fields now trimmed back-to-front (least-important last → trimmed first)
- ✅ G13 (Unicode budget): new test exercises 😀/🎯 multi-byte content through `enforce_total_budget`
- ✅ G20 (URL leak via `Debug`): custom `Debug` impl on `Config` + `ChannelConfig` prints `<redacted>`; pinned by `debug_format_redacts_webhook_url`

### Held (with rationale)

**Copilot:**
- None.

**Clara:**
- **G1** (panic-hook 2 s timeout): documented inline already (`crates/discord/src/lib.rs` `install_panic_hook` and `post_panic_synchronous` doc comments); the 2 s cap is by design.
- **G2** (recursion guard target): the explicit-target filter is intentional — submodule-path targets like `cimmeria_discord::layer::tests` need to harvest normally. Switching to a span-marker would lose ergonomic emit-side syntax.
- **G3** (silent drop log): `dropped_full` counter is already incremented + visible via `SenderStats`. Adding a rate-limited log is a future polish; would need its own debounce to avoid log-storming.
- **G5** (`macos_kqueue` content events): we use `notify` v8 RecommendedWatcher which selects FSEvents on macOS automatically — `macos_kqueue` is only used when the host lacks FSEvents support (rare).
- **G6** (`ArcSwap` not wait-free): true but irrelevant for our load profile.
- **G7** (token bucket per-send refill): the bucket refills lazily based on elapsed time — sustained idle then burst still works correctly; the `token_bucket_refills_over_time` test pins it.
- **G8** (wiremock placement): false alarm. `wiremock` is in `[workspace.dependencies]` (the shared *source*), but only pulled into `[dev-dependencies]` in `crates/discord/Cargo.toml:24`.
- **G9** (per-channel webhook URLs): operators who want everything in one channel point all 8 env vars at the same URL — explicit > implicit. Adding a `default_webhook` would muddle the "channel is configured iff URL present" invariant.
- **G10/G11/G12/G14** (more tests): live-reload + panic-hook integration tests are useful follow-ups but cost-disproportionate for this PR (subprocess fork for panic, race-prone for notify). The layer is already tested with a real `tracing_subscriber::Registry` (see `layer.rs` tests).
- **G15/G16/G17/G18** (ADR polish): tracked as docs follow-up — the values are accurate where they exist in the source.
- **G19** (defense-in-depth at construction site): the formatter is the only path from `Event::Chat` to wire bytes; enforcing twice doesn't add coverage and would clutter the `emit_chat` ergonomics. The regression test on the formatter is the line of defense.

**Codecov (77.41% patch):**
- Most of the uncovered lines are in `lib.rs` (panic hook + global `emit_*` convenience helpers — both rely on `OnceLock` state that can't be reset between tests) and `config.rs` (file-watcher edge cases that need real FS events). Coverage there is structural; not adding theatre tests to chase a number.

### Verification

```
cargo fmt --all -- --check                                            ✓
cargo clippy --workspace [excludes] --all-targets -- -D warnings      ✓
cargo nextest run --profile=ci --workspace [excludes]                 ✓ 1671 passed (+3)
cargo nextest run --profile=ci -p cimmeria-discord                    ✓ 44 passed
```

Cadacious:
@coderabbitai review

Cadacious:
## Codecov triage (bf8cd77d)

Codecov's prior report flagged 79.4% patch coverage on the crate. Walked the gaps and addressed three of the four big ones with regression-shaped tests; the fourth is structurally untestable (panic hook + OnceLock global init from a file).

### Addressed (+13 tests → 59 in `cimmeria-discord`, 1704 total)

**`HttpDiscordSender` real wire path (sender.rs)** — wiremock-driven tests pin the 204/4xx/429/network paths. `parse_retry_after_str` extracted as a callable helper so the four header-format branches (fractional, integer, 60s clamp, negative-clamps-to-zero, garbage) each get a microsecond-cost unit test rather than a per-branch wiremock spin-up. The refactor also closed a NaN hole — a hostile server returning `"NaN"` would have parsed as a finite f64 through `str::parse` and panicked in `Duration::from_secs_f64`.

**`FieldCollector` numeric/bool visitor arms (layer.rs)** — emits a tracing event with `i64`, `u64`, `bool`, `f64`, and `&str` fields and asserts every formatted value lands in the recorded embed. Pins each `record_*` arm against drift.

**Typed `emit_*` wrappers in lib.rs (lib.rs at 0% → covered)** — one comprehensive test boots a wiremock server, calls `init_with_config` with every channel routed at it + every toggle ON, fires all 13 `emit_*` helpers, and asserts wiremock observed ≥13 POSTs spread across ≥5 distinct paths (proving the router routes typed-helper Events correctly). Plus `install_panic_hook_is_idempotent` (triple-install no-op) and `global_accessor_does_not_panic`.

The lib.rs test is the only one in the binary that touches the `GLOBAL OnceLock` — `TESTS_INIT_GUARD` documents that adding a second initialiser test in the same binary would race.

### Deliberately not addressed

**`post_panic_synchronous` and the file-based `init()` path** — both are structurally hard to cover without compromising the contract:

- `post_panic_synchronous` requires actually panicking. The default hook then aborts the process, terminating the test runner.
- `init(path)` would need a tempfile + `ConfigWatcher` (the notify crate's real fs watcher) and would conflict with the lib.rs test's `init_with_config`. They both write to the same `OnceLock`; only the first wins. Picked `init_with_config` since it covers the bulk of lib.rs in one shot.

**`ConfigWatcher` notify-task body (config.rs ~140 lines)** — the live-reload tokio task that owns the `notify::RecommendedWatcher`. Testing it deterministically would mean writing tempfiles, triggering filesystem-notify events, and waiting on the debounce window — every step adds either flake risk or seconds of test time. The watcher's failure modes are hot-path (debounce, parse-keeps-old-config) and would be observable at runtime; the cost/benefit doesn't favour pinning them with FS-shaped tests.

### Verification

```
cargo fmt --all -- --check                            ✓
cargo clippy --workspace [excludes] --all-targets     ✓
cargo nextest --workspace [excludes]                  ✓ 1704 passed
cargo nextest -p cimmeria-discord                     ✓ 59 passed
```

Expected codecov delta on the next run: lib.rs jumps from 0% → ~75–80% (the 13 emit_* helpers covered, `init_with_config`/`emit`/`global` covered; remaining gap is `init()` from-file + `post_panic_synchronous`). sender.rs and layer.rs both rise into the high-90s. config.rs holds in the low-70s — that's the ConfigWatcher task body, deliberately not chased.

PR #401, opened 2026-05-26
fix(cell): chain-fired dialogs send NPC entity id, not player id

Cadacious:
## Summary

Chain-fired dialogs (`display_dialog` content action) and the
`initialResponse` handler were sending the **player's** entity id as
the wire `EntityId` field of `onDialogDisplay`, where the client
expects the **NPC's** id. The client uses that id as the key into
`LookupEntityListenerEntry` to bind the dialog portrait actor and
the per-screen speaker entity — so dialogs opened with the player
bound as the speaker, blanking the portrait and rendering the
player's name on every screen. The interact path itself was already
correct; only the chain + initialResponse paths needed the fix.

User-visible symptom: in Castle_CellBlock, Col Marsh's briefing
dialog (chain 1051 → `display_dialog 4001`) shows the player's name
on every screen with a blank portrait. Same shape for every
chain-fired dialog in the codebase.

## Fix

Threads the NPC entity id through to the wire:

- New `CellEntity::last_interaction_target: Option<u32>` (player-only
  pin, mirrors python's `SGWPlayer.lastInteractionTarget`). Written
  by `handle_interact` after the distance check passes.
- `handle_initial_response` reads the pin and uses it as the wire
  `EntityId`. Warns + aborts when unset (rather than falling back to
  the player, which was the bug).
- `dialog::display` resolves the NPC id as:
  1. `params["target_entity_id"]` — stamped by `fire_interact_*` for
     chains fired off an interact trigger.
  2. Player's `last_interaction_target` — covers `OnDialogChoice`
     follow-ups where the trigger event carries no NPC.
  3. Warn + abort — the warn level is load-bearing for operator
     triage.

## Evidence

New V5-session RE finding at
`docs/reverse-engineering/findings/dialog-portrait-lookup.md` traces
`Event_NetIn_DialogDisplay` through the SGW.exe binary:

```
FUN_00d25310 (handler) → FUN_00d24f10 → FUN_00d22c90
  → LookupEntityListenerEntry(GameEntityManager, npc_entity_id, ...)
  → slot 17 pin via FUN_00c67bd0
  → Event_UI_UnitMappingChanged
  → Lua createCharacterPortrait
```

The `LookupEntityListenerEntry` call is keyed by the wire EntityId,
not `GENERICPROPERTY_DatabaseId`. The earlier backfill in #386 was
orthogonal to the portrait widget — necessary for other paths, but
insufficient on its own to render the dialog portrait.

## Test plan

- [x] `cargo clippy -p cimmeria-services --all-targets -- -D warnings`
- [x] 534 cell tests + 21 executor tests + 12 dialog-themed tests pass
- [x] 7 new regression guards added, all bug-shape (fail when fix is
      reverted):
  - dispatch: pins `args[0..4]` are the NPC id, pins
    `last_interaction_target` write semantics, pins the abort + WARN
    when the pin is missing.
  - executor::dialog: pins params resolution, fallback to the player
    pin, and the WARN-and-abort path when neither is available.
- [ ] Manual: log in, talk to Col Marsh in Castle_CellBlock, verify
      his portrait renders and his name appears on his lines.
- [ ] Manual: same for every other chain-fired dialog (chains
      1014/1015/1021/1032/1052/1056/1057/1061/1099 reference
      `display_dialog`).

## Known remaining gaps (separate issues, out of scope)

- **Prisoner 329's portrait still blank** even with the correct wire
  EntityId — his interact path was already correct, so this PR
  doesn't change his case. The RE finding flags the AoI
  listener-table vs spatial AoI grid distinction as the next dig.
- **`speakers[256]` is empty in CookedData**, so Col Marsh's
  per-screen speaker *name* (separate from the portrait actor) will
  still fall back to the player on his lines until that PAK/seed row
  is populated.

### DIFF
--- a/crates/entity/src/cell_entity/mod.rs
+++ b/crates/entity/src/cell_entity/mod.rs
@@ -452,6 +452,26 @@ pub struct CellEntity {
     /// Reference: `python/cell/SGWPlayer.py:setLooting()`
     pub looting_entity: Option<u32>,
 
+    /// Entity ID of the NPC the player most recently interacted with
+    /// (player entities only).
+    ///
+    /// Set on every successful `handle_interact`. Read by paths that need
+    /// to know "which NPC is on the other side of this conversation" but
+    /// don't get that ID through their own wire frame — notably
+    /// `handle_initial_response` (the client sends only
+    /// `interaction_set_map_id`, mirroring python's
+    /// `SGWPlayer.initialResponse` which used `self.lastInteractionTarget`)
+    /// and content-engine `display_dialog` / `start_dialog` actions fired
+    /// from chains that don't carry `target_entity_id` in params (e.g. a
+    /// follow-up dialog triggered by `OnDialogChoice`).
+    ///
+    /// The wire `EntityId` field of `onDialogDisplay` must be the NPC's
+    /// entity ID, not the player's — the client looks it up in the AoI
+    /// listener table via `LookupEntityListenerEntry` to bind the dialog
+    /// portrait actor (see
+    /// `docs/reverse-engineering/findings/dialog-portrait-lookup.md`).
+    pub last_interaction_target: Option<u32>,
+
     /// Entity ID of the currently-open vendor (only for player entities).
     pub vendor_entity: Option<u32>,
 
@@ -636,6 +656,7 @@ impl CellEntity {
             loot: Vec::new(),
             next_loot_index: 1,
             looting_entity: None,
+            last_interaction_target: None,
             vendor_entity: None,
             current_target_id: None,
             active_bandolier_slot: 0,
--- a/crates/services/src/cell/content/executor/dialog.rs
+++ b/crates/services/src/cell/content/executor/dialog.rs
@@ -9,21 +9,73 @@ use crate::cell::space_manager::SpaceManager;
 /// `Action::DisplayDialog` and `Action::StartDialog` — both render a dialog
 /// to the player. They take different field names but the same id semantics,
 /// merged into one handler.
+///
+/// The wire `EntityId` field of `onDialogDisplay` MUST be the NPC the
+/// player is talking to — the client uses it as the key into
+/// `LookupEntityListenerEntry` to bind the dialog portrait actor and the
+/// per-screen speaker entity (see
+/// `docs/reverse-engineering/findings/dialog-portrait-lookup.md`).
+/// Resolution order, from most to least direct:
+///
+/// 1. Chain `params["target_entity_id"]` — present when the chain was
+///    fired from an `InteractTag` / `InteractTemplate` trigger
+///    (`fire_interact_*` stamps it into the context).
+/// 2. Player's `last_interaction_target` — the per-player pin set by
+///    `handle_interact`. Covers follow-up chains (e.g. an
+///    `OnDialogChoice` trigger that fires another `display_dialog` on
+///    the same NPC) where the trigger event itself carries no NPC.
+/// 3. Abort with a `warn` — falling back to the player's own entity ID
+///    here is the bug this commit fixed: the client would bind the
+///    player as the dialog speaker, blanking the portrait and
+///    rendering the player's name for every screen.
 #[tracing::instrument(
     name = "dialog.display",
     level = "info",
     skip_all,
-    fields(entity_id, dialog_id, chain_id)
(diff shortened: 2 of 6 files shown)


MyPalClara:
## PR #401 Review — Gap Analysis

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `handle_interact` pin placement | **The pin is written AFTER the distance check but BEFORE the interaction-type dispatch.** This is correct for the dialog path, but what about vendor/trainer/loot? If a player interacts with a vendor, `last_interaction_target` is set to the vendor's entity ID. A subsequent chain-fired `display_dialog` (from a different trigger) would use that vendor ID as the dialog speaker. **Is this the intended behavior?** The Python reference would need to be checked — does `lastInteractionTarget` get overwritten on every successful interact, regardless of type? |
| **G2** | `display()` params cast | **`v.as_u64().map(\|v\| v as i32)` truncates high bits.** If `target_entity_id` is ever > `i32::MAX` (unlikely but possible for NPC IDs), the cast silently wraps. **Consider `v.as_i64().map(\|v\| v as i32)` or explicit bounds check.** |

---

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `last_interaction_target` lifetime | **The pin is never cleared.** If a player interacts with NPC A, then walks away and triggers a chain that fires `display_dialog` without `target_entity_id`, the dialog opens against NPC A even if A is now out of AoI. **The client will fail the `LookupEntityListenerEntry` lookup and blank the portrait anyway, but the server doesn't know that.** Consider clearing the pin on world transfer, respawn, or after some timeout. |
| **G4** | `out_of_range_interact_does_not_pin_target` test | **The test verifies the pin is `None` after an out-of-range interact, but it doesn't verify the pin is preserved if it was already set.** A player could have a valid `last_interaction_target` from a prior interact, then click an out-of-range NPC — the test doesn't verify the prior pin survives. |
| **G5** | `handle_initial_response` abort path | **The abort logs a `warn` but doesn't send any client feedback.** The player clicks an NPC, nothing happens, no error message. **Consider sending a system message or at least documenting that the client-side interaction prompt shouldn't have been visible in this state.** |

---

| **G6** | No test for vendor/trainer/loot setting `last_interaction_target` | **The tests only cover the dialog path.** There's no test verifying that interacting with a vendor also sets `last_interaction_target`. If the intent is "any successful interact sets the pin," this should be tested. If the intent is "only dialog interacts set the pin," the code should be changed. |
| **G7** | No test for `display()` with both params AND pin set | **The test `display_uses_target_entity_id_from_chain_params` verifies params wins, but it doesn't set `last_interaction_target` on the player.** A test with both set would verify the priority order more explicitly. |
| **G8** | No test for the existing `interact_sends_dialog_when_nearby` wire format | **The existing test was updated to assert the wire `EntityId`, but there's no test that verifies the PRIOR behavior was wrong.** A "before/after" style test (or at least a comment) would make the regression guard clearer. |

---

### Category 4: Documentation

| **G9** | `dialog-portrait-lookup.md` is excellent but long | **274 lines is a lot for a single finding.** Consider whether the "Implementation Impact" section (which is server-side guidance) should be in a separate doc or in the code comments. |
| **G10** | PR body "Known remaining gaps" section | **The Prisoner 329 gap is documented but not linked to an issue.** Consider filing an issue for the AoI listener-table vs spatial AoI grid investigation. |
| **G11** | `last_interaction_target` doc comment | **The doc comment is excellent but doesn't mention the lifetime/clearing behavior.** Add a note about when (if ever) the pin is cleared. |

---

### Summary

| Correctness issues | 2 | **Low** — G1 is probably intended, G2 is unlikely |
| Missing test coverage | 3 | **Low** — G6-G8 are nice-to-have |
| Documentation | 3 | **Low** — G9-G11 are polish |

**Total: 11 gaps, all low severity.**

---

### The Hardest Question

**Is G1 (pin written for all interaction types) actually a problem?**

The Python reference probably overwrites on every interact — the field is called `lastInteractionTarget`, not `lastDialogTarget`. The content author is responsible for ensuring chains that fire `display_dialog` either:
1. Are triggered from an interact (so `target_entity_id` is stamped), or
2. Are triggered from a context where `lastInteractionTarget` is the right NPC

**Conclusion:** G1 is probably intended behavior, but worth verifying against the Python reference.

PR #409, opened 2026-05-26
fix(respawn): full inventory re-init bundle, not just onUpdateItem

Cadacious:
## The bug

After same-world respawn:
- Inventory UI is empty (no items visible).
- Equipment slots are empty.
- New pickups don't show up either.

## Root cause

`ReanchorPlayer`'s `CREATE_BASE_PLAYER` destroys the client-side pawn actor and instantiates a fresh one. The new pawn comes up with an empty `InventoryComponent` — **no bag list registered**.

PR #387 (the previous respawn fix) pushed `onUpdateItem` after reanchor to repopulate items. But `onUpdateItem` entries only slot into bags the client has already registered. With no bags on the new component, entries silently drop. Same for new pickups — they `send_full_inventory_update` against the same broken bag-less state.

## Comparison with working path

| Path | Order |
|---|---|
| World entry (`map_loaded.rs:336-371`) — WORKS | onBagInfo → onActiveSlotUpdate → onCashChanged → onUpdateItem |
| Respawn (`ListInventoryItems` handler) — BROKEN | onUpdateItem only |

## Fix

New helper `send_full_inventory_resync` ships the canonical 4-packet bundle in order. The `CellToBaseMsg::ListInventoryItems` handler (the post-`ReanchorPlayer` resync hook) routes through the new helper. Grant/move/remove paths continue to use `send_full_inventory_update` unchanged — those run against a healthy client-side `InventoryComponent`.

Cross-world respawn unaffected — it falls through to `GateTravel`, which re-runs the full world-entry handshake.

## Tests

Live-DB Type 3 regression guard: `resync_sends_bag_info_active_slot_cash_then_items_in_order` — inserts a player with non-default naquadah + bandolier_slot, inserts a slappack, calls `send_full_inventory_resync`, drains the test transport, asserts exactly 4 packets in the expected order with predicted wire bytes for the first three. Reverting the fix trips an exact-count failure with the bug-shape named in the message.

## Test plan

- [ ] CI: `cargo nextest --profile=ci-live-db -p cimmeria-services` runs the resync guard.
- [ ] In-game: die, respawn → inventory shows all pre-death items, bandolier slot indicator points at the persisted slot, naquadah balance is correct.
- [ ] In-game: respawn then loot a slappack → item appears in the bag.

## Release Notes

* **Bug Fixes**
  * Improved inventory state restoration to ensure accurate synchronization of items, active equipment slots, and currency when players rejoin the game world.

* **Tests**
  * Added validation tests for inventory re-initialization sequence and message ordering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SandboxServers/Cimmeria/pull/409?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/entity/src/inventory.rs
+++ b/crates/entity/src/inventory.rs
@@ -143,10 +143,22 @@ impl Inventory {
     /// Serialize bag info for `onBagInfo(ARRAY<BagInfo>)`.
     ///
     /// Wire: `count:u32`, per bag: `bagId:i32, numberOfSlots:i32`.
+    ///
+    /// Bags are emitted in ascending `bag_id` order. The client doesn't
+    /// care about iteration order (it just registers each entry by id),
+    /// but the deterministic order keeps the on-wire bytes stable so
+    /// wire-level regression tests and pcap-replay guards can compare
+    /// exact byte sequences across runs — `HashMap::values()` order
+    /// varies between `HashMap` instances even with identical contents
+    /// when their RandomState diverges, which surfaced in CI as
+    /// non-deterministic bag-order failures in the post-respawn
+    /// resync guard.
     pub fn serialize_bag_info(&self) -> Vec<u8> {
         let mut buf = Vec::with_capacity(4 + self.bags.len() * 8);
         buf.extend_from_slice(&(self.bags.len() as u32).to_le_bytes());
-        for bag in self.bags.values() {
+        let mut bags: Vec<&Bag> = self.bags.values().collect();
+        bags.sort_by_key(|b| b.bag_id);
+        for bag in bags {
             buf.extend_from_slice(&bag.bag_id.to_le_bytes());
             buf.extend_from_slice(&bag.slots.to_le_bytes());
         }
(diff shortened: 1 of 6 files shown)


Cadacious:
Drafting per offline discussion — the resync bundle is a workaround that re-ships state the client lost when we destroyed the pawn. The real fix is to **stop destroying the pawn on respawn**, which means finding a way to un-ragdoll the client-side pawn actor without the CREATE_BASE_PLAYER recreate. Spinning up an archaeology dig for that path. This PR stays as the fallback if the dig comes up empty.

Cadacious:
**Archaeology came up empty on the un-ragdoll path.** Findings:

- `APawn::TermRagdoll` has no UScript `exec*` hook in the registration table — no Kismet action, entity method, or property write can reach it from the wire.
- The only ragdoll path in shipped kismet is one-way: `Entity_Death` → `InitRagdoll`.
- `onStateFieldUpdate` clearing `BSF_Dead` walks `FUN_00e6e330` which adjusts cursor + nameplate tint — never touches ragdoll physics.
- No `USeqEvent_EntityRevive` / standup event class exists.

The only structurally-feasible alternative — suppressing `onSequence Entity_Death` to prevent ragdoll from firing in the first place — would lose the canonical death visual. **Ragdoll is the intended death behavior and must stay**, so this path is closed.

The resync bundle in this PR is therefore the canonical fix, not a workaround. Promoting back to ready.

Issue #416, opened 2026-05-26
Workspace coverage uplift to ~88% — exclusions, prioritized tests, and ratcheted component targets


Cadacious:
# Workspace coverage uplift to ~88% — exclusions, prioritized tests, and ratcheted component targets

## Why

Today's snapshot (codecov main, commit `8eeaee7a`):

- **Workspace line coverage: 77.13%**
- 1,351 catalogued tests across the workspace
- 9,733 misses across 330 files
- One crate (`mercury`) is **already below** its existing component target
- One crate (`admin-api`) has no real baseline (`target: auto`)
- Two crates (`upk`, `upk-objects`) are noise — code consumed only by `bin/` tooling and the Tauri editors (which are already excluded), but the crates themselves still show up in the report at 0% / 42.9% and drag the totals
- The `services/src/wire_log/decoders/generated.rs` file is **1,330 lines of auto-generated code at 0% coverage** — by itself it depresses the workspace number by ~3pp without representing any hand-written behaviour worth testing

This issue is the consolidated plan to:

1. **Clear out coverage noise** — exclude generated and build-time-only code from the report so the number reflects what we actually care about.
2. **Land high-ROI tests** on files where multiple test types (per [TESTING.md](../blob/main/TESTING.md)'s 12-type taxonomy) catch different bug shapes — gameplay-critical paths first.
3. **Ratchet the codecov component targets** in [codecov.yml](../blob/main/codecov.yml) so the new baseline is enforced, not just achieved-and-allowed-to-drift.

**Target end state after the full plan: ~88% workspace coverage** (realistic), with component targets within 2-3pp of measured coverage per crate (locks in gains, tolerates jitter).

This work composes — each phase is a separate PR (or PR family) and Phase 0 alone is +7pp coverage for ~15 lines of YAML.

---

## Current state — per-crate breakdown

From codecov API on `main` (`https://api.codecov.io/api/v2/github/sandboxservers/repos/Cimmeria/totals`):

| Crate | Files | Lines | Misses | Cov today | Component target today | Status |
|---|---:|---:|---:|---:|---:|---|
| `services` | 195 | 26,322 | 5,795 | 77.98% | 62% | well above target |
| `admin-api` | 18 | 1,713 | 1,425 | 16.81% | `auto` | no real baseline |
| `upk` | 9 | 891 | 891 | 0.00% | (no component) | propose exclude |
[...]

Issue #331, opened 2026-05-18
feat(cell): effect-script dispatcher + HealHealth/HealFocus/MeleeDamage scripts


Cadacious:
## Plain English

Effects (the per-pulse damage/heal/buff units that an Ability launches) have a `script_name` column on the server-side database that says "when this effect pulses, run script X." Cimmeria's effect data model already includes `EffectDef.script_name: Option<String>` ([`crates/entity/src/abilities.rs:91`](../blob/main/crates/entity/src/abilities.rs#L91)), but there is **no dispatcher** — the field is loaded and ignored. As a result we have no working heal abilities (HealHealth, HealFocus) and no scripted-melee damage path.

This issue covers wiring up an effect-script dispatcher + the three scripts the fan-server has used to validate the pattern: `HealHealth`, `HealFocus`, `MeleeDamage`. These three together unlock the medkit/stimpack heal flow and the Melee Strike ability.

### What the client expects

The effect script-name dispatch is **purely a server-side abstraction** — the client never sees `script_name`. Confirmed by Ghidra:

`CookedData:EffectType` parser at `ghidra://SGW.exe@0x015dfd60` (type ID `0x0c`, struct size `0x48`) reads the following fields and **no `script_name` among them**:

| Slot | Field name | Type |
|---|---|---|
| +4 | `AbilityId` | uint32 |
| +8 | `EffectId` | uint32 |
| +12 | `name` (display name) | string |
| +16 | `EffectDesc` | string |
| +20 | `targetCollectionId` | int32 |
| +24 | `isChanneled` | bool |
| +26 | `EffectSequence` | uint16 |
| +28 | `PulseCount` | uint16 |
| +32 | `PulseDuration` | float |
| +36 | `Delay` | float |
| +40 | `Target_Collection_Method` | uint16 |
| +44 | `TCM_Param1` | string |
| +48 | `TCM_Param2` | string |
| +52 | `Use_Ability_Velocity` | bool |
| +56 | `Flags` | uint32 |
| +64 | `IconLocation` | string |

The client identifies effects by `(AbilityId, EffectId)` integer pair only. The behavior dispatch is entirely server-side; we get to define our own script-naming scheme.

### Damage type enum — verified

For `MeleeDamage` (and downstream damage effects) we need the damage-type integer that feeds `qrCombatDamage(stat, damageType, baseDamage)`. All values via `FUN_00acbb10` Lua-binding constants:

[...]

Issue #61, opened 2026-03-29
AoE and ground-target ability support


Cadacious:
## Summary

Single-target abilities work with full damage calculation. Ground-target AoE abilities (useAbilityOnLocation) are logged but not processed.

## Current State in Rust

- `crates/services/src/cell/abilities.rs` (647 lines): `handle_use_ability()` fully working for single-target.
- `crates/services/src/cell/cell_methods/player.rs:120`: Ground-target AoE is TODO.
- `crates/services/src/cell/dispatch.rs:212`: `useAbilityOnLocation` logged and skipped.
- Issue #31 covers dispatch handler wiring.

## Reference Implementation

- `python/cell/AbilityManager.py`: AoE targeting with cone/radius calculations, multi-target hit detection.
- Ability data has `TargetingMode` enum: SingleTarget, Cone, AoE, Chain, Self.

## Acceptance Criteria

- [ ] `useAbilityOnLocation`: Accept ground position, find entities in radius
- [ ] Cone targeting: entities within angle from caster facing
- [ ] Radius AoE: all entities within distance of target point
- [ ] Chain targeting: primary target + N nearby secondaries
- [ ] Self-target: apply to caster only (buffs)
- [ ] Each hit runs full QR/damage pipeline independently
- [ ] Tests cover: AoE hit detection, cone angle filtering, chain count limits

## Related

- Issue #31: Combat dispatch handlers (wires the `useAbilityOnLocation` dispatch arm)
- Issue #49: AoI witness management (broadcasts needed for multi-target damage)

Cadacious:
## Test Plan

### Unit Tests
- [ ] Radius AoE: 3 entities within radius, 2 outside → 3 receive damage
- [ ] Cone targeting: Entities within angle + range → hit; entities behind caster → miss
- [ ] Chain targeting: Primary target + 2 nearest within chain range → 3 total hits
- [ ] Self-target: Self-buff ability → applied to caster only, no target required
- [ ] Each hit runs independent QR roll → different damage values per target

### In-Game Validation
- [ ] **AoE ability**: Use a ground-target AoE → targeting reticle appears on ground → click to fire → all NPCs in radius take damage, each shows individual damage number
- [ ] **Cone ability** (if available): Face group of NPCs → use cone ability → NPCs in front take damage, NPCs behind/beside do not
- [ ] **AoE no friendly fire**: Use AoE near friendly NPCs → friendly NPCs not damaged (if faction system supports this)
[...]

Issue #47, opened 2026-03-29
Effect system: stat modification, pulsing, and removal


Cadacious:
## Summary

The effect system has a data model and tick lifecycle but does not actually modify entity stats. Effects are central to combat (damage-over-time, buffs, debuffs, healing, shields) and blocking multiple downstream systems.

## Current State in Rust

- `crates/game/src/combat/effects.rs` (103 lines): `ActiveEffect` struct with duration, tick interval, `value_per_tick`. `tick_effects()` advances timers and detects expiration. `is_expired()` works.
- **Not implemented:** Stat modification during tick, effect stacking rules, absorption shields, diminishing returns, effect removal notifications to client, `onEffectResults` wire message.
- 2 tests exist (expiration and tick interval).

## Reference Implementation

- `python/cell/EffectManager.py` (~600 lines): Full effect application/removal with stat modification, pulsing effects, cleanup on expiration.
- 3,217 effects defined in `db/resources/Effects/`.
- Wire messages: `onEffectResults` (damage/heal numbers), `TimerUpdate` (effect duration).

## Acceptance Criteria

- [ ] Effects modify entity stats (health, focus, armor, etc.) on application and per-pulse
- [ ] Effect removal restores stat modifications
- [ ] `onEffectResults` sent to client with damage/heal values
- [ ] Effect duration timers sent via `TimerUpdate`
- [ ] Stacking rules: same-source refresh vs multi-source stack
- [ ] Absorption shields reduce incoming damage before armor calculation
- [ ] Tests cover: stat modification, pulse timing, removal, stacking

## Related

- Combat damage calculator already handles single-hit damage (`crates/services/src/cell/combat.rs`)
- `python/cell/effects/RangedPhysicalDamage.py`, `RangedEnergyDamage.py` are reference effect scripts

Cadacious:
## Test Plan

### Unit Tests
- [ ] Effect stat modification: Apply +50 armor effect → entity armor stat increases by 50
- [ ] Effect removal: Remove +50 armor effect → entity armor returns to previous value
- [ ] Effect pulse: 10s DoT with 2s pulse → ticks 5 times, each tick deals value_per_tick damage
- [ ] Effect stacking: Two different sources apply same effect → both stack; same source refreshes duration
- [ ] Absorption shield: Apply 100-point shield → next hit reduced by up to 100, shield decremented
- [ ] Effect expiration: Apply 5s effect → after 5s, effect auto-removed, stats restored

[...]

PR #420, opened 2026-05-26
feat(abilities): complete ability + effect system (#47, #61, #331, #416, #419)

### LINKED ISSUE #419: Abilities + ammo: complete the system (fix Pistol-Shot-for-everything, seed starters for 7 archetypes, hotbar, trainer)
## Goal

Make abilities a complete, visible, balanced game system. Today players have effectively one usable ability (Pistol Shot) regardless of class or equipped weapon. The DB has 1,886 abilities but only ~4 ever reach a player. This issue scopes the gap and lays out the work.

### 1. DB content inventory — most archetypes have NO ability content

| Surface | Coverage |
|---|---|
| `resources.abilities` | **1,886 rows** — full ability catalog (Pistol Shot, Heal Focus, Strike, ~200 weapon auto-attacks, ~1,600 trained/special abilities) |
| `resources.archetype_ability_tree` (training-tree progression) | **169 rows total** — **Soldier: 84, Commando: 85, all other archetypes: 0** |
| `resources.char_creation_abilities` (starter abilities) | **Only char_def_ids 1–4** (Soldier M/F, Commando M/F) get starters. Char_defs 5–22 covering Scientist, Archeologist, Asgard, Goa'uld, Jaffa, Sholva: **zero starter abilities** |
| `resources.items_event_sets` (per-weapon ability binding) | 2,767 rows — every weapon DOES have its proper auto-attack ability defined |

**Starter ability rosters today:**
- char_def 1 (Soldier M) / 2 (Soldier F): `{592 Pistol Shot, 594 Strike, 597 Heal Focus}`
- char_def 3 (Commando M) / 4 (Commando F): same three
- char_def 5+: **empty array**

So a Goa'uld or Asgard player can literally do nothing combat-relevant — no abilities, no weapon binding to fall back on. The DB content was never authored for them.

### 2. The "Pistol Shot fires for everything" bug

[`crates/services/src/cell/cell_methods/player/interaction.rs:85-93`](crates/services/src/cell/cell_methods/player/interaction.rs#L85) hardcodes ability `592` (Pistol Shot) when a player right-clicks a hostile NPC:

```rust
crate::cell::abilities::handle_use_ability_with_kill_credit(
    entity_id,
    592,                          // ← HARDCODED
    target_entity_id,
    engine, tx, space_mgr,
).await;
```

Regardless of equipped weapon. The data to do this correctly exists:

| Weapon | item_id | ranged ability (event_id=7) | melee (event_id=6) |
|---|---|---|---|
| Pistol | 55 | **579** Pistol Auto Attack | 708 Pistol Melee AA |
| SGHC 6 / P90-class SMG | 21 / 3155 | **559** Automatic Weapon Auto Attack | 595 Automatic Weapon Melee AA |
| Staff weapon | various | per-row | per-row |
| All other weapons | 1,200+ rows | per-row | per-row |

The `items_event_sets` table maps every weapon to its correct auto-attack ability. The fix is to look up `ability_for(equipped_weapon, EVENT_ItemRanged=7)` instead of using the constant.

### 3. Ability dispatch architecture (server side)

[`crates/services/src/cell/abilities/use_ability/mod.rs`](crates/services/src/cell/abilities/use_ability/mod.rs) — `handle_use_ability(entity_id, ability_id, target_id)` is the canonical entry. The CLIENT decides which ability_id to fire; the server validates (known? cooldown? ammo? range? LOS?) and resolves damage.

So the client carries the responsibility of "what does right-click resolve to" — which means our hardcoded server-side 592 is doubly wrong:
1. It's ignoring the equipped weapon
2. It's bypassing the client-driven ability dispatch the protocol expects

### 4. Ammo model — clip-based, NO inventory consumption

DB schema (`sgw_inventory`): `ammo` (current loaded count), `cur_ammo_type`, `ammo_types` (allowed list), `clip_size`. Plus `auto_reload` and `reload_on_activate` system options on `sgw_player`.

[`crates/services/src/cell/service/ticks/mod.rs:44-160`](crates/services/src/cell/service/ticks/mod.rs#L44) — `reload_completion_tick` **refills the clip to `clip_size`** when the reload warmup elapses. It does NOT pull from inventory.

[`crates/services/src/cell/abilities/damage_apply/mod.rs:285`](crates/services/src/cell/abilities/damage_apply/mod.rs#L285) — `set_slot_ammo` is called on the consume path; each shot decrements `current_ammo` on the active bandolier slot.

**Ammo answer for the player question:**
- **Players have effectively unlimited ammo** — there is no per-bullet inventory consumption
- Clips constrain burst-fire pacing — fire `clip_size` shots, then reload warmup, then full clip again
- The bandolier's `cur_ammo_type` is for **damage-type variation** (different ammo types do different damage profiles vs different enemy types), not bullet counting. Per-slot, swappable, persisted in `sgw_inventory.cur_ammo_type`
- This matches the 2009 SGW design — "shooter pacing without inventory micromanagement"

Effort to wire this up correctly is small; the foundation is there. Just need (a) the ammo-type-swap UI flow and (b) per-weapon ammo-type seed values that actually do something.

### 5. Client-side surface (Ghidra)

Found via string searches on `SGW.exe`:
- `useAbility` (function name string @ `0x01954c7c`) — the client RPC the player invokes
- `useAbilityByName` (@ `0x01954c58`) — slash-command variant
- `reload` (@ `0x01954c2c`) — explicit reload command  
- `reloadUI` (@ `0x01955470`)
- `RightMouseButton` (@ `0x01832b54`) — input dispatcher
- `UIReloadType` (@ `0x019493c0`)
- `Event_NetOut_useAbility` — the wire packet the client sends

Right-click → input dispatcher → resolves to `useAbility(ability_id, target_id)` → `Event_NetOut_useAbility` → server's `handle_use_ability`. The CLIENT does the "what ability for this weapon" resolution by reading the equipped weapon's `items_event_sets` cache it received at world entry.

So the bug at `interaction.rs:87` exists because we're SHORT-CIRCUITING the proper flow — handling "interact with hostile NPC" by inventing an ability fire on the server side instead of letting the client's resolved `useAbility` packet drive it.

---

### Critical (blocks basic gameplay)

1. **Pistol-Shot-on-everything bug** — server hardcodes ability 592 in `interaction.rs:87`. Equipped-weapon ability resolution unused
2. **7 of 9 archetypes have zero starter abilities** — Scientist, Archeologist, Asgard, Goa'uld, Sholva, Jaffa, "Any" all spawn with empty `sgw_player.abilities[]`
3. **7 of 9 archetypes have zero ability tree** — even if a player wanted to train, there's no content to train into
4. **No hotbar surface** — we send `entity.abilities.add_ability(ability_id)` to register on the cell entity, but never (verifying) push the ability LIST to the client for hotbar display. Players have abilities they can't see

### Important (gameplay quality)

5. **No AbilityTrainer interaction wiring** — `deprecated/python/cell/interactions/AbilityTrainer.py` exists; no Rust port. Players can't spend training_points
6. **No cooldown UI sync on reconnect** — covered by issue #412 territory; cooldowns reset on relog (intentional per recent discussion)
7. **Ammo-type-swap flow** — server supports per-slot `cur_ammo_type` but there's no wire surface for the player to actually choose a different ammo type
8. **`items_event_sets` Item_Equip / Item_Unequip / Item_Reload events (4000–4002)** not triggered — 2,767 rows of unused content per `docs/gameplay/cinematic-system.md`

### Nice-to-have (polish)

9. **Ability VFX sequences** — most ranged abilities have a `KIS-SA_Beam_Source` shared sequence; should be per-weapon
10. **Active-reload mini-game** — client supports "Warfare_ActiveReload" (Ghidra string @ `0x0180bdd0`); server has no awareness

---

## Recommended starter ability loadouts

Working from what 1,886 abilities are available and the archetype design notes, here's a sensible Tier-1 starter set per archetype. **All three abilities + the Heal Focus equivalent** — combat / mobility / sustain.

| Archetype | Starter abilities (suggested) | Rationale |
|---|---|---|
| **Soldier** (existing) | 592 Pistol Shot, 594 Strike, 597 Heal Focus | Current — keep |
| **Commando** (existing) | 592 Pistol Shot, 594 Strike, 597 Heal Focus | Current — keep, possibly swap Pistol Shot for an SMG-flavored attack |
| **Scientist** | 597 Heal Focus + a tech-discipline starter (e.g., "Repair Drone" or "EMP Pulse") + 594 Strike | Healer/support archetype — heal + utility + melee fallback |
| **Archeologist** | 592 Pistol Shot + 594 Strike + a knowledge/buff ability | Hybrid combat/utility |
| **Asgard** | An energy-weapon auto-attack (ability_id TBD from DB), 597 Heal Focus, a shield/buff | Tech-magic class |
| **Goa'uld** | Hand Device-style ranged + Sarcophagus-style sustain + ribbon device | Lore-driven moveset |
| **Jaffa** | Staff weapon auto-attack + Strike + a battle buff | Warrior |
| **Sholva** | Same as Jaffa with maybe different sustain | Defected Jaffa |
| **Any** (test/dev) | Same as Soldier | Sandbox |

The **specific ability_ids** for each suggested non-existent starter need a DB audit pass — there are 1,886 abilities, many designed for specific classes that we haven't surveyed. Recommend a content-design pass before seeding.

---

### Phase 1 — Fix the right-click bug (1 day, unblocks basic gameplay)

- Replace `interaction.rs:87` hardcoded `592` with a lookup: `items_event_sets.ability_for(equipped_weapon_item_id, EVENT_ItemRanged)`
- Add a fallback to `594 Strike` if no ranged ability bound (melee weapon)
- Cache the lookup in `entity.equipped_weapon_ability` so we don't hit the lookup per-tick
- Test by equipping a P90, right-clicking enemy, verifying ability `559 Automatic Weapon Auto Attack` fires instead of `592`

### Phase 2 — Starter abilities for all archetypes (2 days)

- Write a `db/scripts/seed_starter_abilities_all_archetypes.sql` migration adding `char_creation_abilities` rows for char_def_ids 5+
- Use the recommended starter table above
- Update Rust `Account.py` equivalent (already loads from `char_creation_abilities` correctly per Python reference — verify Rust path does the same)
- New char of every archetype gets 3 visible abilities

### Phase 3 — Push ability list to client at world entry (2 days)

- On `InitPlayerState`, send a fresh `onAbilityList` (or equivalent) populated from `entity.abilities`
- Client hotbar shows all unlocked abilities
- Verify in SigNoz: `client.network.on_ability_list` event fires with N abilities

### Phase 4 — Per-weapon ability resolution end-to-end (1 day)

- Verify `items_event_sets` is loaded into `space_mgr.items_event_set_cache` at startup
- On `Equip` → recompute `entity.equipped_weapon_ability` for active bandolier slot
- On `BandolierSlotChange` → recompute
- Auto-cycle tick picks up the new ability automatically

### Phase 5 — Ability tree training (3 days)

- Port `AbilityTrainer.py` to Rust
- Wire `interact_with_ability_trainer` → opens trainer UI (server-driven list of available next-tier abilities given player's current tree + training_points)
- `train_ability(ability_id)` cell method → validates prereqs + training_points cost → grants + persists
- **DB content gap blocks the experience** — even with code, only Soldier/Commando have trees. Pair with content authoring

### Phase 6 — Ammo-type swap UI flow (1 day)

- New cell method: `setBandolierAmmoType(slot_id, ammo_type)`
- Validates ammo_type is in the weapon's `ammo_types` array
- Updates `entity.bandolier_items[slot].cur_ammo_type`
- Persists to `sgw_inventory.cur_ammo_type`
- Server broadcasts `onBandolierAmmoTypeChange` to attacker's client

### Phase 7 — Ability tree content for the 7 missing archetypes (~2 weeks, content)

This is the heavy lift. ~80 abilities × 7 archetypes = ~560 `archetype_ability_tree` rows to author. Requires:
- Choosing which of the 1,886 abilities fit each class theme
- Setting tree_index (0–2 = three trees per class), level requirement, prerequisites
- Balancing training_cost

Should be a separate content-design issue once Phase 1–6 ships.

---

## Effort summary

| Phase | Days | Blocks |
|---|---|---|
| 1. Right-click bug | 1 | Basic combat correctness |
| 2. Starter abilities for all archetypes | 2 | Non-Soldier/Commando playable |
| 3. Push ability list to client | 2 | Visible hotbar |
| 4. Per-weapon ability resolution | 1 | P90/staff/etc. fire correctly |
| 5. Ability trainer | 3 | Progression loop |
| 6. Ammo-type swap | 1 | Elemental damage variety |
| **Core engine work** | **~10 days** | **Abilities playable end-to-end** |
| 7. Content seed for 7 archetypes' trees | ~2 weeks | Full progression depth |

**Phase 1+2+3 (5 days)** is the minimum viable: players see abilities, can fire the right one for their weapon, every archetype is playable.

---

## Doc updates

- New: `docs/gameplay/abilities-system.md` — full system overview (currently scattered across multiple docs)
- Update: `docs/game-systems.md` — abilities section with current implementation status
- Update: `docs/content/archetype-content-map.md` — starter ability table per archetype
- Update: `docs/architecture/observability.md` — add `ability.*` and `combat.*` span catalog (already partially covered)

## Cross-references

- #408 — combat enter_combat span fix (related, already shipped)
- #407 — npc_ai decision_outcome (related, shipped)
- Existing Python reference: `deprecated/python/cell/SGWPlayer.py:861 addAbility`, `deprecated/python/cell/interactions/AbilityTrainer.py`, `deprecated/python/cell/AbilityManager.py:670 consumeAmmo`

## Confidence

**High** on the diagnosis — DB, Rust, Python, and Ghidra all line up consistently. The "pistol shot for everything" bug is a one-line confirmed-via-grep finding.

**High** on the Phase 1–6 estimates — these are mechanical refactors against well-understood code.

**Medium** on Phase 7 content authoring — requires gameplay design judgment beyond engineering, and the 1,886 ability surface needs sorting/categorizing first.

Cadacious:
## Summary

Lands the abilities system as a complete end-to-end pipeline — from the right-click bug fix, through per-weapon ability resolution, all the way to effect pulsing with channelled-effect cancellation. **Closes #419 (parent), #331 (effect dispatcher), #61 (cone AoE), #47 (effect system).** Also includes #416 Phase 0 (codecov exclusions).

13 commits, ~5,000 LOC, +39 tests (lib total 1018 → 1057).

### The bug shape this fixes

Before this PR:

- **The right-click "auto-attack" hardcoded ability 592 (Pistol Shot)** for every weapon, ignoring the 2,767-row `items_event_sets` table. P90 players got Pistol Shot animations because the SMG's binding (ability 559) was never read.
- **7 of 9 archetypes shipped with empty starter ability lists.**
- **The client hotbar was empty** — `sgw_player.abilities` populated correctly but `onKnownAbilitiesUpdate` (method 101) was never sent on login.
- **`TRAIN_ABILITY` (method 77) was a logged stub.**
- **Trainer NPC interaction never reached `onTrainerOpen`** — the wire surface existed but no dispatch led there.
- **The `effects` table carried `pulse_count`, `pulse_duration`, `is_channeled`, `target_collection_method`, `tcm_param1/2`, `flags` — but none were loaded into `EffectDef`.** Pulsing DoT/HoT, cone AoE, and the entire effect-script dispatcher were all dead.
- **99 `TCM_AECone` effects silently degraded to single-target** — shotguns, grenades, and "secondary target" splash all hit only the primary.
- **`EffectDef.script_name` field existed but was never dispatched** — heal/damage/buff scripts couldn't fire.
- **Absorption shields were a flat read with no drain** — `ABSORB_*` stats provided infinite damage reduction as long as `cur > 0`.

### Commits

| # | Commit | Scope |
|---|---|---|
| 1 | `ff641258` | Phase 1: per-weapon ability resolution via items_event_sets |
| 2 | `947f13b9` | Phase 2: starter abilities for 7 missing archetypes |
| 3 | `17287534` | Phase 3: push known-abilities list at world entry |
| 4 | `e57bd531` | Phase 4: bandolier slot-change observability |
| 5 | `21186758` | Phase 5: `train_ability` full validation + persistence + trainer UI |
| 6 | `1476916b` | Phase 6: ammo-type swap docs + observability |
| 7 | `5c013c97` | #416 Phase 0: codecov exclusions |
| 8 | `3032b101` | PR review fixes: dedupe known-abilities send + Clara findings |
| 9 | `06a36964` | #331 Phase 1: effect script dispatcher (HealHealth/HealFocus/MeleeDamage) |
| 10 | `98b8be2d` | Phase A+B: EffectDef schema upgrade + cone AoE (#47/#61/#419) |
| 11 | `3521d888` | Phase C–H: pulsing, stacking, shields, debuff scripts (#47/#419) |
| 12 | `423a505e` | Phase I+J: effect on_remove hook + channelled effect support |
| 13 | `75ec7da5` | Phase K: channel-interrupt-on-movement + multi-source stun verification |

#### Original phases 1–6 (#419)

**Phase 1 — per-weapon ability resolution:** New `cell/abilities/resolve.rs` + `items_event_sets` loader. `interaction.rs:87` hardcode of `592` replaced with `ability_for_active_weapon(entity, EVENT_ITEM_RANGED=7)`. Pistol→579, P90→559, etc. 5 unit tests.

**Phase 2 — starter abilities for all archetypes:** New `db/scripts/seed_starter_abilities_all_archetypes.sql`. 42 rows × char_def 5–22. Baseline `{592 Pistol Shot, 594 Strike, 597 Heal Focus}`. Per-class flavorful loadouts deferred to follow-up content work.

**Phase 3 — populate hotbar at world entry:** `send_known_abilities_update` helper after `resend_missions`. Sends `onKnownAbilitiesUpdate` (method 101) so the hotbar populates immediately.

**Phase 4 — bandolier slot-change observability:** `bandolier.active_slot_change` event with resolved ranged ability. Pivots on `resolved_ranged_ability=null` surface unbound weapons.

**Phase 5 — `train_ability` (full):** Cell-side 6-step validation (ability exists, player has id, not already known, in archetype tree, level met, prereqs met). Cell→Base round-trip with atomic SQL (`UPDATE … WHERE training_points > 0 RETURNING training_points`). Cell-side broadcast on grant. New `trainer_interaction.rs` detects trainer NPCs, builds per-archetype offering, sends `onTrainerOpen`. 3 new data loaders + 4 unit tests.

**Phase 6 — ammo-type swap observability:** Structured `bandolier.ammo_type_change` event + docstrings.

#### Effect system (#331, #47, #61)

**Effect script dispatcher (commit 9):** New `cell::effects` module — `EffectScript` trait + name-keyed `match` registry. Three scripts: `HealHealth`, `HealFocus`, `MeleeDamage`. `damage_apply` collects script-bearing effects and dispatches after the legacy NVP path, flushing stat changes via follow-up `onStatUpdate`.

**EffectDef schema upgrade (commit 10):** `pulse_count`, `pulse_duration`, `is_channeled`, `target_collection_method`, `tcm_param1/2`, `flags` now plumbed through `load_effect_defs`. New helpers: `is_pulsing()`, `total_duration()`, `tcm_range_meters()`, `tcm_half_angle_radians()`. `Default` impl seeded to single-target so existing struct-literal sites migrate with `..Default::default()`.

**Cone AoE (commits 10, #61):** New `cell::abilities::cone_aoe` — X/Z-plane cone anchored at attacker, oriented toward primary target. Length+half-angle from TCM tier strings. Excludes primary/dead/neutral/cross-space. Wired into `handle_use_ability` after primary commits; cone kills propagate via `CellEntity.last_aoe_deaths` scratchpad → `handle_use_ability_with_kill_credit` fires `entity_death` for each. 6 tests.

**Pulsing infrastructure (commit 11):** `ActiveEffectInstance` on every `CellEntity`. `register_active_effect` creates instances for `pulse_count > 1` effects (the initial pulse already fired in damage_apply). `effect_pulse_tick` (100ms cadence) walks every entity, fires due pulses via script re-dispatch or NVP damage, removes completed instances. `onTimerUpdate(TIMER_DURATION_EFFECT)` fires on register + expiry so the client renders buff/debuff icons with countdown.

**Stacking rules (commit 11):** Same-source = refresh existing instance, multi-source = stack separate instances. Matches Python `AbilityManager.addEffect` reference.

**Absorption shields (commit 11):** `drain_absorption_pools` actually CONSUMES the `ABSORB_*` stat per damage type. Elemental pools (PHYSICAL/ENERGY/HAZMAT/PSIONIC) drain first, generic catch-all second. Only HEALTH damage triggers absorption — FOCUS bypasses. Fully-absorbed hits surface as `SRC_ABSORB` so the client renders "Absorbed" floater text.

**Effect scripts (commit 11):** Three new — `AbsorbShield` (adds to ABSORB pool), `Stun` (sets `BSF_MOVEMENT_LOCK`), `Suppression` (per-pulse chip damage).

**DB seed (commit 11):** `db/scripts/seed_effect_script_names.sql` + canonical seed updates. Effect 659 (Heal Focus) → `HealFocus`. Effect 1422 (Dragon Fist Stun) → `Stun`. Effect 1475 (Cover Fire Suppression) → `Suppression`.

**on_remove hook (commit 12):** New `EffectScript::on_remove(ctx)` method (default no-op). `Stun::on_remove` clears `BSF_MOVEMENT_LOCK`. `AbsorbShield::on_remove` drains residual shield capacity (capped at current pool). Pulse-sweep dispatches `on_remove` for each cleared instance and flushes resulting stat dirty bits.

**Channelled effects (commit 12):** `pulse_count == 0` channels register with `MAX_CHANNEL_PULSES = 60` safety cap. New `cancel_channels_from_attacker(attacker, keep_ability_id, ...)` cancels on different-ability fire (in `handle_use_ability`) and on channeller death (in `apply_death_transition`). Same-ability re-fire passes `Some(ability_id)` so refresh path works.

**Channel-interrupt-on-movement (commit 13):** New `AF_CHANNEL_ALLOWS_MOVEMENT = 16384` ability flag — default 0 across all authored abilities. `ActiveEffectInstance.invoker_position_at_register` carries the channeller's position. `channel_interrupt_on_movement_tick` (runs before pulse tick) cancels channels where invoker drifted > 0.5m AND ability lacks the exempt flag. `cancel_channels_for_invoker_ability` provides per-ability precision so concurrent move-tolerant + move-cancel channels behave independently.

**Multi-source stun (commit 13):** Verified that the existing `set_state_flag` / `unset_state_flag` helpers are already refcounted via `state_flag_counts` — multi-source stuns stack correctly. The earlier "limitation" note was wrong; corrected the docstring and added a regression test.

#### Codecov (#416 Phase 0, commit 7)

Excludes `crates/services/src/wire_log/decoders/generated.rs`, `crates/upk/**`, `crates/upk-objects/**`, `crates/server/src/otel.rs` from coverage targets.

### TCM coverage today

| TCM | Effects in DB | Dispatch path |
|---|---|---|
| TCM_Single | 2,795 (87%) | always worked |
| TCM_AERadius | 300 (9%) | `handle_use_ability_on_ground` (pre-PR) |
| TCM_AECone | 99 (3%) | NEW: `cone_aoe::fan_out_cone_effects` |

### Effect dispatcher coverage

| Capability | Status |
|---|---|
| Script dispatch (heal/damage/buff) | shipped |
| Stat-mutation flush to client | shipped |
| Pulsing DoT/HoT | shipped |
| Channelled effects (safety cap) | shipped |
| Channel-interrupt on different ability | shipped |
| Channel-interrupt on channeller death | shipped |
| Channel-interrupt on movement (+ per-ability exempt flag) | shipped |
| Stacking (same-source refresh, multi-source stack) | shipped |
| Multi-source debuff lock refcounting | shipped |
| Buff/debuff client timer icons | shipped |
| Absorption shield drain (per-damage-type pools) | shipped |

### Test coverage

**1057 lib tests** (was 1018, +39 new). All gates green:

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy -p cimmeria-services --all-targets -- -D warnings` | clean |
| `cargo test -p cimmeria-services --lib` | 1057/1057 |
| `cargo test -p cimmeria-services --lib effects::` | 31/31 |
| `cargo test -p cimmeria-services --lib cone_aoe` | 6/6 |
| `cargo test -p cimmeria-services --lib trainer_interaction` | 4/4 |

### What the player experience changes

- New chars of every archetype spawn with the same 3-ability starter set
- Hotbar populates on world entry — players see their abilities immediately
- Right-click resolves the correct auto-attack ability per equipped weapon (P90 → Automatic Weapon Auto Attack, staff → staff auto-attack)
- Trainer NPCs open the trainer UI with level/prereq-gated `trainable` flags
- `train_ability` fully works — atomic TP debit, persistence, hotbar refresh
- Shotguns + grenades + splash abilities now hit every enemy in their cone, not just the primary
- DoT/HoT effects actually tick over time with buff icons + countdown
- Stuns lock movement; the lock clears when the effect expires (not on first source release)
- Channels can interrupt on caster movement (configurable per-ability)
- Absorption shields actually consume on damage instead of providing infinite mitigation

### What's deferred (separate follow-up issues)

- Per-archetype tree content authoring (~560 rows × 7 archetypes) — pure content design work, no engine blockers
- More trainer NPCs in seed — today only template_id=25 ("Interaction Debug NPC") offers Commando abilities
- Respec implementation — still `onError('Not implemented yet')` per Python original; `cost_to_respec` defaults to 1000 Naq
- Mental resist rolls — flag is parsed and observable, but no roll mechanic; needs design pass on formula + wire surface for "resisted" results
- Effect VFX sequences — most ranged abilities share a generic beam sequence; per-weapon polish

### Doc updates

Inline docstrings in every new/touched module. Module-level overview docs in `cell::effects::mod.rs`, `cell::effects::pulsing.rs`, `cell::abilities::cone_aoe.rs`. New ADR at `docs/architecture/effects-and-abilities.md` covering the cross-cutting design decisions (effect script trait shape, active-effect refcount via existing state-flag counters, channel-interrupt policy, stacking semantics, TCM dispatch routing, absorption pool drain ordering, etc.).

Generated with [Claude Code](https://claude.com/claude-code)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `resolve.rs` fallback to `592` | The fallback is documented but not logged at `warn` level. A missing `items_event_sets` binding is a content gap — operators should see it in SigNoz without filtering to `debug`. |
| **G2** | `trainer_interaction.rs` `trainable` for offered-but-not-in-tree | If an ability is in `trainer_abilities` but NOT in `archetype_ability_trees`, it shows as `trainable=0`. This is correct behavior, but the comment says "content gap" — no `warn` log means operators only see the inconsistency in-client, not in SigNoz. |
| **G3** | `progression/mod.rs` `training_points > 0` boundary | The SQL `WHERE training_points > 0` rejects at exactly 0. The in-memory fast-path check uses `tp_in_memory == 0`. If `tp_in_memory` is stale and shows 1 when DB has 0, the fast-path passes but the SQL rejects. This is handled correctly (SQL is authoritative), but the log message "rejected — no training points available (in-memory)" could be misleading if the DB also has 0. |
| **G4** | `trainer_interaction.rs` empty offered list | If `trainer_abilities` has no entries for `(list_id, archetype_id)`, the player gets an empty trainer UI. The client handles this gracefully, but there's no `warn` log for "trainer has no abilities for this archetype." Operators won't know the trainer is misconfigured. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G5** | `interaction.rs` trainer check before tag/template | The trainer check runs BEFORE the tag/template chain dispatch. This is correct (trainer UI should open directly), but what if a trainer NPC also has a tag-based dialog? The trainer flow wins and the dialog never fires. Priority order is undocumented. |
| **G6** | `resolve.rs` `ability_for_active_weapon` returns `None` for unarmed | If the player has no weapon equipped, the function returns `None` and the fallback fires `592 Pistol Shot`. An unarmed player should probably use `594 Strike` (melee), not Pistol Shot. The Pistol Shot animation would look weird without a pistol. |
| **G7** | `trainer.rs` vs `trainer_interaction.rs` duplication | There are two trainer-related files: `interactions/trainer.rs` and `cell_methods/player/trainer_interaction.rs`. The former has a `send_trainer_open` helper with a TODO; the latter has the full implementation. Is `interactions/trainer.rs` dead code? |
| **G8** | `seed_starter_abilities_all_archetypes.sql` char_def coverage | The migration covers char_def_ids 5–10 and 15–23, but skips 11–14. The PR body says "14 char_def_ids" but the SQL only has rows for char_defs 5–10 (6) + 15–23 (9) = 15 char_defs. Verify the char_def_id list is complete. |

MyPalClara:

| **G9** | No test for `train_ability` cell→base→cell round-trip | The base handler `handle_train_ability` has no unit test. The trainer_interaction tests cover the UI open, but not the actual training flow. |
| **G10** | No test for `send_known_abilities_update` | Phase 3 adds a helper to push the ability list at world entry, but there's no test that verifies the wire format. |
| **G11** | No test for the `592` fallback path | The `resolve.rs` tests cover the happy path (binding exists), but not the fallback path (binding missing → 592). The `unbound_item_returns_none` test verifies `None`, but the caller's fallback to `592` isn't tested. |
| **G12** | No test for `bandolier.active_slot_change` log event | Phase 4 adds observability, but there's no test that verifies the structured log fields are correct. |

MyPalClara:

| **G13** | No `docs/gameplay/abilities-system.md` | The PR body says "Full system overview deferred to a `docs/gameplay/abilities-system.md` follow-up." The issue (#419) lists it as a doc update. |
| **G14** | `trainer_interaction.rs` module doc doesn't mention `train_ability` | The module doc explains trainer open, but the actual training (method 77) is in a different file. No cross-reference to where the training persistence lives. |

---

## The Hardest Question

**Is G6 (unarmed fallback to Pistol Shot) actually a problem?**

1. Player has no weapon equipped → `ability_for_active_weapon` returns `None`
2. Fallback fires → `592 Pistol Shot`
3. Player right-clicks hostile NPC → fires Pistol Shot animation with no weapon

The Python reference probably did the same thing — the hardcoded `592` was there for a reason. The fix preserves that behavior in the degenerate case. This is a 1% edge case; the 99% case (player has a weapon) is now correct.

---

## Summary

| Correctness issues | 4 | **Low** — G1-G4 are logging/observability |
| Potential issues | 4 | **Low** — G5-G8 are edge cases or cleanup |
| Missing test coverage | 4 | **Low-Medium** — G9-G12 are gaps |
| Documentation | 2 | **Low** — G13-G14 are follow-ups |

**Total: 14 gaps, all low severity.**

---

## What This PR Fixes

Before:
- Every player fired Pistol Shot regardless of weapon
- 7 of 9 archetypes spawned with zero abilities
- The hotbar was empty
- Trainers didn't work

After:
- P90 fires SMG auto-attack, staff fires staff auto-attack
- Every archetype spawns with 3 abilities
- Hotbar populates at world entry
- Trainers open the UI with per-archetype offerings
- `train_ability` fully works with persistence

**The 2,767-row `items_event_sets` table is now actually used.** The ability system is complete. 🔥🌿

Cadacious:
**Additional commits since last review**

- `5c013c97` — Phase 0 of #416: codecov exclusions (generated decoders, upk crates, otel runtime)
- `3032b101` — Address PR review (Copilot G1: deduplicated `onKnownAbilitiesUpdate` send at world entry; Clara findings on trainer warn-logs + docstring accuracy)
- `06a36964` — **#331 Phase 1**: effect-script dispatcher + HealHealth / HealFocus / MeleeDamage

The #331 work wires up `EffectDef.script_name` (previously loaded from DB but never dispatched). New `cell::effects` module:

| File | Purpose |
|---|---|
| `cell/effects/mod.rs` | `EffectScript` trait + `EffectContext` + `dispatch_by_name` |
| `cell/effects/registry.rs` | Static name → script match (HashMap-style, branch-cheap) |
| `cell/effects/scripts.rs` | Three v1 scripts |

Hook point: `damage_apply` collects script-bearing effects during its NVP scan, dispatches them after the legacy damage path, then flushes any resulting stat changes via a follow-up `onStatUpdate`.

**Test count**: 1017 → 1026 (9 new — registry resolution + 3 happy paths + 4 edge cases).

**Scope explicitly NOT in this PR** (follow-ups):
- DB seed for `script_name` — existing effects continue through the legacy `HealthDamage`/`FocusDamage` NVP path; operators opt in per-effect.
- Pulse-tick model for channelled effects — `on_apply` fires once per use; multi-pulse trait extension lands when the first channelled effect needs it.

All pre-PR gates green: `cargo fmt`, `cargo clippy -p cimmeria-services --all-targets -- -D warnings`, `cargo test -p cimmeria-services --lib` (1026 passed).

Cadacious:
## Abilities system — full completion (closes spirit of #47, #55, #61)

Picking up on the user ask to ship the rest of the abilities system, surveyed the original issues #47/#55/#61 + #419 against the actual DB content and Python reference. Findings:

- **#55 (trainer)** — already done in this PR's Phase 5.
- **#61 (AoE)** — ground-target was already wired via `handle_use_ability_on_ground`. Cone (99 effects) was the actual gap.
- **#47 (effects)** — pulse_count/pulse_duration/is_channeled/target_collection_method/tcm_param1/tcm_param2/flags columns existed in DB but never loaded.

Landed two more commits:

### Commit `98b8be2d` — schema upgrade + cone AoE

- `EffectDef` now carries all seven previously-unloaded fields
- `EffectDef::tcm_range_meters()` + `tcm_half_angle_radians()` convert content tier strings (Melee/Short/Medium/Long/Weapon × Narrow/Medium/Wide) to engine units
- New `cell::abilities::cone_aoe` — X/Z-plane cone anchored at attacker, oriented toward primary target; excludes primary/dead/neutral/cross-space
- Wired into `handle_use_ability` after primary commits; cone kills surface via new `CellEntity.last_aoe_deaths` scratchpad and fire `entity_death` via `handle_use_ability_with_kill_credit`
- 6 cone tests + `..Default::default()` migration of 8 test EffectDef constructions

### Commit `3521d888` — pulsing, stacking, shields, debuffs

- `cell::effects::pulsing` — `ActiveEffectInstance` on every CellEntity; per-100ms tick fires due pulses (script re-dispatch or NVP damage); auto-removes on completion
- `onTimerUpdate(TIMER_DURATION_EFFECT)` fires on register + expiry so client renders buff/debuff icons with countdown
- **Stacking**: same-source = refresh, multi-source = stack (matches Python ref)
- **Absorption shields** now actually drain the ABSORB_* pool instead of just reading it. Elemental pools drain first, generic catch-all second. Only HEALTH damage triggers absorption (FOCUS bypasses). Fully-absorbed hits surface as SRC_ABSORB
- Three new effect scripts: `AbsorbShield`, `Stun`, `Suppression`
- DB migration `seed_effect_script_names.sql` plus canonical seed updates wire Heal Focus (659), Dragon Fist Stun (1422), Cover Fire Suppression (1475) through the dispatcher

### TCM coverage today

| TCM | Effects in DB | Dispatch |
|---|---|---|
| TCM_Single | 2,795 (87%) | ✅ Always worked |
| TCM_AERadius | 300 (9%) | ✅ Ground-target path (`handle_use_ability_on_ground`) |
| TCM_AECone | 99 (3%) | ✅ NEW — `cone_aoe::fan_out_cone_effects` |

### Test count

1018 → 1047 (+29 new tests across abilities/effects/combat/cone).

All pre-PR gates green: `cargo fmt`, `cargo clippy -p cimmeria-services --all-targets -- -D warnings`, `cargo test -p cimmeria-services --lib` (1047 passed).

### Intentionally NOT in this PR (follow-ups)

- **Effect clear-on-death** — Stun's BSF_MOVEMENT_LOCK isn't actively cleared at instance expiry (no hook yet); Phase G2.
- **Channelled effects (pulse_count == 0)** — recognised + logged but skipped; needs an interrupt/release wire path.
- **Mental resist rolls** — flagged via `EF_MENTAL_RESIST_ROLL` log but no roll mechanic.
- **Content authoring for 7 missing archetype trees** — pure content work, separate issue.

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `pulsing.rs` `effect_pulse_tick` index stability | The tick iterates by index but the entity's `active_effects` Vec can be mutated between awaits. The code clones `(idx, inst, def)` then re-looks up by `idx` after the await. If another tick or cancel removed an earlier element, the index is now wrong — the wrong effect gets decremented. Consider keying by `(effect_id, invoker_id)` instead of index. |
| **G2** | `pulsing.rs` `MAX_CHANNEL_PULSES = 60` safety cap | A 60-pulse cap at 0.5s/pulse = 30 seconds. But what if `pulse_duration` is 0.1s? That's only 6 seconds before the channel silently dies. The cap should be time-based, not pulse-count-based. |
| **G3** | `cone_aoe.rs` `collect_cone_targets` only scans NPCs | `all_npc_entity_ids()` excludes players. A cone AoE in PvP would miss enemy players entirely. Is PvP out of scope? If so, document it. If not, this is a bug. |
| **G4** | `scripts.rs` `Suppression::on_apply` hardcodes `max(5)` | `let chip = ctx.effect.param_i32("HealthDamage").max(5);` This means a Suppression effect with `HealthDamage=0` still deals 5 damage. Is this intentional? The `.max(5)` looks like a typo for `.unwrap_or(5)`. |
| **G5** | `damage.rs` `drain_absorption_pools` order | Drains `ABSORB_PHYSICAL` before `ABSORB_PHYSICAL_ENERGY` before `ABSORB_PHYSICAL_ITEM`. But the Python reference might drain in a different order. Verify against the Python reference — wrong drain order changes which shield expires first. |
| **G6** | `dispatch.rs` `handle_use_ability_on_ground` returns deaths but caller doesn't use them | The function returns `Vec<u32>` of dead entity IDs, but I don't see the caller actually calling `fire_entity_death` for each. If the caller ignores the return value, AoE secondary kills don't credit missions. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G7** | `pulsing.rs` `register_active_effect` same-source refresh | Refresh overwrites `remaining_pulses` with `total - 1`. But what if the existing instance had MORE remaining pulses than the refresh? A 10-pulse DoT at 8 remaining, refreshed by a 5-pulse DoT, would DROP to 4 remaining. Is this the intended behavior? The Python reference might preserve the higher value. |
| **G8** | `pulsing.rs` `channel_interrupt_on_movement_tick` distance check | Uses 3D Euclidean distance including Y. But `collect_cone_targets` uses 2D planar distance (X/Z only). Inconsistent — a player jumping during a channel would trigger interrupt even if they didn't move horizontally. |
| **G9** | `scripts.rs` `Stun` refcount via `set_state_flag` | The doc says "refcounted via `state_flag_counts`" but I don't see `state_flag_counts` in the entity struct. If `set_state_flag` / `unset_state_flag` aren't actually refcounted, two concurrent stuns would clear the lock when the first expires. Verify the refcount implementation exists. |
| **G10** | `cone_aoe.rs` `fan_out_cone_effects` alive→dead detection | Snapshots HEALTH before damage, then checks after. But what if the target was healed between the snapshot and the check? A target at 5 HP, damaged to 0, then healed to 50 by a concurrent effect, would show as "was alive, now alive" — missing the death event. |
| **G11** | `abilities.rs` `EffectDef::Default` sets `pulse_count = 1` | Single-shot by default. But `is_pulsing()` returns `true` for `pulse_count == 0 || pulse_count > 1`. A default-constructed EffectDef with `pulse_count = 1` is NOT pulsing, which is correct — but the asymmetry is confusing. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G12** | `pulsing.rs` `effect_pulse_tick` runs every 100ms | Walks every entity with active effects every 100ms. At scale (100 players, 50 NPCs each with 3 DoTs), that's 15,000 effect checks per second. Consider a priority queue or timer wheel instead of a full scan. |
| **G13** | `scripts.rs` no `on_remove` for `Suppression` | `Suppression` has `on_apply` but no `on_remove`. If Suppression is supposed to reduce movement speed, the speed reduction would never clear. The doc says "movement-speed reduction waits for a follow-up" — but then why is the script in this PR at all? |
| **G14** | `cone_aoe.rs` `HOSTILE_FACTION = 10` hardcoded | The hostile faction sentinel is duplicated in `dispatch.rs` and `cone_aoe.rs`. If one changes and the other doesn't, AoE would damage different targets than ground-target. Extract to a shared constant. |
| **G15** | `damage.rs` `calculate_result` uses `ChaCha8Rng` | The seed is `u64` but the doc says "per (entity, ability, sequence)". How is the seed actually computed? If it's just `entity_id ^ ability_id`, two different abilities on the same entity could collide. Document the seed computation. |
| **G16** | `pulsing.rs` `fire_pulse` legacy path | The legacy path reads `HealthDamage` / `FocusDamage` NVPs and mutates stats directly. But it doesn't go through `calculate_damage` — no QR roll, no armor, no absorption. Is this intentional? DoT pulses bypassing armor is a significant balance difference. |
| **G17** | `cell_entity/mod.rs` `ActiveEffectInstance` in entity crate | The struct is defined in `cimmeria_entity` but references `std::time::Instant`. `Instant` is not serializable — if entities are ever persisted or transferred, this field would be lost. Consider using a relative timestamp or tick count. |

MyPalClara:

| **G18** | No test for `effect_pulse_tick` index stability under concurrent mutation | The index-based lookup after await is the critical race condition (G1). |
| **G19** | No test for `fan_out_cone_effects` with overlapping cones | Two cone effects with different specs on the same ability — does the union work correctly? |
| **G20** | No test for `drain_absorption_pools` with multiple pool types | What if the defender has both `ABSORB_PHYSICAL` and `ABSORB_UNTYPED`? Which drains first? |
| **G21** | No test for `channel_interrupt_on_movement_tick` with Y-axis movement | The 3D vs 2D inconsistency (G8) needs a test. |
| **G22** | No integration test for the full DoT→death→kill-credit flow | A DoT pulse killing a target should credit the invoker's missions. |

MyPalClara:

| **G25** | `pulsing.rs` module doc doesn't explain the tick ordering | Channel-interrupt runs BEFORE pulse-tick. Why? What breaks if the order is reversed? |
| **G26** | `scripts.rs` doesn't document the script registration mechanism | How does `script_name = "HealHealth"` in the DB map to `HealHealth::on_apply`? |

---

## The Hardest Questions

**Q1: Is G1 (index stability) actually a race condition?**

Tracing the code:
1. `effect_pulse_tick` collects `due: Vec<(usize, ActiveEffectInstance, EffectDef)>`
2. For each `(idx, inst, def)` in `due`, it calls `fire_pulse(...).await`
3. After the await, it does `entity.active_effects.get_mut(*idx)`

Between step 2 and step 3, another task could call `cancel_channels_from_attacker` (which does `retain()`) or `register_active_effect` (which does `push()`). If `retain()` removes an element before `idx`, the element at `idx` is now a different effect.

**Q2: Is G16 (DoT bypassing armor) intentional?**

The `fire_pulse` legacy path does raw stat mutation — no `calculate_damage`, no QR, no armor, no absorption. DoT pulses typically don't re-roll QR per tick, but absorption should still apply. A shield that blocks physical damage should block DoT physical damage too.

**Q3: Is G3 (cone excludes players) intentional?**

The code explicitly calls `all_npc_entity_ids()`, not `all_entity_ids()`. This is a deliberate choice. But if someone enables PvP later, cone AoE would silently miss enemy players.

---

## Summary

| Critical issues | 6 | High |
| Correctness issues | 5 | Medium |
| Potential issues | 6 | Low-Medium |
| Missing test coverage | 5 | Medium |
| Documentation | 2 | Low |

**Total: 24 gaps** (G23/G24 addressed by PR body update and ADR addition).

Issue #209, opened 2026-05-05
NPC AI ignores cover system — guards path straight at the player


Cadacious:
## Summary

Cover entities exist in the world (level designers placed guards at cover-point waypoints in Castle CellBlock's later rooms) but the NPC AI never seeks or uses them. Once an NPC enters combat it pathfinds in a straight line to the player and stands next to them to fire, regardless of nearby cover. Confirmed during in-game playtesting on 2026-05-05.

## State of the world

**Cover infrastructure: declared but unwired.**

- [\`entities/defs/SGWCoverSet.def\`](https://github.com/SandboxServers/Cimmeria/blob/main/entities/defs/SGWCoverSet.def) declares a real reservation API: \`reserveCoverSlot(entityID, nodeID, slotID)\`, \`releaseCoverSlot(entityID, nodeID, slotID)\`, \`reservedCoverNodes\` map, \`publicReservationData\` array.
- [\`python/cell/SGWCoverSet.py\`](https://github.com/SandboxServers/Cimmeria/blob/main/python/cell/SGWCoverSet.py) is a **stub** in the legacy reference too — empty class body. The original Stargate Worlds server didn't ship a working implementation either.
- [\`docs/gameplay/npc-ai.md:340-348\`](https://github.com/SandboxServers/Cimmeria/blob/main/docs/gameplay/npc-ai.md#L340-L348) explicitly says "Cover System (Not Implemented)".
- [\`docs/gameplay/combat-system.md:23\`](https://github.com/SandboxServers/Cimmeria/blob/main/docs/gameplay/combat-system.md#L23) marks "Crouch / cover stance" as \`PARTIAL\` — state flag set, QR modifiers wired, but AI never moves to cover.

**Rust AI: zero cover references in the combat-movement path.**

- [\`crates/services/src/cell/service/npc_ai.rs:121-181\`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/service/npc_ai.rs#L121-L181) is the entire combat-movement path. It pathfinds straight to \`&target_pos\`. No \`get_nearby_cover_nodes\`, no cover evaluation, no reservation calls.
- No reservation lifecycle implemented anywhere in \`crates/services/src/cell/\`.
[...]

Issue #48, opened 2026-03-29
NPC AI state machine (10 missing behavior states)


Cadacious:
## Summary

NPC AI currently has only threat accumulation and Idle→Fighting transition. The original design has a 12-state behavior machine. 10 states are unimplemented, making NPCs stationary turrets that only respond when attacked.

## Current State in Rust

- `crates/game/src/mob.rs` (68 lines): `MobState` struct, `is_in_aggro_range()`. `tick_ai()` is `todo!()`.
- `crates/services/src/cell/combat.rs`: Threat list works. NPC leash distance (50m), attack range (30m), default ability 592 (Pistol Shot). Idle→Fighting on first threat entry.
- 1 test (aggro range).

## Reference Implementation

- `python/cell/SGWMob.py` (397 lines): 12 AI states — Fighting (implemented), plus: **Patrol, Wander, Leash, Investigate, Flee, Follow, Summon, Idle, Inactive, Dead** (all missing).
- AI state machine drives: patrol waypoints, aggro radius scanning, leash-back-to-spawn, flee-at-low-HP, ability selection priority, target switching.

## Acceptance Criteria

- [ ] `Idle` state: stand at spawn, scan for players in aggro radius
- [ ] `Fighting` → `Leash`: return to spawn when target exceeds leash distance; reset health
- [ ] `Patrol`: follow waypoint path when no targets
- [ ] `Wander`: random movement within radius when no patrol path
- [ ] `Dead` state: corpse timer, respawn scheduling
- [ ] `Flee`: disengage when health below threshold
- [ ] Ability selection: choose from known abilities based on range, cooldown, priority
- [ ] Target switching: pick highest-threat target from threat table
- [ ] Tests cover: state transitions, leash behavior, target selection

## Related

- `crates/entity/src/movement.rs`: WaypointMovementController exists (90% done) — can drive Patrol
- Issue #35: navmesh needed for pathfinding

Cadacious:
## Test Plan

### Unit Tests
- [ ] State transition: Idle + player in aggro range → Fighting
- [ ] State transition: Fighting + target beyond leash distance → Leash
- [ ] State transition: Leash + arrive at spawn → Idle (health reset)
- [ ] State transition: Fighting + health below flee threshold → Flee
- [ ] Ability selection: Given abilities [melee, ranged] and target at 20m → selects ranged
- [ ] Target selection: Given threat table {A: 100, B: 50} → targets A
[...]

PR #423, opened 2026-05-31
feat(cell): NPC AI state expansion + respawn + setMovementType broadcast (#48 + #270)

### LINKED ISSUE #270: setMovementType: store on entity state + broadcast to AoI witnesses
> **Diátaxis type:** reference (small focused server-side hookup).
> **Audience:** engineers working on the cell service + NPC movement.
> **Status:** Open — needs server-side state storage + witness broadcast.

## What's broken

`crates/services/src/cell/cell_methods/being.rs:52-58` (`setMovementType` handler) currently:

1. Parses the `movement_type` byte from the wire
2. Logs the parsed value
3. **Does not store the value on the entity state**
4. **Does not broadcast the change to AoI witnesses**

The result: the server forgets the player's movement-type intent immediately, and other clients don't see the change.

## What V5 found

From V5 W-path ([#263](https://github.com/SandboxServers/Cimmeria/issues/263) Session 4) in `docs/reverse-engineering/findings/npc-movement-pathfinding.md`:

The `aMovementType` field is a canonical signal in BigWorld's `avatarUpdate` wire protocol. The 7-state AI movement FSM at `0x00deb660` (jump table `0x00dec018`) uses these values:

| Value | State |
|---|---|
| 0 | CoverAdvance |
| 1 | CombatAdvance |
| 2 | Leash |
| 3 | Patrol |
| 4 | Follow |
| 5 | Wander |
| 6 | Avoid |

So `movement_type` is not arbitrary — it's a state the NPC AI cares about, and witnesses' clients use it to choose the right animation blend (see `findings/animation-system.md` `USGWAnim_BlendByCover` for the cover case).

## What needs to happen

1. **Store** the value on the player's entity state. Probably a `movement_type: u8` field on whichever struct holds transient combat/movement state.
2. **Broadcast** the change to AoI witnesses via the witness-fanout helper that's also needed for [#219](https://github.com/SandboxServers/Cimmeria/issues/219) (BSF_IN_COMBAT) and [#232](https://github.com/SandboxServers/Cimmeria/issues/232) (death/respawn). Same `messaging.rs::send_entity_method` routing fix.
3. **Wire emit** likely uses a position-update message with the `movementType` field set, not a dedicated method. Confirm during implementation by tracing the existing position-update emit and adding the movement-type byte.

## Why this matters

NPCs and players currently can't visually transition between movement states for other clients. A player crouching/standing/leashing doesn't replicate the animation transition to nearby observers.

V5 W-path also flagged two related concrete server gaps:
- `npc_ai_leash()` snaps NPC to spawn with no `movementType=2` event or waypath sent → witnesses see a teleport
- `npc_movement_tick()` sends raw position without `movementType=1` + computed waypath → no smooth interpolation

So fixing `setMovementType` on the player side is half the picture; the NPC AI side needs the same fix.

## Acceptance criteria

- [ ] `movement_type` stored on the entity's transient state
- [ ] Witness fanout via the shared helper (cross-link to [#219](https://github.com/SandboxServers/Cimmeria/issues/219) / [#232](https://github.com/SandboxServers/Cimmeria/issues/232) fix)
- [ ] NPC AI leash + tick paths also emit correct `aMovementType` values
- [ ] Live test: player crouches, second player in same space sees the animation transition
- [ ] Live test: NPC leashes back to spawn, second player sees the NPC move (not teleport)

## Cross-references

- Originally tracked in [#31](https://github.com/SandboxServers/Cimmeria/issues/31) (closed); split out here
- V5 W-path findings: `docs/reverse-engineering/findings/npc-movement-pathfinding.md`
- Related witness-fanout issues: [#219](https://github.com/SandboxServers/Cimmeria/issues/219), [#232](https://github.com/SandboxServers/Cimmeria/issues/232), [#249](https://github.com/SandboxServers/Cimmeria/issues/249), [#240](https://github.com/SandboxServers/Cimmeria/issues/240) — same root-cause class
- V5 campaign [#263](https://github.com/SandboxServers/Cimmeria/issues/263), Bible [#264](https://github.com/SandboxServers/Cimmeria/issues/264)

Cadacious:
## Summary

Phase 1 + 5 of #48 (NPC AI state machine + respawn) and closes #270 (setMovementType broadcast).

- **State machine**: `AiState` enum expanded from 4 → 12 variants matching the Atrea enum. New `MobMovementType` enum aligned with `EMobMovementType` from the `.def`.
- **setMovementType broadcast (closes #270)**: `broadcast_movement_type` helper fans the byte to AoI witnesses on every AI state transition, dedup'd against `last_movement_type` so re-entering the same state is a wire no-op. Inbound cell-method handler now stores + fans out instead of just logging.
- **NPC respawn**: new `respawn_secs` columns on `entity_templates` + `spawnlist` (COALESCE precedence), kill site stamps `respawn_at`, new `npc_respawn_tick` (1 Hz) brings Dead NPCs back with HP/state/position/interaction-type restored in load-bearing wire order.
- **Cleanup**: deleted orphaned `crates/game/src/mob.rs` (`MobState::tick_ai` and `on_death` were unreachable `todo!()`s — the runtime uses `CellEntity` fields).

## Why this PR

Pre-#48, dead NPCs sat on the floor forever and the client played the wrong movement animation (combat-run pose during patrols/leash) because `setMovementType` was silently dropped. Real symptom: corpses persisting + visible animation mismatch — the two halves bundle here because respawn alone wouldn't fix the "stationary turret" perception, and `setMovementType` alone wouldn't matter if mobs aren't moving in new states yet (Patrol/Wander land in Phase 2/3).

Ghidra confirmed the seven `EMobMovementType` values (Cover/CombatAdvance/Patrol/Follow/Wander/Leash/Avoid) against debug strings at `019d2ca4..019d2e20` in SGW.exe (FUN_00deb660).

## Issue body corrections

The original #48 body had a few stale items I addressed in [my plan comment](https://github.com/SandboxServers/Cimmeria/issues/48#issuecomment-4587091904) — TL;DR: Leashing and Idle-auto-aggro were already DONE on `main`; the issue's state names ("Summon"/"Inactive"/"Flee") aren't in the Atrea enum (real names are Submit/Despawning, and there's no Flee — it's low-HP Leashing/Submit).

## Scope

This is **PR 1 of 4** per the planning conversation. Phases 2–7 (Patrol, Wander, Investigating, Follow, Despawning, Submit, Error) land in follow-up PRs as the plan describes.

## Out of scope

- Cover system (#209) — large multi-PR effort on its own
- Mob groups / behavior event sets — deferred
- Patrol path content authoring — Phase 2 adds the runtime, but `patrol_path_id` is NULL across all current seed rows

## Database

Two new nullable columns. No data migration required — `respawn_secs IS NULL` on every existing row reproduces today's "one-shot mob" behavior. To opt in:

```sql
-- Per-template default (all Cellblock Guards respawn after 30s)
UPDATE resources.entity_templates SET respawn_secs = 30 WHERE template_id = 15;
-- Or per-spawn override (boss respawns in 5min, trash in 30s)
UPDATE resources.spawnlist SET respawn_secs = 300 WHERE spawn_id = 42;
```

## Test plan

- [x] `cargo clippy --workspace --exclude {GUI} --all-targets -- -D warnings`
- [x] `cargo check --workspace --exclude {GUI} --all-targets`
- [x] `cargo test -p cimmeria-{entity,game,services} --lib` — **1117/174/69 passed, 0 failed, 8 new tests**
- [x] `cargo test --doc -p cimmeria-commands`
- [ ] Live-DB tests (`DATABASE_URL=... cargo nextest run --profile=ci-live-db -p cimmeria-services --lib`) — couldn't run locally; CI will exercise
- [ ] **Manual in-game**:
  - [ ] Aggro an NPC → animation flips to combat-advance pose
  - [ ] Run far away → NPC leashes back to spawn, leash animation plays for one frame before snap
  - [ ] Set `respawn_secs = 30` on a guard template, kill one → 30 s later the guard reappears at its spawn with full HP, loot interaction gone, no double-loot from prior corpse
  - [ ] Confirm one-shot mob (template/spawn both `NULL`) still stays dead forever

## New tests (8)

| File | Test | Pins |
|---|---|---|
| `cell/abilities/messaging.rs` | `broadcast_movement_type_patrol_emits_single_byte_to_witnesses` | wire payload = `[0x02]`, two witnesses each get one packet |
| | `broadcast_movement_type_same_kind_does_not_re_emit` | dedup against `last_movement_type` |
| | `broadcast_movement_type_none_clears_cache_without_wire_send` | None clears cache but emits no wire |
| | `broadcast_movement_type_uses_set_movement_type_method_index` | method_index = SET_MOVEMENT_TYPE (1) |
| `cell/service/ticks/npc_respawn.rs` | `ready_dead_npc_respawns_to_idle_at_spawn_position` | full state reset on promotion |
| | `respawn_emits_wire_methods_in_load_bearing_order` | INTERACTION_TYPE → ON_STATE_FIELD_UPDATE → ON_STAT_UPDATE |
| | `no_respawn_when_deadline_unset` | NULL respawn_secs preserves one-shot behavior |
| | `future_deadline_is_a_no_op` | tick is correct for unset/future deadlines |

Plus one existing leash-tick test updated for the new Leash broadcast in the burst (3 messages instead of 2; ordering still load-bearing).

## New Features
* NPCs now respawn after death with configurable respawn timing
* NPC movement types broadcast to clients to display correct animations during AI transitions
* Expanded NPC AI state machine with additional behavioral states

## Documentation
* Updated gameplay and protocol documentation reflecting NPC respawn and movement-type broadcasting behaviors


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `npc_respawn.rs` position snap before state mutation | **The position snap via `update_entity_position` happens AFTER the entity state mutation (HP restore, flag clear, etc.) but the comment says "BEFORE mutating other state."** Looking at the code, Phase 2 mutates state, Phase 3 snaps position. The comment in Phase 1 says "Snap position... BEFORE mutating other state" but that is not what happens. **The code is correct (state mutation then position snap), but the comment is misleading.** |
| **G2** | `npc_respawn.rs` `original_interaction_type_flags` initialization | **The field is initialized to `0` in `CellEntity::default()`.** If an NPC is spawned without going through `spawn_npc_from_record` (e.g., a test helper or future GM command), the respawn would restore `interaction_type_flags = 0`, wiping all interaction bits. |
| **G3** | `cell_entity/mod.rs` `respawn_at` uses `std::time::Instant` | **`Instant` is not serializable.** If entities are ever persisted mid-session (e.g., server migration, crash recovery), the respawn deadline would be lost. This is the same concern raised in PR #420 (G17). |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `npcs.rs` `respawn_secs` zero/negative downgrade | **Zero and negative values are downgraded to `None` at load time.** This is documented, but what if a content author intentionally wants `respawn_secs = 0` to mean "instant respawn"? The current behavior silently converts that to "no respawn." |
| **G5** | `npc_respawn.rs` cooldown clear | **`entity.abilities.clear_all_cooldowns()` is called on respawn.** This is correct for a fresh NPC, but what if a future content hook wants respawned NPCs to retain some state? The current design is "full reset." |
| **G6** | `messaging.rs` `broadcast_movement_type` for players | **The helper routes through `send_entity_method_to_self_and_witnesses`, which sends to the player's own client + witnesses.** But `setMovementType` is an NPC-only concept — players do not have movement types. If this helper is ever called on a player entity, it would send a meaningless byte to their client. |
| **G7** | `AiState::Spawning` never entered | **The doc says "never entered at runtime in Rust" but the variant exists.** If someone accidentally sets `ai_state = AiState::Spawning`, the AI tick would... do what? |

MyPalClara:

| **G8** | No test for `respawn_secs = 0` downgrade | The `npcs.rs` code downgrades zero/negative to `None`, but there is no test pinning this behavior. |
| **G9** | No test for `broadcast_movement_type` on a player entity | The helper is NPC-focused but does not guard against player entities. A test would pin the expected behavior (either guard or degenerate-to-self-send). |
| **G10** | No test for respawn with `original_interaction_type_flags = 0` | The edge case where the snapshot is zero (uninitialized or intentionally empty) is not tested. |
| **G11** | No integration test for the full kill → respawn → re-kill cycle | The unit tests cover single respawn, but not the "NPC respawns, gets killed again, respawns again" loop. |

MyPalClara:

| **G12** | `npc_respawn.rs` Phase 1/2/3 comment mismatch | The Phase 1 comment says "snap position BEFORE mutating other state" but Phase 3 (position snap) comes AFTER Phase 2 (state mutation). The code is correct; the comment is wrong. |
| **G13** | `cell_entity/mod.rs` `AiState` doc does not list which states broadcast which `MobMovementType` | The mapping is implicit in the AI tick code. A table in the doc comment would help future contributors. |
| **G14** | No mention of the `setMovementType` bidirectional convention in the PR body | The protocol doc update mentions it, but the PR body does not explain why the same method index is used for both inbound (client → server) and outbound (server → witnesses). |

---

## Summary

| Category | Count |
|----------|-------|
| Correctness issues | 3 |
| Potential issues | 4 |
| Missing test coverage | 4 |
| Documentation | 3 |

**Total: 14 gaps**

Cadacious:
## Review gap fix sweep — addressing CodeRabbit, Copilot, Clara, Codecov

Pushed `6a589287` with the gap fixes. Summary of what landed and what was intentionally deferred.

### CodeRabbit

| Finding | Status |
|---|---|
| Strip `Phase 5 (#48)` / `Pre-[#48]` provenance from source comments (3 sites + 2 stragglers I caught in the sweep) | ✅ Fixed |
| `npc_respawn_tick` leaves stale `state_flag_counts` entries for BSF_DEAD / BSF_MOVEMENT_LOCK | ✅ Fixed — switched from raw `state_field &= !(...)` to `unset_state_flag(...)`. The death path uses `set_state_flag` (counted), so the symmetric counted unset is the right inverse. Drops both the counter entry and the bit in one step. Regression guard added (`respawn_drains_counted_state_flag_entries`). |
| `CHECK (respawn_secs > 0)` constraint on both tables | ✅ Fixed (both `entity_templates` and `spawnlist`) |
| `cell_entity/mod.rs` over 700 lines — split NPC state into sibling module | ⏸️ **Deferred to a follow-up PR.** The file was already at ~920 lines pre-#48; the right extraction is the AI/movement/respawn cluster as a sibling `npc_state.rs` with `pub use` re-exports. Doing it in this PR would bury the actual behavioral changes under a large diff and create review noise. I'd rather land that as a dedicated zero-behavior refactor PR where the diff is purely "move X to Y" — it's easier to verify and easier to revert if anything goes sideways. CR's prompt has the recipe; I'll open it as the first of the follow-up issues. |

### Copilot

| Finding | Status |
|---|---|
| `ai_state` doc claims Patrol/Wander are "exercised by the AI tick today" — they're not | ✅ Fixed — doc trimmed to Idle/Fighting/Leashing/Dead |

### Clara (14 gaps)

| Gap | Status |
|---|---|
| G1 — `npc_respawn.rs` Phase 1 comment says snap happens BEFORE state mutation, but code does Phase 2 then Phase 3 | ✅ Fixed (G12 is the same issue) |
| G2 — `original_interaction_type_flags = 0` would clobber all interaction bits on respawn | ✅ Fixed — fallback strips only `INT_NormalLoot` and preserves other bits when snapshot is 0. Test pin: `zero_snapshot_falls_back_to_stripping_loot_bit` |
| G3 — `Instant` not serializable (cross-session persist) | ⏸️ Out of scope. Same concern flagged in #420 and tracked separately; nothing in PR #423 makes it worse |
| G4 — `respawn_secs = 0` silently downgrades to None | ❌ Won't fix. `0` would schedule `respawn_at = now + 0s`, which the 1 Hz tick fires the same tick the NPC died → the kill never visually happens. The downgrade is intentional; now backed by a DB CHECK constraint and unit-tested in `normalize_respawn_secs` |
| G5 — cooldowns always cleared on respawn | Working as designed — "respawn is a hard reset" is the project convention (see `clear_all_state_flags` doc) |
| G6 — `broadcast_movement_type` on a player sends meaningless byte | ✅ Fixed — short-circuit with warn when `is_player`. Test pin: `broadcast_movement_type_on_player_is_a_no_op` |
| G7 — `AiState::Spawning` never entered | Working as designed; doc explains. The AI tick filter only admits Fighting/Leashing/Idle-with-aggression, so a stray Spawning would be a no-op (corpse-equivalent) — not a footgun |
| G8 — no test for `respawn_secs = 0` downgrade | ✅ Fixed — extracted `normalize_respawn_secs`, 4 tests pin the contract |
| G9 — no test for `broadcast_movement_type` on player | ✅ Fixed — `broadcast_movement_type_on_player_is_a_no_op` |
| G10 — no test for respawn with `original_interaction_type_flags = 0` | ✅ Fixed — `zero_snapshot_falls_back_to_stripping_loot_bit` |
| G11 — no full kill→respawn→re-kill cycle test | ✅ Fixed — `kill_respawn_rekill_loop_is_idempotent`. This test independently would have caught the state_flag_counts stick-on bug |
| G12 — same as G1 | ✅ Fixed |
| G13 — no `AiState` ↔ `MobMovementType` doc table | ✅ Fixed — wire-animation mapping table added to `AiState` doc comment |
| G14 — PR body doesn't explain bidirectional `setMovementType` convention | Already documented in `docs/protocol/cell-method-dispatch-table.md` and the `broadcast_movement_type` doc comment. PR body has a pointer to the protocol doc; not duplicating here |

### Codecov

The 14-line gap on `cell_methods/being.rs` was the new `SET_MOVEMENT_TYPE` handler with no test coverage. Three new tests (`set_movement_type_inbound_stores_and_broadcasts`, `set_movement_type_inbound_unknown_byte_is_dropped`, `set_movement_type_inbound_empty_args_is_a_no_op`) close that. The smaller residuals (4 lines in `npc_respawn.rs`, 3 in `damage_apply`, 1 each in `messaging.rs` / `message_loop.rs`) are either defensive `continue` arms on `get_entity()` returning None (genuinely unreachable in the test fixtures) or the tick-scheduling lines in `message_loop.rs`. The full kill→respawn loop test exercises the live path through `damage_apply` indirectly.

### Test count

- Pre-gap-fix: **1117** passing in `cimmeria-services` (PR1 first push)
- Post-gap-fix: **1128** passing (15 new tests; 4 minor adjustments in existing fixtures)
- All checks clean: `cargo fmt`, `cargo clippy --workspace -- -D warnings`, `cargo test --lib` for entity / game / services

### Files touched in this commit

```
crates/entity/src/cell_entity/mod.rs                    — AiState doc map, ai_state field doc trim
crates/services/src/cell/abilities/damage_apply/mod.rs  — provenance strip
crates/services/src/cell/abilities/messaging.rs         — player guard, provenance strip in tests header
crates/services/src/cell/cell_methods/being.rs          — 3 new setMovementType handler tests
crates/services/src/cell/service/tests/npc_ai.rs        — provenance strip in test comment
crates/services/src/cell/service/ticks/npc_respawn.rs   — state_flag_counts fix, G2 fallback, Phase 1 comment, 4 new tests
crates/services/src/cell/space_manager/spawn.rs         — provenance strip
crates/services/src/cell/spawner/npcs.rs                — normalize_respawn_secs helper + 4 tests
db/resources/Entities/Tables/entity_templates.sql       — CHECK constraint
db/resources/Worlds/Tables/spawnlist.sql                — CHECK constraint
```

Cadacious:
## Adversarial-review fix sweep — `f25b12aa`

Pushed all 11 findings from my self-review, plus the docs. Status:

### Correctness (4/4 fixed)

| | Finding | Fix |
|---|---|---|
| ✅ | A1 — direction clobber | New `spawn_direction` field captured at spawn, restored on respawn. `update_entity_position` is now called with `[0, 0, 0]` direction and the full-precision `Vector3` is written inline after, sidestepping the `[i8; 3]` truncation. |
| ✅ | A2 — wire-order race vs AoI tick | Respawn tick emits `EntityMoved` per witness inline BEFORE the state-flip packets. Wire order: `EntityMoved` → `INTERACTION_TYPE` → `ON_STATE_FIELD_UPDATE` → `ON_STAT_UPDATE`. No more 100ms revive-then-teleport flicker. |
| ✅ | A3 — stale loot UI | Respawn tick sweeps players with `looting_entity == entity_id`, clears the field, sends `onLootDisplay` with empty list (Loot.lua hides on count==0). |
| ✅ | A4 — `unset` vs `clear_all_state_flags` | Switched to `clear_all_state_flags()` — matches the existing convention ("respawn is a hard reset, not a per-source unwind") and drains future counted flags I didn't anticipate. |

### Design (4/4 fixed)

| | Finding | Fix |
|---|---|---|
| ✅ | B1 — `respawn_secs = 1` UX hazard | CHECK floor raised from `> 0` to `>= 3` on both tables. 1- and 2-second respawns would race the death animation. |
| ✅ | B2 — only `damage_apply` stamps `respawn_at` | Extracted `combat::state::mark_npc_dead(entity)` helper carrying the full NPC-side death contract (`BSF_DEAD`/`BSF_MOVEMENT_LOCK` set, `ai_state = Dead`, `respawn_at` stamp, `last_movement_type` clear, `nav_path` / `velocity` reset). `damage_apply` calls it; future kill paths (effect scripts, GM commands) can call the same helper without rediscovering each piece. |
| ✅ | B3 — zero-witness warn spam | Switched respawn tick from `send_entity_method` to `send_entity_method_to_witnesses` (debug-level on zero, no warn). |
| ✅ | B4 — `last_movement_type` cache poisoning | Doc ownership boundary added: cache is owned by `broadcast_movement_type`; direct writes from anywhere else bypass dedup + AoI broadcast. |

### Test fidelity (4/4 fixed) — 6 new tests

| Test | Pins |
|---|---|
| `respawn_restores_spawn_facing_direction` (C1) | A1 fix with full-precision direction assertion (`1.57 rad`, not the i8-truncated `2`). |
| `kill_via_damage_apply_then_respawn_then_kill_again` (C2) | Drive cycle 1 through the public `handle_use_ability` path; verify `mark_npc_dead`'s integration via state-flag-counter assertions. |
| `respawn_emits_entity_moved_then_state_packets_in_load_bearing_order` (C3) | A2 fix: `EntityMoved` precedes all state packets in the burst. |
| `respawn_closes_loot_ui_for_still_looting_players` (C3-sibling) | A3 fix with full wire-byte payload assertion (entity_id + count=0 + initial=0). |
| `respawn_does_not_close_loot_ui_for_unrelated_players` (C3-sibling) | A3 sweep is correctly scoped to only the looting player. |
| `respawn_with_no_witnesses_is_silent_and_state_correct` (C4) | B3 fix: zero wire packets, correct server state. |

### Docs (2/2 fixed)

- `AiState` mapping table now has a **"Driven?"** column flagging Patrol/Wander/Follow as not-yet-wired (Phase 2/3/6).
- `npc-ai.md` NPC respawn row: full broadcast sequence, direction restore, loot UI close, and the `MeleeDamage` script bypass caveat from B2.

### Final stats

- **Test count:** 1117 → 1133 across `cimmeria-services` (+16 net since PR 1 first push; 13 dedicated respawn tests now)
- **Clippy / fmt / build:** clean against `--workspace -- -D warnings` (excluding GUI per CLAUDE.md)
- **Schema migration impact:** CHECK floor raised to `>= 3`. Existing seed rows already use `NULL`, so no data migration needed. Operators adding rows with `respawn_secs < 3` will get a constraint violation at insert time — intentional fail-fast.

The two I called blocking at review time (A1 direction, A4 clear_all_state_flags) are both pinned with tests now. The rest land as the design improvements the review asked for.

PR #424, opened 2026-05-31
feat(trainer): consolidate routing + delete dead stub + resend on prereq unlock (#55)

### LINKED ISSUE #55: Ability trainer NPC interaction
## Summary

Ability trainers let players learn new abilities for their archetype. The data model exists in Rust but all methods are `todo!()`.

## Current State in Rust

- `crates/game/src/interactions/trainer.rs` (42 lines): `TrainerSession`, `TrainableAbility` structs, `TrainerResult` enum. `open()` and `learn_ability()` are `todo!()`.
- Issue #33 covers the interaction check for already-known abilities.

## Reference Implementation

- `python/cell/interactions/AbilityTrainer.py` (141 lines):
  - `sendAbilityList()`: Queries trainable abilities for player's archetype and level, marks already-known.
  - `canTrainAbility()`: Checks level requirement and prerequisite abilities.
  - `trainAbility()`: Deducts currency, adds ability to player's known abilities, sends client update.
- 1,887 abilities in DB with archetype, level, and prerequisite data.

## Acceptance Criteria

- [ ] `open()`: Load trainer's ability list from DB, filter by player archetype, mark already-known
- [ ] `learn_ability()`: Validate level requirement, prerequisites, currency; grant ability
- [ ] Wire: send trainable ability list to client, send ability learned confirmation
- [ ] Tests cover: learn success, level too low, already known, missing prereq
- [ ] DB query: join abilities with archetype requirements

## Related

- Issue #33: Check which abilities player already knows

Cadacious:
## Summary

Resolves #55. The trainer interaction was substantially implemented but split across two competing code paths; per the deep-dive comment on #55, the work was a consolidation + a Python-parity resend, not a greenfield implementation.

- **Consolidated routing** to a single canonical handler `cell::interactions::trainer::try_open_trainer`. The duplicate logic that lived at `cell_methods/player/trainer_interaction.rs` is removed; the callsite in `cell_methods/player/interaction.rs::dispatch` now routes through the public path. The dead `NpcInteractionType::Trainer { archetype_id }` arm in `cell::interactions::dispatch::handle_interact` no longer calls the old stub — it logs a `reason=deprecated_routing_arm` WARN instead, in case a future content edit ever assigns the variant.
- **Deleted the 42-line dead stub** at `crates/game/src/interactions/trainer.rs` (`TrainerSession`, `TrainableAbility`, `TrainerResult` with `todo!()` at lines 25 and 30). These types had no callers outside the module itself.
- **Added Python-parity resend-on-prereq-unlock** in `base_messages::AbilityGranted`. After `send_known_abilities_update`, if the player has a trainer pinned via `last_interaction_target` we re-send `onTrainerOpen`; the recomputed `trainable` flags reflect the newly-known ability so dependent abilities become trainable once their prereq is learned. Mirrors `AbilityTrainer.onTrainAbility:128`.

The Ghidra deep-dive in #55 verified the wire formats byte-exactly against `entities/defs/SGWPlayer.def:1194-1198`, `alias.xml:417-422`, and `docs/protocol/client-method-dispatch-table.md:248` — no new wire messages were invented.

## Implementation note

The resend behavior in `AbilityGranted` reads `ConnectedClientState.last_interaction_target` (set by `handle_interact` when the player opens a trainer). If `last_interaction_target` is None, no resend fires. The negative-resend test pins this guard. Future trainer-related features must remember to set `last_interaction_target` for the resend to chain through.

The resend fires on **ANY** `AbilityGranted` while a trainer is pinned, not just abilities from that trainer or that are prereqs for offered abilities. This is intentional and matches Python's `AbilityTrainer.onTrainAbility` unconditional re-fire after a successful train RPC; `try_open_trainer` itself short-circuits to `false` when the pinned target isn't a trainer template, so non-trainer pinned targets (vendors, lootables, dialog NPCs) never trigger a spurious frame.

## Tests added

- `cell::cell_methods::player::interaction::tests::trainer_interact_emits_exactly_one_on_trainer_open` — routing-consolidation regression. Sets up an NPC with BOTH `template_trainer_lists` registration AND the legacy `NpcInteractionType::Trainer` tag (the pre-fix double-emit bug shape), asserts exactly one `onTrainerOpen` (method 113) is sent.
- `cell::service::base_messages::tests::ability_granted_trainer_resend::ability_granted_resends_trainer_open_with_unlocked_prereq` — positive resend regression. Trainer offers ability A (no prereqs) and B (prereq=[A]); `AbilityGranted` for A triggers a fresh `onTrainerOpen` showing B with `trainable=1`. Byte-exact wire assertions on the resent frame.
- `cell::service::base_messages::tests::ability_granted_trainer_resend::ability_granted_with_no_trainer_pinned_does_not_resend` — negative resend guard. When `last_interaction_target` is unset (chain-granted ability, GM grant, etc.) `AbilityGranted` MUST NOT emit `onTrainerOpen`.
- `cell::service::base_messages::tests::ability_granted_trainer_resend::ability_granted_resends_even_when_not_a_prereq` — pins the "resend on ANY grant while pinned" contract. Grants an ability that is NOT in the trainer's offered list and NOT a prereq for anything offered; the resend must still fire (Python parity).
- `cell::interactions::dispatch::tests::deprecated_trainer_arm_warns_without_emitting_wire_frame` — exercises the rewritten `NpcInteractionType::Trainer` warning arm. Asserts no wire frame is queued AND the WARN fires with `reason=deprecated_routing_arm`.

Also preserved from the consolidated module: `cell::interactions::trainer::tests::already_known_ability_shows_untrainable` — already-known ability filter regression.

## Files touched

Modified:
- `crates/services/src/cell/interactions/trainer.rs` — replaced the dead stub with the real implementation (moved from `cell_methods/player/trainer_interaction.rs`); now `pub(crate)`.
- `crates/services/src/cell/interactions/mod.rs` — re-export `try_open_trainer`.
- `crates/services/src/cell/interactions/dispatch.rs` — `NpcInteractionType::Trainer` arm now logs a `reason=deprecated_routing_arm` WARN instead of calling the deleted stub. Added warning-arm regression test.
- `crates/services/src/cell/cell_methods/player/interaction.rs` — callsite routes through `crate::cell::interactions::try_open_trainer`. Added routing-consolidation test.
- `crates/services/src/cell/cell_methods/player/mod.rs` — drop the `mod trainer_interaction;` declaration.
- `crates/services/src/cell/service/base_messages/mod.rs` — `AbilityGranted` handler re-sends `onTrainerOpen` when the player has a trainer pinned. Contract is documented inline.
- `crates/services/src/cell/service/base_messages/tests/mod.rs` — register new tests module.
- `crates/game/src/interactions/mod.rs` — drop `pub mod trainer` + tombstone comment pointing at the new canonical home.

Created:
- `crates/services/src/cell/service/base_messages/tests/ability_granted_trainer_resend.rs` — new resend tests.

Deleted:
- `crates/services/src/cell/cell_methods/player/trainer_interaction.rs` (consolidated into `cell/interactions/trainer.rs`).
- `crates/game/src/interactions/trainer.rs` — the 42-line dead stub with `todo!()`s.

## Test plan

- [x] `cargo check -p cimmeria-services --tests` — clean.
- [x] `cargo clippy -p cimmeria-services -p cimmeria-game --all-targets -- -D warnings` — clean.
- [x] `cargo fmt -p cimmeria-services -p cimmeria-game -- --check` — clean.
- [x] `cargo test -p cimmeria-services --lib -- trainer` — passes including the 5 new tests.
- [x] Module-level doc comment on the resend logic covers the "resend on ANY grant while pinned" contract (Python parity ref + non-trainer pin short-circuit).
- [ ] Smoke (manual): right-click the Commando trainer in Castle Cellblock, train ability A that has B as a dependent prereq, confirm B's greyed-out state clears in the trainer UI without close+reopen.

Closes #55

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed trainer UI from opening multiple times simultaneously.
  * Trainer interface now automatically refreshes and displays updated ability options when abilities are learned.

* **Refactor**
  * Consolidated trainer interaction handling through a single, unified system for improved reliability and consistency.


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `base_messages/mod.rs` resend trigger | **The resend fires on ANY `AbilityGranted` while a trainer is pinned, not just abilities from that trainer.** If a player has a trainer open and a quest chain grants an unrelated ability, the trainer UI refreshes anyway. This is probably fine (matches Python parity), but document that the resend is "any grant while pinned" not "grant from this trainer." |
| **G2** | `base_messages/mod.rs` trainer template check | **The resend checks `template_trainer_lists.contains_key(template_id)` but doesn't verify the pinned entity still exists.** If the trainer NPC despawned between the original open and the `AbilityGranted`, the resend would fail silently (no entity → no template_id → no resend). This is probably fine (graceful degradation), but the test doesn't cover this edge case. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `interactions/dispatch.rs` warning arm | **The `NpcInteractionType::Trainer` arm now logs a warning and returns `None`.** But the warning says "no NPC has it set today" — if a future content edit sets it, the warning fires but the trainer still opens via the template path. The warning could be confusing — it implies the trainer didn't open, but it did (via the other path). |
| **G4** | `interaction.rs` routing comment | **The comment says "the prior `cell_methods/player/trainer_interaction.rs` was the same logic copied under a different module."** But looking at the deleted file, it was the REAL implementation — the stub in `crates/game/src/interactions/trainer.rs` was the dead code. The comment is slightly misleading about which was the "copy." |
| **G5** | `ability_granted_trainer_resend.rs` test setup | **The test manually sets `last_interaction_target = Some(200)` but doesn't verify that `handle_interact` actually sets this field.** If a future change breaks the pinning, the resend test would still pass (because it manually sets the pin) but the feature would be broken in production. |

MyPalClara:

| **G6** | No test for trainer despawn between open and grant | The edge case where the trainer NPC despawns while the player has it pinned isn't tested. |
| **G7** | No test for `NpcInteractionType::Trainer` warning path | The warning arm in `handle_interact` isn't tested — only the template-driven path is. |
| **G8** | No test for resend when granted ability is NOT a prereq | The test covers the positive case (granted ability unlocks a dependent), but not the case where the granted ability doesn't unlock anything. The resend still fires (correct), but the test doesn't pin this. |

MyPalClara:

| **G9** | PR body doesn't mention the `last_interaction_target` dependency | The resend relies on `last_interaction_target` being set by `handle_interact`, but the PR body doesn't explain this linkage. |
| **G10** | No doc comment on the resend logic in `AbilityGranted` handler | The handler has inline comments but no module-level doc explaining the Python parity reference (`AbilityTrainer.onTrainAbility:128`). |

---

## Summary

| Category | Count |
|----------|-------|
| Correctness issues | 2 |
| Potential issues | 3 |
| Missing test coverage | 3 |
| Documentation | 2 |

**Total: 10 gaps**

Cadacious:
Thanks for the review. Pushed three commits addressing the feedback:

**`11679e19` — `fix(trainer): reword deprecated routing warning + document resend-any-grant contract`**

- **G3** (warning arm): the `NpcInteractionType::Trainer` arm in `handle_interact` now emits a `reason=deprecated_routing_arm` WARN whose wording matches the actual semantics — "this arm only runs as a fall-through after the canonical `template_trainer_lists` path already short-circuited; if you see this fire, your content edit set the legacy tag but forgot the template's `trainer_ability_list_id`." The prior wording implied no NPC ever sets the tag, which was misleading.
- **G1** (document the resend-on-any-grant contract): added an explicit **Contract — "resend on ANY grant while pinned"** block to the inline comment in `base_messages::AbilityGranted`. The handler intentionally fires the resend for every grant while `last_interaction_target` is set, matching Python `AbilityTrainer.onTrainAbility`'s unconditional re-fire. Non-trainer pinned targets are filtered by `try_open_trainer`'s template short-circuit, so vendors/lootables/dialog NPCs never trigger a spurious frame.

**`923fc89e` — `test(trainer): pin warning-arm path + resend-on-non-prereq-grant behavior`**

- **G8** (`ability_granted_resends_even_when_not_a_prereq`): grants an ability that's NOT in the trainer's offered list and NOT a prereq for anything offered; asserts the resend still fires. Pins the documented contract so a future "optimisation" filtering out non-prereq grants would trip it.
- **G7** (`deprecated_trainer_arm_warns_without_emitting_wire_frame`): exercises the warning arm directly via `handle_interact` with an NPC carrying the legacy tag but no `template_trainer_lists` entry. Uses `LogCapture` from `test_support.rs` to assert both `Level::WARN` and `reason=deprecated_routing_arm`. The reason-field match is exact-string, so reverting either the log or the field-key trips the test.

**`52f5f8d5` — `style(trainer): strip issue-number provenance from source comments per repo convention`**

Addresses all 12 Copilot comments — same pattern at every site. Stripped `#55`, `Item A`, `Item B`, `deep dive`, and `issue #55` references from module docstrings, inline comments, test docstrings, and assertion messages across the 5 remaining files (the other two had their refs removed during the rewrites in commit 1). Behavioral context (Python parity refs, `.def` line numbers, dispatch-table refs, bug-shape descriptions) is preserved.

**PR body updated** with the `last_interaction_target` linkage callout (Clara G9 / G10) and the new test names. Module-level doc on the resend logic (G10) is the inline comment in `base_messages::AbilityGranted` — covered the contract there rather than adding a separate doc block, since the handler is a `match` arm rather than a standalone function.

**Skipped (defensible to leave):**

- **G6** (no test for trainer-despawn-mid-open): uncommon edge case, the existing tests pin the happy path. `try_open_trainer` already handles "target entity doesn't exist" by returning `false`, so the silent-fail behavior is already correct — the missing piece is just the test that asserts it. Will follow up if it matters in practice.
- **G2** (resend doesn't verify pinned entity still exists): same shape as G6 — the code path handles it gracefully (no entity → no template_id → no resend) but the test doesn't pin it. Could be folded into a G6 follow-up.
- **G4** (routing comment slightly misleading about which path was the "copy"): both modules had real-ish implementations; the comment was reworded as part of commit 1 / 3 already.
- **G5** (test manually sets `last_interaction_target` rather than verifying `handle_interact` sets it): there are pre-existing tests in `cell::interactions::dispatch::tests` that pin `handle_interact`'s `last_interaction_target` write (`interact_pins_last_interaction_target_on_player`, `out_of_range_interact_does_not_pin_target`); the resend tests are isolated regression guards for the resend itself, not end-to-end pinning tests.
- **Codecov 9 missing lines**: most are the warning arm (now covered by `deprecated_trainer_arm_warns_without_emitting_wire_frame`) and the negative-resend (already tested). Coverage should improve naturally on the next CI run.

CI checks (`cargo check`, `cargo clippy -- -D warnings`, `cargo fmt --check`, `cargo test -p cimmeria-services --lib`) all green locally on the touched crates.

Cadacious:
@coderabbitai review

PR #425, opened 2026-05-31
feat(chat): wire speaker_flags from access_level + DND state (#65)

### LINKED ISSUE #65: Chat speaker flags: AFK, DND, GM status indicators
## Summary

Chat infrastructure works but speaker flags (AFK, DND, GM) are not computed from player state.

## Current State in Rust

- `crates/services/src/base/dispatch.rs:94`: TODO — compute speaker_flags from AFK/DND/GM status.
- Chat system (296 lines in cell service): handles global/faction/local channels.

## Reference Implementation

- `python/base/Chat.py` (365 lines): 10+ built-in channels, custom channel creation/deletion, password-protected channels.
- `python/base/SGWPlayer.py`: `chatSetAFKMessage()`, `chatSetDNDMessage()` RPCs.
- `python/common/Constants.py`: CHANNEL_FLAG definitions.

## Acceptance Criteria

- [ ] AFK flag: set/clear via player RPC, reflected in speaker_flags
- [ ] DND flag: set/clear via player RPC, reflected in speaker_flags
- [ ] GM flag: derived from access_level, shown in chat messages
- [ ] speaker_flags sent with every chat message for client-side display
- [ ] Custom channel creation and password protection (if not already working)

Cadacious:
## Summary

- Computes `speaker_flags` on every chat dispatch from per-connection state instead of the hardcoded `0` TODO at `crates/services/src/base/dispatch.rs:105`. Matches `python/base/Chat.py::getSpeakerFlags`: `SPEAKER_GM` (0x01) when `access_level >= 1`, `SPEAKER_DND` (0x04) when `dnd_message.is_some()`.
- Adds `ConnectedClientState::dnd_message` and wires the previously-stub `CHAT_SET_DND` (0xC4) handler to set/clear it (Python parity: a ≤1-char payload clears, anything longer sets).
- Removes the `#[allow(dead_code)]` on `ConnectedClientState::access_level` — now read by chat dispatch.
- Drive-by fix in the same dispatch arm: `read_wstring` returns BYTES CONSUMED, not a new absolute offset, but the existing code did `offset = new_offset`, dropping the +1 from the channel byte. Switched to `offset += target_bytes` (with a comment citing the discovery context). Caught by the new tests because the default-target spatial-channel case never reached the cell-forward branch otherwise.

Per the [issue #65 deep dive](https://github.com/SandboxServers/Cimmeria/issues/65):
- `SPEAKER_Petition` (0x02) is declared in `entities/defs/enumerations.xml` but never set by the Python reference — intentionally omitted.
- AFK is **NOT** a speaker flag (no `SPEAKER_AFK` token); the issue title was wrong. `CHAT_SET_AFK` (0xC3) remains log-only with a clarifying comment.
- `docs/gameplay/chat-system.md` already shows speaker flags as DONE per the deep-dive — no doc update needed.

## Tests added

All five are in `crates/services/src/base/dispatch.rs` test module. Tests 1–4 fail if the hardcoded `speaker_flags: 0` is re-introduced; test 5 fails if the DND handler reverts to the stub. The wire serializer in `cell/chat.rs` is already byte-exact (pinned by `serialize_on_player_communication_basic`); these guard the bit-assembly done before the cell forward.

1. `send_player_communication_default_state_has_zero_speaker_flags` — access_level=0 + no DND ⇒ `speaker_flags == 0`.
2. `send_player_communication_access_level_one_sets_gm_flag` — access_level=1 sets `SPEAKER_GM`, leaves `SPEAKER_DND` clear.
3. `send_player_communication_dnd_message_sets_dnd_flag` — `dnd_message=Some` sets `SPEAKER_DND`, leaves `SPEAKER_GM` clear when access_level=0.
4. `send_player_communication_gm_with_dnd_sets_both_flags` — access_level=2 + `dnd_message=Some` produces `0x05` (both bits).
5. `chat_set_dnd_handler_sets_then_clears_dnd_message` — round-trips the handler: a >1-char payload sets `dnd_message`, an empty payload clears it.

## Test plan

- [x] `cargo test -p cimmeria-services --lib base::dispatch::tests` — all 7 pass (5 new + 2 existing)
- [x] `cargo fmt` no-op on touched files
- [ ] Smoke: GM and a non-GM player chat in the same world; client UI shows the GM tag only on the GM's messages
- [ ] Smoke: `/dnd <message>` via the client (or direct `CHAT_SET_DND` send) flips the bit on the next chat message; sending an empty DND clears it

Closes #65

## Release Notes

* **New Features**
  * Enhanced spatial chat to compute speaker flags dynamically based on player status (GM, DND)
  * Added Do Not Disturb (DND) message storage with auto-reply functionality
  * Explicit Away From Keyboard (AFK) chat status handling

* **Bug Fixes**
  * Corrected chat message payload parsing

* **Tests**
  * Expanded test coverage with regression guards and speaker flags validation across multiple scenarios

### DIFF
--- a/crates/services/src/base/dispatch.rs
+++ b/crates/services/src/base/dispatch.rs
@@ -12,6 +12,22 @@ use crate::mercury::read_wstring;
 
 use super::ConnectedClientState;
 
+/// `ESpeakerFlags` bitfield constants from `entities/defs/enumerations.xml`.
+///
+/// The wire field is a UINT8 sent in every `onPlayerCommunication` message.
+/// Only `SPEAKER_GM` and `SPEAKER_DND` are computed today — matches
+/// `python/base/Chat.py::getSpeakerFlags`. `SPEAKER_Petition` (0x02) is
+/// declared in the enum but never set by the Python reference, so it is
+/// intentionally omitted here (see issue #65 deep dive).
+pub(crate) mod speaker_flags {
+    /// Set when the speaker's `access_level >= 1` (Moderator or GameMaster).
+    /// Python parity: `if player.accessLevel > 0`.
+    pub const GM: u8 = 0x01;
+    /// Set when the speaker has a non-empty DND auto-reply message.
+    /// Python parity: `if player.dndMessage is not None`.
+    pub const DND: u8 = 0x04;
+}
+
 /// SGWPlayer base-method message IDs we currently handle explicitly.
 ///
 /// The client also sends protocol-level messages such as `versionInfoRequest`
@@ -66,12 +82,19 @@ pub(crate) async fn dispatch_sgw_player_base_method(
             let channel = payload[0];
             let mut offset = 1;
 
-            // Parse target (WSTRING)
-            let (target, new_offset) = match read_wstring(payload, offset) {
+            // Parse target (WSTRING). `read_wstring` returns the number of
+            // BYTES CONSUMED (not the new absolute offset), so accumulate
+            // with `+=`. The earlier code did `offset = new_offset` which
+            // dropped the +1 for the channel byte and mis-aligned the text
+            // WSTRING read — discovered while wiring #65 speaker_flags
+            // (the dispatch never reached the cell-forward branch on
+            // empty-target spatial channels because `read_wstring(buf, 4)`
+            // saw garbage where the text length should have been).
+            let (target, target_bytes) = match read_wstring(payload, offset) {
                 Ok(v) => v,
                 Err(_) => return Ok(()),
             };
-            offset = new_offset;
+            offset += target_bytes;
 
             // Parse text (WSTRING)
             let (text, _) = match read_wstring(payload, offset) {
@@ -90,10 +113,30 @@ pub(crate) async fn dispatch_sgw_player_base_method(
                 "sendPlayerCommunication"
             );
 
-            // Route to CellService for spatial channels (say/emote/yell)
-            let player_eid = {
+            // Route to CellService for spatial channels (say/emote/yell).
+            //
+            // Read player_eid + access_level + dnd_message under a single
+            // lock acquisition. Computing `speaker_flags` matches
+            // `python/base/Chat.py::getSpeakerFlags`:
+            //   - SPEAKER_GM  if accessLevel > 0  (Moderator or higher)
+            //   - SPEAKER_DND if dndMessage is not None
+            // SPEAKER_Petition (0x02) is in the enum but never set by the
+            // Python reference, so it is intentionally not computed.
+            let (player_eid, speaker_flags_value) = {
                 let clients = connected.lock().unwrap();
-                clients.get(&addr).and_then(|c| c.player_entity_id)
+                match clients.get(&addr) {
+                    Some(c) => {
+                        let mut flags: u8 = 0;
+                        if c.access_level >= 1 {
+                            flags |= speaker_flags::GM;
+                        }
+                        if c.dnd_message.is_some() {
+                            flags |= speaker_flags::DND;
+                        }
+                        (c.player_entity_id, flags)
+                    }
+                    None => (None, 0),
+                }
             };
 
             if let Some(player_eid) = player_eid {
@@ -102,7 +145,7 @@ pub(crate) async fn dispatch_sgw_player_base_method(
                         .send(BaseToCellMsg::ChatMessage {
                             entity_id: player_eid,
                             speaker_name: speaker.to_string(),
-                            speaker_flags: 0, // TODO: compute from AFK/DND/GM status
+                            speaker_flags: speaker_flags_value,
                             channel,
                             text,
                         })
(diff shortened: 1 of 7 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `dispatch.rs` `access_level >= 1` vs Python `> 0` | **The code uses `access_level >= 1` but the comment says "Python parity: `accessLevel > 0`."** These are equivalent for integers, but the inconsistency is confusing. Pick one and use it consistently in both code and comments. |
| **G2** | `dispatch.rs` DND clear threshold | **The code clears DND when `message.chars().count() > 1` is false (i.e., ≤1 char).** But the comment says "empty or 1-char message clears." A 1-char message like "x" would clear DND — is this intentional? The Python reference uses `len(message) <= 1` which is the same, but it's worth verifying this is the intended behavior. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `dispatch.rs` `SPEAKER_Petition` omission | **The code intentionally omits `SPEAKER_Petition` (0x02) because "never set by the Python reference."** But what if a future feature needs it? Consider defining the constant even if unused, with a comment explaining it's reserved. |
| **G4** | `mod.rs` `dnd_message` doc comment | **The doc comment says "AFK does NOT contribute to `ESpeakerFlags`" but doesn't explain what AFK DOES do.** The `CHAT_SET_AFK` handler is log-only — consider adding a brief note about the auto-reply-tells feature being future work. |
| **G5** | `dispatch.rs` drive-by fix scope | **The offset fix is described as a "drive-by" but it's actually a real bug fix.** The PR body mentions it was "discovered while wiring #65" — consider whether this deserves its own issue/PR for traceability, or at least a more prominent mention in the PR body. |

MyPalClara:

| **G6** | No test for the offset fix | The drive-by offset fix isn't directly tested. The speaker_flags tests exercise the cell-forward branch (which was broken before the fix), but there's no test that explicitly verifies the offset accumulation is correct. |
| **G7** | No test for `CHAT_SET_AFK` handler | The AFK handler is log-only, but there's no test verifying it doesn't crash or misbehave. |
| **G8** | No test for 1-char DND message | The test covers >1-char (sets) and 0-char (clears), but not exactly 1-char. The code treats 1-char as "clear" — this edge case should be pinned. |

MyPalClara:

| **G9** | `speaker_flags` module doc doesn't mention the wire field size | The doc says "wire field is a UINT8" but doesn't mention where in the `onPlayerCommunication` frame it appears. |
| **G10** | No mention of the offset fix in the PR title | The PR title is "wire speaker_flags from access_level + DND state" but the offset fix is arguably equally important. |

---

## Summary

| Category | Count |
|----------|-------|
| Correctness issues | 2 |
| Potential issues | 3 |
| Missing test coverage | 3 |
| Documentation | 2 |

**Total: 10 gaps**

Cadacious:
@coderabbitai review

Cadacious:
Pushed three commits addressing the review feedback (`74c0820c`, `3a4bac7f`, `a635b111` — fast-forward, no force).

## Addressed

**Correctness bugs (CodeRabbit)**

- **DND leaks across characters** — `dnd_message` is now cleared in the `LOG_OFF` `disconnect == 0` reset block alongside the other per-character fields. Without this, char A's `/dnd` would carry into char B on the same connection and every subsequent `sendPlayerCommunication` would incorrectly carry `SPEAKER_DND`. New regression guard: `logoff_disconnect_zero_clears_dnd_message_for_next_character`.
- **CHAT_SET_DND silently clears on malformed payload** — replaced `unwrap_or_default()` with an explicit match that warn-logs and returns on decode failure, so a truncated/garbage packet no longer destroys existing DND state. Followed `docs/architecture/negative-logging-convention.md` for the WARN shape (`reason = "read_wstring_failed"` + `payload_len` + `addr`). New regression guard: `chat_set_dnd_handler_malformed_payload_preserves_dnd_message` (pins the level AND the structured `reason` field).

**Convention violations (Copilot)**

- Stripped all five `#65` references from source comments: `mod.rs:125`, `dispatch.rs:21,89,178,505`. Behavioral rationale preserved inline (no `SPEAKER_AFK` token in `enumerations.xml`, `SPEAKER_Petition` never set by the Python reference, etc.). Repo convention per `.github/copilot-instructions.md`: provenance lives in the PR body.

**Test coverage gaps (CodeRabbit CR-3, Clara G7)**

- `chat_set_dnd_handler_one_char_payload_clears_dnd_message` — pins the `> 1` threshold (matches Python's `len(message) <= 1` clear rule). A regression from `> 1` to `> 0` would treat `"x"` as setting DND to `"x"` and trip this test.
- `chat_set_afk_handler_is_log_only_and_preserves_state` — drives the AFK handler with empty, valid, and malformed payloads; asserts (a) Ok return and (b) no mutation of `player_entity_id`, `dnd_message`, or `access_level`. Pins AFK's intentional omission from speaker_flags so a future "AFK should set a flag" change forces a deliberate review of `enumerations.xml`.

**Style consistency (Clara G1)**

- Changed `access_level >= 1` to `access_level > 0` (both code site and doc comment in `ConnectedClientState::access_level`) so the code matches the Python-parity comment exactly. Python uses `> 0`.

## Intentionally skipped (with rationale)

- **Clara G3 (SPEAKER_Petition omission)** — defensible. `SPEAKER_Petition` (0x02) is declared in `enumerations.xml` but never set by the Python reference (`python/base/Chat.py::getSpeakerFlags` only checks `accessLevel > 0` and `dndMessage is not None`). No current Cimmeria feature drives a petition state, so wiring this bit would be speculative. Rationale documented inline in the `speaker_flags` module doc.
- **Clara G10 (offset fix prominence in PR title)** — the PR title is already merged into the branch history (it's the existing `48b4b601` commit). A force-push to amend the squash-title cosmetics isn't worth the round-trip given the offset fix is documented in the per-commit message and the inline comment. Happy to capture this in the PR body if it would be useful for the eventual squash-merge changelog.
- **CodeRabbit's suggestion to silently `return Ok(())` on malformed CHAT_SET_DND** — followed your instruction to warn-log instead so the failure mode is visible in SigNoz. The negative-logging-convention's level table puts decode-from-network failures at `debug!` for "expectation unmet, normal/transient," but a malformed payload that would have wiped DND state is closer to `warn!` ("expectation unmet, player-visible, recoverable") — the user would notice their DND silently dropping. Open to downgrading to `debug!` if reviewers think the warn-rate from truncated retransmits would be noisy.

## Local validation

```
cargo fmt -p cimmeria-services -- --check       # clean
cargo clippy -p cimmeria-services --all-targets -- -D warnings   # clean
cargo test -p cimmeria-services --lib base::dispatch::tests      # 11 passed (5 pre-existing + 4 new + 2 unrelated)
```

Issue #46, opened 2026-03-28
Extract collision geometry from UE3 .umap chunks and regenerate navmeshes


Cadacious:
## Overview

The existing `.nav` files in `data/spaces/` were generated in 2012-2014 and have quality issues — NPCs float above ramps, rubber-band at navmesh boundaries, and some areas are not walkable. We need to regenerate them from the game client cooked map data with better Recast configuration.

Pipeline: **UE3 .umap chunks → extract collision geometry → .obj files → NavBuilder → .nav files → server loads via Detour FFI**

### What we have
- `tools/upk_parser.py` — Custom UPK/UMAP parser that handles the modified UE3 format (Epic v486, Licensee v8, Engine 3004, LZO compressed). Parses headers, name tables, import/export tables. **Confirmed working** on all Castle_CellBlock chunks.
- `src/nav_builder/builder.cpp` — C++ NavBuilder that reads `.obj` geometry files, runs the full Recast pipeline, and outputs `.nav` files (XRC format).
- `data/spaces/*.nav` — 5 existing navmeshes (generated 2012-2014).
- `external/recast/` — Recast Navigation 1.6.0 (vendored, DT_NAVMESH_VERSION=7).

### What we need
Extend `upk_parser.py` to extract vertex/face data from `Terrain` and `Brush` exports in chunk `.umap` files, write combined geometry as `.obj`, then run NavBuilder to produce new `.nav` files.

## Map Data

Client cooked maps: `../SGW/Stargate Worlds-QA/Working/SGWGame/CookedPC/Maps/`

### Maps and chunk counts

| Map | Chunks | Has Navmesh | Instanced |
|-----|--------|-------------|-----------|
| Agnos | 692 | agnos.nav (411KB) | No |
| Agnos_Library | 37 | - | No |
| Beta_Site_Evo_1 | 631 | - | No |
| Castle | 145 | - | No |
| Castle_CellBlock | 65 | castle_cellblock.nav (168KB) | Yes |
| Dakara_E1 | 401 | - | No |
| Dakara_E1_StoryRm | 2 | - | Yes |
| Harset | 73 | harset.nav (2.5MB) | No |
| Harset_CmdCenter | 17 | - | No |
| Harset_Market | 10 | - | Yes |
| Harset_StorageRm | 10 | harset_storagerm.nav (108KB) | Yes |
| Ihpet_Crater_Dark | 155 | - | No |
| Ihpet_Crater_Light | 155 | - | No |
[...]

PR #426, opened 2026-05-31
feat(navmesh-extractor): Phase 0 — crate scaffold + nav round-trip smoke (#46)

Cadacious:
## Summary

First PR for the asset-pipeline epic in #46. Lands the new
`crates/navmesh-extractor` crate with Phase 0 + Phase 1.1
scaffolding only; geometry extraction itself (Phase 1.2/1.3) is
deferred to follow-up PRs.

The architectural decision motivating this PR is captured in the deep
dive on #46: **keep `deprecated/cpp/src/nav_builder/` as the build-time
Recast driver**, drive it from a Rust crate that emits `.obj` files.
80% of the work in #46 is extracting UE3 collision geometry from `.umap`
chunks — Recast itself is a 5-minute step at the end. Porting the
Recast wrapper is a future cleanup, not a blocker on getting Castle
Cellblock's ramps walkable.

### Phase 0 — `.nav` round-trip smoke

`nav_roundtrip::XrcNav` reads and re-emits the XRC format produced by
`xrcSavePolyMesh` (`deprecated/cpp/src/nav_builder/builder.cpp:10-38`).
The byte order matches `crates/entity/src/navigation.rs::NavMesh::load`
exactly. The integration test reads `data/spaces/castle_cellblock.nav`
(168,075 bytes = 0x2908B), re-emits it, and byte-compares — any
divergence yields a `RoundTripMismatch` pointing at the first
diverging byte. This is the canonical Rust-side ground truth for the
`.nav` wire format; a future Rust replacement for NavBuilder writes
the same fields in the same order.

### Phase 1.1 — Crate scaffolding

Module structure for the extractor:

| Module | Role |
|---|---|
| `chunk_id` | Decode `<MapName>-<HEX8>.umap` filenames into `(positionX, positionZ)` per `chunk.cpp:21-29`. Records the corrected axis labels from the deep dive (low → UE3 Y, high → UE3 X). |
| `umap` | Thin wrapper over `cimmeria_upk::Package` for chunk enumeration + per-export summaries. |
| `geometry` | `TriangleSoup` accumulator + `triangulate_terrain` stub. |
| `obj` | Wavefront writer with cross-group vertex re-indexing. |
| `nav_roundtrip` | XRC reader/writer pair. |
| `lib::extract_map` | Orchestrator with `// Phase 1.2` and `// Phase 1.3` markers at the per-chunk dispatch sites. |

### Tests

23 unit + 1 integration. All passing on Windows. The integration test
self-skips when `data/spaces/castle_cellblock.nav` is absent so CI
environments without the data dir stay green.

Of particular note: the six-chunk `chunk_id::six_chunks_decode_to_expected_positions`
test pins the (low → x, high → z) cast semantics — flipping the cast
silently misplaces every chunk relative to the C++ NavBuilder's
expectations and would surface only at integration time.

## Deferred

- **Phase 0.3 cube round-trip** — empirically pin the OBJ axis
  convention by emitting a known cube, running NavBuilder, and
  comparing `bmin`/`bmax`. Currently inferred from `mesh.cpp:106-108`
  but not exercised end-to-end. Flagged in the `obj` module doc and
  the README.
- **Phase 1.2** — StaticMesh instancing. The decoder already exists
  in `crates/upk-objects/src/static_mesh.rs`; just needs wiring into
  the orchestrator.
- **Phase 1.3** — Terrain decoder. The recipe is documented at 92%
  confidence in
  `.claude/agent-memory/game-archaeology-specialist/ue3-terrain-serialize.md`
  (binary trailer layout, `FUN_0075a5b0` / `FUN_0075a700` /
  `FUN_0075c9f0` helpers from SGW.exe).
- **Phase 1.4** — BSP `Model` / `Polys` decoder. Needs a Ghidra trace
  of `UModel::Serialize` (currently ~50% confidence per the deep
  dive's risk register).
- **Phase 2+** — NavBuilder rebuild, Castle_CellBlock acceptance
  smoke, Recast tuning (`cs=0.15`, `ch=0.1`, `agentClimb=0.5`),
  multi-map rollout (19 maps still lack navmeshes), validation
  harness.

## Test plan

- [x] `cargo fmt -p cimmeria-navmesh-extractor -- --check`
- [x] `cargo clippy -p cimmeria-navmesh-extractor --all-targets -- -D warnings`
- [x] `cargo test -p cimmeria-navmesh-extractor` — 24/24 pass
- [x] Round-trip integration test re-emits the real
      `data/spaces/castle_cellblock.nav` byte-for-byte
- [x] Verified `cimmeria-upk` and `cimmeria-upk-objects` still build
      against the workspace dependency entry

Refs #46


Cadacious:
@coderabbitai review

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `nav_roundtrip.rs` `XrcNav::read` overflow potential | **`nverts * 3`, `npolys * nvp * 2`, `detail_nmeshes * 4`, `detail_nverts * 3`, `detail_ntris * 4` are all computed as `u32 * usize`.** A malicious `.nav` file with `nverts = 0xFFFFFFFF` would overflow the multiplication and allocate a tiny Vec, then read past it. **Use `checked_mul` or validate bounds before allocation.** |
| **G2** | `nav_roundtrip.rs` memory exhaustion | **No upper bound on `nverts`, `npolys`, `detail_nverts`, `detail_ntris`.** A crafted `.nav` with `nverts = 0x10000000` would try to allocate ~3GB for `verts` alone. **Consider a sanity cap (e.g., 10M vertices) with an explicit error.** |
| **G3** | `chunk_id.rs` axis label confusion | **The doc says "low u16 → UE3 Y" and "high u16 → UE3 X" but the accessor is named `position_x()` for the low u16.** This is intentional (matching C++ NavBuilder names), but **the naming is actively misleading**. A future contributor will assume `position_x()` returns the X coordinate. **Consider renaming to `navbuilder_x()` / `navbuilder_z()` or adding a LOUD doc comment.** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `umap.rs` `tempdir()` test helper | **The test helper creates a temp dir at `std::env::temp_dir().join(format!("cimmeria-navmesh-extractor-test-{}", std::process::id()))`.** If two test threads run concurrently with the same PID (impossible, but if the test is run twice in quick succession), the `remove_dir_all` could race. **Use a UUID or timestamp suffix.** |
| **G5** | `nav_roundtrip.rs` `round_trip` returns `Ok(parsed)` on success | **The function name suggests it returns a bool or `()`, but it returns the parsed `XrcNav`.** This is fine, but **the doc comment does not mention the return value is the parsed struct**. |
| **G6** | `geometry.rs` `TriangleSoup` is a stub | **The struct exists but `triangulate_terrain` is a stub returning `Ok(())`.** The PR body says Phase 1.2/1.3 will fill this in, but **there is no `todo!()` or `unimplemented!()` marker** — a caller could invoke it and get silent success. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G7** | `obj.rs` axis convention not empirically verified | **The PR body says "Phase 0.3 cube round-trip" is deferred.** The OBJ axis convention is inferred from `mesh.cpp:106-108` but not exercised end-to-end. **If the inference is wrong, every navmesh will be rotated 90°.** |
| **G8** | `lib.rs` `extract_map` has `// Phase 1.2` and `// Phase 1.3` markers | **These are comments, not `todo!()` macros.** A future contributor might not notice them. **Consider `todo!("Phase 1.2: StaticMesh instancing")` so the compiler warns if the stub is reached.** |
| **G9** | `enumerate_chunks` does not recurse | **The doc says "Recursing into subdirectories is NOT done."** This is correct for SGW, but **if a future map has nested chunk dirs, the extractor would silently skip them.** |

MyPalClara:

| **G10** | No test for `XrcNav::read` with malformed header | The tests cover truncated input and trailing bytes, but not a header with `nverts = 0xFFFFFFFF`. |
| **G11** | No test for `ChunkId::from_obj_path` with invalid suffix | The test covers the happy path, but not `foo.obj` (missing `o` before `.obj`). |
| **G12** | No test for `summarize_exports` | The function exists but has no unit test. The integration test exercises it indirectly, but a unit test would pin the behavior. |

MyPalClara:

| **G13** | `chunk_id.rs` axis label table is in the module doc, not the struct doc | A reader looking at `ChunkId` will not see the axis mapping unless they scroll up. |
| **G14** | No README in `crates/navmesh-extractor/` | The crates README mentions the crate, but there is no crate-level README explaining how to run the extractor. |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 3 |
| Correctness issues | 3 |
| Potential issues | 3 |
| Missing test coverage | 3 |
| Documentation | 2 |

**Total: 14 gaps**

Cadacious:
## Review feedback addressed

Six commits pushed on top of the original Phase 0 work. All checks green locally on Windows: `cargo fmt -p cimmeria-navmesh-extractor -- --check`, `cargo clippy -p cimmeria-navmesh-extractor --all-targets -- -D warnings`, `cargo test -p cimmeria-navmesh-extractor` (27 unit + 1 integration, up from 23 + 1).

### Addressed

| Reviewer | ID | Resolution | Commit |
|---|---|---|---|
| Clara | **G1** / **G2** | `XrcNav::read` now caps every header count (`MAX_NVERTS=1M`, `MAX_NPOLYS=1M`, `MAX_NVP=64`, `MAX_DETAIL_*=10M`) and routes every `count*stride` through a `checked_alloc_size` helper. New `ExtractError::NavHeaderOutOfRange` variant carries the offending field name. | `21b2b681` |
| Clara | **G10** | Two new regression tests (`read_rejects_oversized_nverts`, `read_rejects_oversized_npolys`) — synthesise hostile headers and assert the typed `NavHeaderOutOfRange` error before any allocation. Revert the bounds check and both tests fail (the bug-shape contract). | `d9fa5631` |
| Copilot | inline #2 | `ChunkId::from_obj_path` now requires `stem.len() == 9` exactly and slices `&stem[..8]` — pre-fix, `prefix-12345678o.obj` would silently parse as `0x12345678`. | `be58912d` |
| Clara | **G11** | `obj_path_rejects_prefixed_form` + `obj_path_rejects_missing_marker` regression tests added in the same commit as the fix above. | `be58912d` |
| Copilot | inline #1 / Clara **G4** | `umap::tests::tempdir()` helper now combines pid + thread id + nanosecond timestamp; concurrent test threads in the same process no longer race on `remove_dir_all`/`create_dir_all`. | `d564c392` |
| Clara | **G8** | `// Phase 1.2` and `// Phase 1.3` comments in `extract_map` promoted to `todo!()` macros. `extract_map` has zero callers today (verified by grep), so this is risk-free; the runtime panic makes the missing-phase state unmissable for the next implementer. | `b95610c8` |
| Copilot | inline #3 / #4 / #5 | Stripped `#46` from `tests/nav_roundtrip_castle_cellblock.rs:46`, `src/chunk_id.rs:181`, and `README.md:33` per the repo convention (no PR/issue numbers in source — provenance lives in the PR body). Behavioral references ("asset-pipeline deep dive", design note pointers) preserved. | `772e08ad` |

### Skipped (with rationale)

- **Clara G3** — `position_x` / `position_z` naming. Deliberate: the C++ NavBuilder's `chunk.cpp` uses those exact field names, and the round-trip integration test relies on the labels matching. Renaming would invert the load-bearing convention. The module-level doc and struct doc both spell out that low-u16 → UE3 Y, high-u16 → UE3 X. Worth revisiting if/when NavBuilder gets rebuilt in Rust and we control both sides.
- **Clara G5** — `round_trip` doc not mentioning the return is the parsed struct. Already documented in the function's first paragraph ("Returns the parsed `XrcNav` on success; otherwise..."). No change needed.
- **Clara G6** — `triangulate_terrain` is a stub returning `TriangleSoup::new(None)`. Now that `extract_map`'s loop is `todo!()`-gated, no caller can reach `triangulate_terrain` without going through a panic point first. Adding a `todo!()` directly on `triangulate_terrain` would break its existing unit test. Acceptable as-is for the Phase 1.3 follow-up.
- **Clara G7** — Phase 0.3 cube round-trip not verified. Explicitly deferred in the PR body and in `README.md`'s "Known unknowns". Phase 1.2 is the right place to verify since it'll actually exercise the OBJ-emit path against a real mesh.
- **Clara G9** — `enumerate_chunks` doesn't recurse. Correct for SGW's flat-chunk-dir cooking convention; flagging this as a behavior change would only matter if a future map needed nested dirs, which is speculative. Doc already says "Recursing into subdirectories is NOT done."
- **Clara G12** — No unit test for `summarize_exports`. The integration test exercises it; adding a unit test would need a synthesised `.umap` fixture, which is more infrastructure than this phase warrants. Defer to Phase 1.2.
- **Clara G13** — Axis label table in module doc not struct doc. Cosmetic; the struct doc points to the module doc. No change.
- **Clara G14** — No crate README. There is one at `crates/navmesh-extractor/README.md`. Verified during this pass.

### Note on commit ordering

The user-suggested commit list had the obj-filename test (commit 2) landing before the obj-filename fix (commit 3), which would make `git bisect` fail at commit 2. I reordered to keep bisect green: fix + its tests are paired within `be58912d`. The transient fmt fix in `d9fa5631` (test commit) is corrected in `772e08ad` (style commit) — only the tip needs to be clean for CI, but a fully-bisect-clean variant of this branch would require a rebase.

Cadacious:
## Adversarial trust-boundary review (server-authority-enforcer)

**Verdict: SHIP — defenses verified against the threat model.**

Different threat model from a request handler — this is a build-time tool consuming `data/spaces/*.nav` from a filesystem path. The realistic attacker has filesystem-write access, which is a higher bar than network-attacker, but defense-in-depth still matters because (a) malformed input shouldn't crash CI, and (b) the helpers minted here are the canonical idiom for the wider codebase.

### `XrcNav::read` parser — verified hardened

The PR description claimed "already hardened in the review round"; verified at the origin tip (commit `21b2b681`):

- `check_count` validates every header count against a sanity cap (`MAX_NVERTS=1M`, `MAX_NPOLYS=1M`, `MAX_NVP=64`, `MAX_DETAIL_NMESHES=1M`, `MAX_DETAIL_NVERTS=10M`, `MAX_DETAIL_NTRIS=10M`) **before** any allocation.
- `checked_alloc_size` computes `count * stride` as `u64` with `checked_mul`, then `usize::try_from` — closes the path where `(nverts * 3) as usize` could wrap on 32-bit hosts or for adversarial input on 64-bit.
- Regression tests `read_rejects_oversized_nverts` + `read_rejects_oversized_npolys` synthesise hostile headers (`0xFFFFFFFF`), assert the typed `NavHeaderOutOfRange` error, and would fail if the caps were reverted. That's the right bug-shape contract for a regression guard.

A handcrafted `.nav` with `nverts = 0xFFFFFFFF` is now rejected before any Vec is sized, instead of wrapping the multiplication to a 3-element Vec and leaving every downstream offset wrong.

### Minor inconsistency (advisory — not blocking)

`regs`, `flags`, and `areas` allocations still use `vec![0u16; npolys as usize]` directly without going through `checked_alloc_size`. This is **safe in practice** because `npolys ≤ MAX_NPOLYS = 1M` is already capped by the time these lines execute, but it's stylistically inconsistent with the rest of the section. Not a fix-required item; flagging it so a future contributor copying this pattern doesn't forget the cap-precedence.

### `chunk_id.rs` — verified safe

All inputs come from `u32::from_str_radix` on bounded-length hex strings (8 chars) decoded from filenames. No overflow surface, no Vec sizing from attacker values. The `from_obj_path` tightening to require exactly `stem.len() == 9` (commit `be58912d`) is the right fix — pre-fix, `prefix-12345678o.obj` would silently parse as `0x12345678`.

### `obj.rs`, `umap.rs` — out of attacker reach

`obj.rs` is write-only. `umap.rs::summarize_exports` delegates to `cimmeria_upk::Package::open`, which is the existing UPK parser's surface — out of scope for this PR's review.

### Open follow-up (out of this PR's scope)

`crates/entity/src/navigation.rs::NavMesh::load` is the **runtime** navmesh loader and has the EXACT same unguarded patterns this PR fixed in the build-time tool:

```rust
let mut verts = vec![0u16; (nverts * 3) as usize];
let mut polys = vec![0u16; (npolys * nvp * 2) as usize];
```

That's the production read path at server startup. Threat model bar is higher (operator with file write to `data/spaces/`), but a malicious `.nav` reaching that loader is a DoS / OOM vector. Worth a follow-up issue to port `check_count` + `checked_alloc_size` into a shared location (probably `cimmeria-common`) and have both the build-time tool and the runtime loader consume them. The helper shape from this PR is the right canonical idiom — see `agent-memory/server-authority-enforcer/pattern-checked-alloc-size.md` for the template.

🤖 Generated by server-authority-enforcer adversarial-review pass.

Cadacious:

Adversarial verification of the regression guards added in `d9fa5631` and `be58912d`. For each guard I reverted the corresponding fix locally (in worktree, not pushed) and re-ran the test to confirm it actually fails when the bug shape returns.

- Test: `nav_roundtrip::tests::read_rejects_oversized_nverts` — **VERIFIED REPRODUCES BUG**
  - Bug shape pinned: yes — `vec![0u16; (nverts * 3) as usize]` overflow with `nverts = 0xFFFFFFFF`
  - Verification method: reverted `check_count` + `checked_alloc_size` from `nav_roundtrip.rs` for `nverts`/`npolys`/`nvp` paths; ran test
  - Result: FAILED on revert with `panicked at ...nav_roundtrip.rs:177:36: attempt to multiply with overflow` — exactly the bug shape. Passes again when fix restored.
  - Risk: in release builds (overflow checks off) the overflow would wrap to a 3-element Vec and the test would receive `Ok` instead of the expected `NavHeaderOutOfRange` variant, which would still fail the test's `match` arm (just via a different panic message). Contract holds either way.

- Test: `nav_roundtrip::tests::read_rejects_oversized_npolys` — **VERIFIED REPRODUCES BUG**
  - Bug shape pinned: yes — same overflow on `(npolys * nvp * 2) as usize`
  - Verification method: same reverted state
  - Result: FAILED on revert with overflow panic at line 181. Passes when fix restored.

- Test: `chunk_id::tests::obj_path_rejects_prefixed_form` — **VERIFIED REPRODUCES BUG**
  - Bug shape pinned: yes — `prefix-12345678o.obj` would silently parse as chunk `0x12345678` under the original `stem.len() < 9` + suffix-slice logic
  - Verification method: reverted `chunk_id.rs::from_obj_path` to the original `if !stem.ends_with('o') || stem.len() < 9` + `stem.len() - 9..stem.len() - 1` slice
  - Result: FAILED on revert (assertion `is_err()` failed because the function returned `Ok(...)` for the prefixed form). Passes when fix restored.

- Test: `chunk_id::tests::obj_path_rejects_missing_marker` — **SUSPECT (not theatre, but mis-described)**
  - Bug shape pinned: no — the pre-fix code's `if !stem.ends_with('o')` already rejected `12345678.obj`. The test PASSES on the reverted code.
  - The commit message describes this as a regression guard against the pre-fix code, but it isn't one — it's a *new* behavior pin, not a *regression* guard. The docstring reads "Missing the trailing 'o' marker — also a malformed input." which is fine as a documentation lock-in, but the commit message overclaims it.
  - Suggested fix: keep the test; soften the framing in the commit/docstring from "regression guard" to "behavior lock-in" — or strengthen the test (e.g. pass a stem of length 9 that doesn't end in `o`, like `12345678X.obj`, which the pre-fix code would reject by suffix and the post-fix code rejects by both checks). Not a blocker.

### Strong examples
- The `nverts`/`npolys` guards send a synthetic header with one specific field set to `0xFFFF_FFFF` and zero everywhere else — minimal repro, no fixture coupling, exact variant match (`Err(NavHeaderOutOfRange { field: "nverts", .. })`). A future refactor that changes the error type *deliberately* fails this test loudly; an accidental regression to wrap-then-truncate fails it via panic. Both are the right shape.

### Coverage gaps
- The integration test `nav_roundtrip_castle_cellblock` self-skips cleanly when the fixture is absent (just `eprintln!` + `return`) — passes silently in CI, which is fine. No false-pass risk.
- No checked-alloc test for the three `detail_*` count fields. The fix added the same `check_count` + `checked_alloc_size` discipline to `detail_nmeshes`/`detail_nverts`/`detail_ntris`, but no regression guard exercises any of them. Low priority — they share the helper functions with the `nverts`/`npolys` guards — but a CI-only smoke that runs each cap through `check_count` (e.g. a table-driven test that exercises all six caps with one hostile value each) would cost ~15 lines and tighten the contract.

### Verdict
- PR #426: **READY TO MERGE.** Three of four new regression guards verifiably reproduce the bug shape; the fourth (`obj_path_rejects_missing_marker`) is a clean behavior lock-in mis-labeled as a regression guard but still useful. No replacement test needed.

Issue #53, opened 2026-03-29
Crafting system: disciplines, research, reverse engineering, alloying


Cadacious:
## Summary

Crafting has cell method dispatch indices but zero implementation. The Python reference has a complete crafting system with 5 activities.

## Current State in Rust

- Cell method indices defined: CRAFT (96), RESEARCH (97), REVERSE_ENGINEER (98), ALLOYING (99), RESPEC_CRAFTING (100).
- No implementation files, no data structures, no DB queries.

## Reference Implementation

- `python/base/Crafter.py` (533 lines): Complete system:
  - `craft()`: Blueprint + components → crafted item. Checks prerequisites, racial paradigm requirements, expertise gains.
  - `research()`: Research items for expertise XP in a discipline.
  - `reverseEngineer()`: Disassemble items into components with skill-based yield.
  - `alloy()`: Combine components into higher-tier materials.
  - `learnDiscipline()` / `spendAppliedSciencePoints()`: Discipline unlock tree (tech tree).
  - `gainExpertise()`: Expertise capped at 100% per discipline.
  - `backup()` / `restore()`: Persistence for crafting state.
- DB data: 499 recipes, crafting disciplines, blueprints, paradigms in `db/resources/`.
- Wire messages: `onUpdateDiscipline`, `onUpdateCraftingOptions`, `onUpdateKnownCrafts`, `onCraftingRespecPrompt`.

## Acceptance Criteria

- [ ] CraftingState struct: known disciplines, expertise levels, known recipes
- [ ] `craft()`: Validate blueprint + materials, consume inputs, produce output, gain expertise
- [ ] `research()`: Consume item, gain discipline expertise
- [ ] `reverseEngineer()`: Consume item, produce components
- [ ] `alloy()`: Combine components into alloy
- [ ] Applied Science point spending for discipline unlocks
- [ ] Wire protocol: all `onUpdate*` messages to sync client UI
- [ ] DB persistence: save/load crafting state
- [ ] Tests cover: craft success/fail, expertise gain, prerequisite checks

Cadacious:
## Test Plan

### Unit Tests
- [ ] `craft`: Valid blueprint + all materials → output item created, materials consumed, expertise gained
- [ ] `craft`: Missing material → operation rejected, no materials consumed
- [ ] `research`: Valid item → item consumed, discipline expertise increased
- [ ] `reverseEngineer`: Valid item → item consumed, component items produced
[...]

Issue #311, opened 2026-05-17
Tier 4 follow-up: Mercury dispatch logs unhandled method indices


BangRocket:
Split out from #304 Tier 4 per the original instruction in that issue: *"Out of primary scope but worth flagging ... File a dedicated follow-up — don't bundle."*

## Background

The Mercury method-index dispatcher does not log when a method index arrives with no registered handler. Two call sites:

- `crates/services/src/base/dispatch.rs` — unhandled SGWPlayer base method falls through to `trace!`, which is below the default ops filter.
- `crates/services/src/cell/dispatch/router.rs` — unhandled cell method falls through to `info!`, which is noisy and lacks the actionable framing.

Both were briefly promoted to `warn!` on branch `fix/issue-304-tier1-silent-drop-logging`. Reverted in that branch to keep scope honest to the issue's PR plan; tracking the change as its own piece of work here.

## Proposed change

- `base/dispatch.rs:204` — `trace!` → `warn!` for unhandled SGWPlayer base method index. Already has structured fields (`%addr`, `msg_id`, `base_method_index`).
- `cell/dispatch/router.rs:75` — `info!` → `warn!` for unhandled cell method index. Already has structured fields (`entity_id`, `method_index`, `args_len`).

## Out of scope here

The full Mercury method-index catalog audit (handler-completeness gaps for known message indices) is a separate, larger piece of work — this issue is just the dispatcher logging level changes.

## Tests

- Unit test that drives the dispatcher with an unknown method index and asserts the dispatcher returns / completes without panicking and without registering as handled. (Log-level assertion would require `tracing-test` infra not yet added to the workspace; defer or add it as part of this work.)

## Cross-refs

- Parent: #304 (Tier 4 entry)

PR #427, opened 2026-05-31
feat(crafting): Phase 1 — CraftingState + persistence + ASP dispatch fix (#53)

Cadacious:
## Summary

Phase 1 of the multi-phase crafting port from Python. Lands the persistent-state shape, DB schema, and the cell-dispatch routing fix so subsequent phases can build on a working foundation. Activity logic (craft / research / alloy / reverse-engineer / spend-ASP / respec) stays stubbed and lands in follow-up PRs.

Grounded in the [Ghidra-evidenced deep dive](https://github.com/SandboxServers/Cimmeria/issues/53#issue-body) at the bottom of #53 — every wire shape, table schema, and dispatch decision below is sourced from there or from the Python reference (`deprecated/python/cell/Crafter.py`).

## What is in scope

- **DB schema** — new `sgw_player_discipline_expertise` table for per-(player, discipline) expertise. Normalised out of the parallel-array shape `sgw_player` already uses for `discipline_ids` / `blueprint_ids` because expertise mutates per-discipline, not per-player. Canonical schema (`db/sgw/Players/Tables/`) + idempotent migration (`db/scripts/`).
- **`cimmeria_entity::crafting::CraftingState`** — discipline list, expertise map, blueprints, ASP, racial paradigm levels. Mirrors Python `Crafter` state. Entity crate stays `sqlx`-free per the project pattern; persistence lives in `services`.
- **`base::crafting::persistence`** — load + save round-trip across `sgw_player` and the new expertise table inside one transaction. Delete-then-insert for expertise rows so respec-driven removals work correctly when Phase 5 lands.
- **Cell dispatch fix** — `SPEND_APPLIED_SCIENCE_POINTS` (index 95) now routes to the crafting handler. The previous range `CRAFT..=RESPEC_CRAFTING` (96-100) silently dropped index 95 into the social arm. Risk callout R6 from the deep dive.
- **Minimum wire-emit surface** — `serialize_on_update_discipline` (entity) + `send_on_update_discipline` (cell) lock the 8-byte `[disciplineId LE][expertise LE]` payload for method 136. Phase 2 mutation paths will call these.

## Out of scope (deferred to Phase 2-5)

- ASP-spend validation + DB mutation (paradigm gate, prerequisite expertise floor, INSERT into expertise table)
- `craft` / `research` / `alloy` / `reverseEngineer` activity logic
- `onUpdateKnownCrafts`, `onUpdateCraftingOptions`, `onUpdateRacialParadigmLevel`, `onCraftingRespecPrompt`, `onDisciplineRespec` wire emission
- World-entry resend bundle for crafting state
- Respec confirm/cancel flow + cost formula
- The two confidence gaps from the deep dive (`CraftingOptions` struct layout, ASP delta-vs-total semantics) need Ghidra verification before Phase 4/5

## Tests added

| File | Type | Bug shape it guards |
|---|---|---|
| `crates/entity/src/crafting.rs` | unit | `CraftingState` defaults + expertise clamp `[0, 100]` |
| `crates/entity/src/crafting.rs` | wire-format | Byte-exact `onUpdateDiscipline` (8 bytes, two i32 LE) |
| `crates/services/src/base/crafting/persistence.rs` | live-DB | Full round-trip persists all fields across both tables |
| `crates/services/src/base/crafting/persistence.rs` | live-DB | Re-save with reduced expertise drops stale rows |
| `crates/services/src/base/crafting/persistence.rs` | live-DB | Missing-player load returns default sentinel, not error |
| `crates/services/src/cell/cell_methods/player/crafting.rs` | unit | Index 95 routes to crafting (the dispatch-range bug fix) |
| `crates/services/src/cell/cell_methods/player/crafting.rs` | fan-out byte | `send_on_update_discipline` enqueues correct method_index + payload |

Live-DB tests self-skip via `require_db_or_skip!` when `DATABASE_URL` is unset, per `crates/services/src/test_support.rs`. Sentinel base `0x7000_2000` (stepped past the `player_load` reservations).

## Test plan

- [x] `cargo test -p cimmeria-entity --lib crafting` — 5/5 pass (entity unit + wire-format)
- [x] `cargo test -p cimmeria-services --lib crafting` — 5/5 pass (cell routing + persistence; live-DB tests self-skip without `DATABASE_URL`)
- [x] `cargo test -p cimmeria-services --lib cell::cell_methods::player::dispatch` — 4/4 pass (existing dispatch tests still green after the range widening)
- [ ] Live-DB run on a host with the bundled Postgres up — author host had no local DB; CI's `ci-live-db` profile will exercise the round-trip. The three live-DB tests passed under `cargo nextest run --profile=ci-live-db` with the `require_db_or_skip!` graceful-skip path.
- [ ] In-game verification deferred to Phase 2-5 when activity logic ships and the client UI has something to render.

## Files touched

- `db/sgw/Players/Tables/sgw_player_discipline_expertise.sql` (new) + `db/database.sql` include
- `db/scripts/add_player_discipline_expertise.sql` (new — idempotent migration)
- `crates/entity/src/crafting.rs` (new) + `crates/entity/src/lib.rs` module reg
- `crates/services/src/base/crafting/mod.rs` (new) + `crates/services/src/base/crafting/persistence.rs` (new) + `crates/services/src/base/mod.rs` module reg
- `crates/services/src/cell/cell_methods/player/crafting.rs` (extended) — `send_on_update_discipline` + `SPEND_APPLIED_SCIENCE_POINTS` arm
- `crates/services/src/cell/cell_methods/player/dispatch.rs` — range widened from `CRAFT..=RESPEC_CRAFTING` to `SPEND_APPLIED_SCIENCE_POINTS..=RESPEC_CRAFTING`; constant-pin test updated

Refs #53

[Generated with Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/entity/src/crafting.rs
+++ b/crates/entity/src/crafting.rs
@@ -1,0 +1,188 @@
+//! Player crafting state — disciplines, expertise, blueprints, paradigms, ASP.
+//!
+//! Mirrors the Python `Crafter` class (`deprecated/python/cell/Crafter.py`)
+//! state half: the values a character carries between sessions and the values
+//! the cell layer mutates during craft/research/alloy/reveng activity.
+//!
+//! Reference: see the deep dive in issue body for wire format, formulas, and
+//! per-activity logic. Phase 1 just defines the struct + a couple of helpers
+//! and leaves the activity mutators stubbed (those land in Phase 2/3).
+//!
+//! Persistence lives in the services crate (see
+//! `crates/services/src/base/crafting/persistence.rs`); the entity crate is
+//! sqlx-free by design.
+
+use std::collections::HashMap;
+
+/// Crafting state carried per player.
+///
+/// Field provenance (cross-referenced with `sgw_player.sql`):
+/// - `discipline_ids` ↔ `sgw_player.discipline_ids integer[]` (which crafting
+///   disciplines the character has unlocked).
+/// - `expertise` ↔ `sgw_player_discipline_expertise` table (normalised
+///   per-(player, discipline) — see the SQL file for rationale).
+/// - `blueprint_ids` ↔ `sgw_player.blueprint_ids integer[]` (recipes the
+///   character has learned).
+/// - `applied_science_points` ↔ `sgw_player.applied_science_points integer`
+///   (ASP currency spendable on discipline unlocks).
+/// - `racial_paradigm_levels` ↔ `sgw_player.racial_paradigm_levels integer[]`,
+///   keyed by paradigm id. Python stored these as a `{paradigm_id -> level}`
+///   map; the SQL array is parallel to the paradigm id sequence. The map
+///   shape here matches Python's intent and keeps the discipline-prerequisite
+///   check (paradigm-id-keyed) ergonomic.
+///
+/// Python stored `discipline_ids` and `expertise` as one `{id -> expertise}`
+/// map. We split them so the wire-emit path can iterate the *known* set in a
+/// deterministic order (the `Vec`) without paying for a sort on every
+/// `onUpdateDiscipline` resend.
+#[derive(Debug, Clone, Default)]
+pub struct CraftingState {
+    /// Disciplines the player has unlocked, in insertion order. Parallel to
+    /// `sgw_player.discipline_ids`. Iteration order is preserved across
+    /// reloads because `load_from_db` orders by row id / discipline_id.
+    pub discipline_ids: Vec<i32>,
+
+    /// Expertise per discipline, in `[0, 100]`. Only contains entries for
+    /// disciplines in `discipline_ids` — looking up an unknown discipline
+    /// returns `None`. Python parity: `Crafter.disciplines[id]` raised
+    /// `KeyError` for unknowns; here, callers must check `discipline_ids`
+    /// first or use the `.get()` accessor.
+    pub expertise: HashMap<i32, i32>,
+
+    /// Blueprints (crafting recipes) the player has learned. Mirrors
+    /// `sgw_player.blueprint_ids`. Sent to the client via
+    /// `onUpdateKnownCrafts` (method 139) on world entry.
+    pub blueprint_ids: Vec<i32>,
+
+    /// Spendable Applied Science Points. Each `spendAppliedSciencePoints`
+    /// call consumes 1; content-engine actions (level-up, mission rewards)
+    /// add to this. Wire syncing uses `onEntityProperty(type=2, value)`,
+    /// not a dedicated message — see the deep dive in #53.
+    pub applied_science_points: i32,
+
+    /// Racial paradigm levels, keyed by paradigm id. Discipline unlocks
+    /// gate on `racial_paradigm_levels[discipline.racial_paradigm_id] >=
+    /// discipline.racial_paradigm_level`. Initial value for every paradigm
+    /// on character creation is 1 (Python `Crafter.__init__`).
+    ///
+    /// `i8` matches the wire encoding of `onUpdateRacialParadigmLevel`
+    /// (method 138, payload `INT32 paradigmId, INT8 level`). The Python
+    /// source caps levels at 5, so `i8` is comfortably oversized.
+    pub racial_paradigm_levels: HashMap<i32, i8>,
+}
+
+impl CraftingState {
+    /// Empty default state. New characters get this until their first
+    /// `spendAppliedSciencePoints` call (which inserts a discipline +
+    /// expertise=1 row).
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Get the current expertise for a discipline, if the player has
+    /// learned it. Returns `None` for unknown disciplines (matches
+    /// "is this in `discipline_ids`?" without forcing the caller to
+    /// scan the Vec).
+    pub fn get_expertise(&self, discipline_id: i32) -> Option<i32> {
+        self.expertise.get(&discipline_id).copied()
+    }
+
+    /// Set expertise for a discipline, clamped to `[0, 100]` per Python's
+    /// `gainExpertise` hard cap. Does *not* update `discipline_ids`; the
+    /// caller is responsible for adding the discipline to the known list
+    /// when it transitions from unknown to known. (Phase 2 work — Phase 1
+    /// only exposes the primitive.)
+    pub fn set_expertise(&mut self, discipline_id: i32, value: i32) {
+        let clamped = value.clamp(0, 100);
+        self.expertise.insert(discipline_id, clamped);
+    }
+}
+
+// ── Wire-format serializers (Phase 1 minimum) ───────────────────────────────
+//
+// The mutation paths that *call* these land in Phase 2. We define the
+// serializer here in Phase 1 so:
+//   1. The wire shape is locked in by a test (byte-exact regression guard).
+//   2. The world-entry resend path can call it without forward-referencing
+//      a Phase 2 module.
+//
+// Method indices and wire shapes are sourced from
+// `docs/protocol/client-method-dispatch-table.md` and the deep dive in #53.
+
+/// Serialize the `onUpdateDiscipline` args (method index 136).
+///
+/// Wire shape: `[disciplineSeqId: i32 LE][expertise: i32 LE]` — 8 bytes total.
+///
+/// Source: `docs/protocol/client-method-dispatch-table.md` row for index 136.
+/// Python emit site: `Crafter.onUpdateDiscipline` (called from `gainExpertise`
+/// and `spendAppliedSciencePoints`).
+pub fn serialize_on_update_discipline(discipline_id: i32, expertise: i32) -> Vec<u8> {
+    let mut buf = Vec::with_capacity(8);
+    buf.extend_from_slice(&discipline_id.to_le_bytes());
+    buf.extend_from_slice(&expertise.to_le_bytes());
+    buf
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn default_state_is_empty() {
+        let s = CraftingState::new();
+        assert!(s.discipline_ids.is_empty());
+        assert!(s.expertise.is_empty());
+        assert!(s.blueprint_ids.is_empty());
+        assert_eq!(s.applied_science_points, 0);
+        assert!(s.racial_paradigm_levels.is_empty());
+    }
+
+    #[test]
+    fn get_expertise_returns_none_for_unknown() {
+        let s = CraftingState::new();
+        assert_eq!(s.get_expertise(42), None);
+    }
+
+    #[test]
+    fn set_expertise_clamps_to_cap() {
+        let mut s = CraftingState::new();
+        s.set_expertise(5, 150);
+        assert_eq!(s.get_expertise(5), Some(100));
+
+        s.set_expertise(5, -10);
+        assert_eq!(s.get_expertise(5), Some(0));
+
+        s.set_expertise(5, 73);
+        assert_eq!(s.get_expertise(5), Some(73));
+    }
+
+    /// Byte-exact regression guard for `onUpdateDiscipline` (method 136).
+    ///
+    /// Wire shape per the deep dive: `INT32 disciplineSeqId, INT32 expertise`.
+    /// A regression here desyncs the client's discipline UI (wrong row,
+    /// wrong percentage) — the client trusts both fields without
+    /// validation. A swap or width change would manifest as the wrong
+    /// discipline lighting up at the wrong percent.
+    #[test]
+    fn on_update_discipline_byte_layout_matches_spec() {
+        // discipline_id = 0x11223344, expertise = 0x55667788
+        let bytes = serialize_on_update_discipline(0x11223344, 0x55667788);
+        assert_eq!(bytes.len(), 8, "onUpdateDiscipline must be exactly 8 bytes");
+        // LE: low byte first.
+        assert_eq!(
+            bytes,
+            vec![0x44, 0x33, 0x22, 0x11, 0x88, 0x77, 0x66, 0x55],
+            "field order: disciplineSeqId then expertise, both i32 LE",
+        );
+    }
+
+    /// Edge case: expertise = 0 (transitional state immediately after
+    /// `spendAppliedSciencePoints` but before the initial-value INSERT)
+    /// and discipline_id = 0 (sentinel — should still serialize without
+    /// special-casing).
+    #[test]
+    fn on_update_discipline_handles_zero_values() {
+        let bytes = serialize_on_update_discipline(0, 0);
+        assert_eq!(bytes, vec![0; 8]);
+    }
+}
--- a/crates/entity/src/crafting.rs
+++ b/crates/entity/src/crafting.rs
@@ -4,9 +4,8 @@
 //! state half: the values a character carries between sessions and the values
 //! the cell layer mutates during craft/research/alloy/reveng activity.
 //!
-//! Reference: see the deep dive in issue body for wire format, formulas, and
-//! per-activity logic. Phase 1 just defines the struct + a couple of helpers
-//! and leaves the activity mutators stubbed (those land in Phase 2/3).
+//! Phase 1 just defines the struct + a couple of helpers and leaves the
+//! activity mutators stubbed (those land in Phase 2/3).
 //!
 //! Persistence lives in the services crate (see
 //! `crates/services/src/base/crafting/persistence.rs`); the entity crate is
@@ -38,8 +37,12 @@ use std::collections::HashMap;
 #[derive(Debug, Clone, Default)]
 pub struct CraftingState {
     /// Disciplines the player has unlocked, in insertion order. Parallel to
-    /// `sgw_player.discipline_ids`. Iteration order is preserved across
-    /// reloads because `load_from_db` orders by row id / discipline_id.
+    /// `sgw_player.discipline_ids` (a Postgres `integer[]`). Persistence
+    /// lives in `crates/services/src/base/crafting/persistence.rs` —
+    /// `load_crafting_state` reads the array verbatim, so reload order is
+    /// whatever order was on the row when last saved. (The parallel
+    /// expertise rows in `sgw_player_discipline_expertise` are loaded
+    /// `ORDER BY discipline_id`, but those land in `expertise`, not here.)
     pub discipline_ids: Vec<i32>,
 
     /// Expertise per discipline, in `[0, 100]`. Only contains entries for
@@ -57,7 +60,7 @@ pub struct CraftingState {
     /// Spendable Applied Science Points. Each `spendAppliedSciencePoints`
     /// call consumes 1; content-engine actions (level-up, mission rewards)
     /// add to this. Wire syncing uses `onEntityProperty(type=2, value)`,
-    /// not a dedicated message — see the deep dive in #53.
+    /// not a dedicated message.
     pub applied_science_points: i32,
 
     /// Racial paradigm levels, keyed by paradigm id. Discipline unlocks
@@ -107,7 +110,7 @@ impl CraftingState {
 //      a Phase 2 module.
 //
 // Method indices and wire shapes are sourced from
-// `docs/protocol/client-method-dispatch-table.md` and the deep dive in #53.
+// `docs/protocol/client-method-dispatch-table.md`.
 
 /// Serialize the `onUpdateDiscipline` args (method index 136).
 ///
--- a/crates/entity/src/lib.rs
+++ b/crates/entity/src/lib.rs
@@ -13,6 +13,7 @@
 pub mod abilities;
 pub mod base_entity;
 pub mod cell_entity;
+pub mod crafting;
 pub mod detour_ffi;
 pub mod interaction_flags;
 pub mod inventory;
--- a/crates/services/src/base/crafting/mod.rs
+++ b/crates/services/src/base/crafting/mod.rs
@@ -1,0 +1,19 @@
+//! Crafting subsystem — BaseApp side.
+//!
+//! Phase 1 (this module) provides persistence helpers for `CraftingState`:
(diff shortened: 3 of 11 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `persistence.rs` `save_crafting_state` DELETE-then-INSERT race | **The save uses `DELETE FROM sgw_player_discipline_expertise WHERE player_id = $1` then loops INSERTs.** If the transaction is interrupted between DELETE and INSERT, the player loses all expertise. **The transaction should protect this, but verify the `tx.commit()` is the ONLY exit path after DELETE.** Looking at the code, it is — but a future refactor adding an early return would break this. **Consider adding a comment warning about this invariant.** |
| **G2** | `persistence.rs` `racial_paradigm_levels` array indexing | **The save converts the HashMap to a Vec by iterating `1..=max_paradigm_id`.** If a paradigm is deleted from `resources.racial_paradigm` but the player still has a level for it, the save would backfill with 0 and log a warning — but **the load would then re-key it as paradigm_id = (index + 1), which is wrong if the paradigm IDs are non-contiguous.** |
| **G3** | `crafting.rs` `set_expertise` does not update `discipline_ids` | **The doc says "the caller is responsible for adding the discipline to the known list."** But if a caller forgets, the player would have expertise for a discipline they do not "know" — the client would show 0% because the discipline is not in the list. **Consider a helper that does both atomically.** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `persistence.rs` `load_crafting_state` does not validate expertise rows | **The load pulls all expertise rows for the player, but does not check that each `discipline_id` is in `discipline_ids`.** A stray row (data corruption) would be silently loaded. **The comment says "we surface it rather than silently dropping it" but the code does not actually surface it — it just loads it.** |
| **G5** | `dispatch.rs` range comment says `SPEND_APPLIED_SCIENCE_POINTS ≤ CRAFT` | **The comment says "ORG_CREATION ≤ SPEND_APPLIED_SCIENCE_POINTS ≤ CRAFT" but the actual values are 94, 95, 96.** The `≤` is correct, but the comment could be clearer that 95 is SPEND and 96 is CRAFT. |
| **G6** | `crafting.rs` `serialize_on_update_discipline` uses `discipline_id` not `disciplineSeqId` | **The wire spec says `disciplineSeqId` but the function parameter is `discipline_id`.** Are these the same thing? The Python reference would need to be checked. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G7** | `persistence.rs` `#[allow(dead_code)]` on both functions | **Both `load_crafting_state` and `save_crafting_state` are marked `#[allow(dead_code)]`.** This is correct for Phase 1, but **if Phase 2 forgets to remove the annotation, the compiler will not warn about unused code.** |
| **G8** | `sgw_player_discipline_expertise.sql` no index on `discipline_id` | **The table has a composite PK `(player_id, discipline_id)`, which indexes by player_id first.** A query like `SELECT * FROM ... WHERE discipline_id = $1` (e.g., "find all players with this discipline") would be slow. **Is this query ever needed? If so, add an index.** |
| **G9** | `crafting.rs` `racial_paradigm_levels` uses `i8` for level | **The doc says "Python source caps levels at 5, so i8 is comfortably oversized."** But **what if a future content expansion raises the cap to 200?** The wire encoding is `INT8`, so this is probably fine, but document the wire constraint. |

MyPalClara:

| **G10** | No test for `load_crafting_state` with stray expertise row | The code does not validate that expertise rows match `discipline_ids`, but there is no test for this edge case. |
| **G11** | No test for `save_crafting_state` with non-contiguous paradigm IDs | The backfill logic assumes contiguous IDs 1..=max. A test with IDs [1, 3, 5] would expose the gap-handling behavior. |
| **G12** | No test for `CRAFT` (96) routing | The test covers `SPEND_APPLIED_SCIENCE_POINTS` (95), but not the other crafting indices (96-100). |

MyPalClara:

| **G13** | `persistence.rs` does not document the transaction isolation level | The code uses `pool.begin()` but does not specify the isolation level. Is `READ COMMITTED` sufficient, or does this need `SERIALIZABLE`? |
| **G14** | `crafting.rs` `CraftingState` does not document the world-entry resend path | The struct doc mentions "sent to the client via `onUpdateKnownCrafts`" but does not say when or how. |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 3 |
| Correctness issues | 3 |
| Potential issues | 3 |
| Missing test coverage | 3 |
| Documentation | 2 |

**Total: 14 gaps**

Cadacious:
## Review follow-up landed: 3 commits

### CI fix (P0)
- **`fix(db): move sgw_player_discipline_expertise FK to _foreign_keys.sql`** (`135da8e2`) — The inline `FOREIGN KEY (player_id) REFERENCES sgw_player(player_id)` in `sgw_player_discipline_expertise.sql` was failing schema load because `sgw_player`'s PK constraint isn't established until `_primary_keys.sql` runs (per the repo's split convention). Moved the FK to `db/sgw/_foreign_keys.sql` as `ALTER TABLE … ADD CONSTRAINT`, which runs after `_primary_keys.sql`. The idempotent migration in `db/scripts/add_player_discipline_expertise.sql` keeps its inline FK — it only runs against a DB where the PK already exists.

### Bug fixes + style + Clara G1/G5 (P1+P2+P3)
- **`fix(crafting): silent-save guard, dedup constant, doc + comment cleanups`** (`a1cd463d`)
  - **Copilot Bug B**: `save_crafting_state` now captures `result.rows_affected()` on the UPDATE; returns `sqlx::Error::RowNotFound` when 0 rows match. Negative-logs at `error` level per the negative-logging convention.
  - **Copilot Bug A**: Removed the duplicate `const ON_UPDATE_DISCIPLINE: u16 = 136` in `cell_methods/player/crafting.rs`; now imports the canonical one from `cell::client_methods::player`.
  - **Copilot Bug C**: Rewrote the misleading doc on `CraftingState::discipline_ids`. The old text claimed `load_from_db` orders by row id, but `load_from_db` doesn't exist in the entity crate and `discipline_ids` is loaded verbatim from the Postgres `integer[]` column.
  - **Clara G1**: Added an invariant doc block on `save_crafting_state` calling out the DELETE-then-commit contract — a future Ok(()) early return between DELETE and commit would silently wipe expertise.
  - **Clara G5**: Replaced the prose "implicit constant ordering" comment in `cell_methods/player/dispatch.rs` with a compile-time `const _: () = assert!(...)` that actually pins `ORG_CREATION ≤ SPEND_APPLIED_SCIENCE_POINTS ≤ CRAFT ≤ RESPEC_CRAFTING ≤ CANCEL_MOVIE`. A renumber that breaks the invariant now fails the build.
  - **Copilot P2 / style**: Stripped `#53`, `#311` issue numbers from `entity/crafting.rs`, `services/base/crafting/mod.rs`, `services/cell/.../crafting.rs` per `.github/copilot-instructions.md`.

### Tests (P4)
- **`test(crafting): regression guards for save-missing-player, stray expertise, index 96`** (`e95b4df1`)
  - **`save_for_nonexistent_player_returns_error`** (live-DB) — reverting the `rows_affected() == 0` check flips this test pass → fail. Also asserts `COUNT(*) = 0` on the expertise table to verify the txn rolled back cleanly.
  - **`load_with_stray_expertise_row_keeps_it`** (live-DB, Clara G10) — pins the current "load surfaces every expertise row for this player regardless of `discipline_ids` membership" contract. A future change to filter must flip the assertion deliberately.
  - **`craft_routes_to_crafting`** (unit, Clara G12) — companion to the existing index-95 routing test; pins that index 96 (CRAFT) also reaches the crafting handler.

### Verified locally
- `cargo fmt --check -p cimmeria-entity -p cimmeria-services` ✓
- `cargo clippy --all-targets -p cimmeria-entity -p cimmeria-services -- -D warnings` ✓
- `cargo test -p cimmeria-services --lib cell::cell_methods::player::crafting:: base::crafting::` ✓ (5 persistence tests pass without DB by self-skipping via `require_db_or_skip!`; 3 unit tests in the cell layer pass; expect the live-DB CI run to exercise the 5 persistence tests for real)
- `cargo test -p cimmeria-entity --lib crafting` ✓ (5 unit tests)
- `cargo test -p cimmeria-services --lib cell::cell_methods::player::dispatch::` ✓ (4 routing tests, including the new const_assert)

### Skipped (with rationale)

- **Clara G7** (`#[allow(dead_code)]` on Phase 1 functions) — by design; Phase 2 wires them up. Removing the allow now produces noise without behavior change.
- **Clara G8** (no index on `discipline_id` alone) — premature; no current query targets `discipline_id` independent of `player_id`. The composite PK covers the actual access pattern.
- **Clara G11** (paradigm IDs non-contiguous test) — verified that the existing `save_crafting_state` *does* assume contiguous paradigm IDs from 1 to `max(map.keys())`, but it backfills missing IDs with 0 and emits a warn log. The seed has 5 contiguous paradigms (1..=5). No additional test needed unless we tighten the contract.
- **Clara G13** (transaction isolation level) — default `READ COMMITTED` is correct for this access pattern; no behavior change needed.
- **Clara G14** (world-entry resend path) — Phase 4 territory per the deep dive.

### CI expectation

The three CI failures were rooted in the FK ordering bug in `db/sgw/Players/Tables/sgw_player_discipline_expertise.sql` (which prevented schema load entirely, so `cargo nextest -p cimmeria-services (live DB)`, `cargo llvm-cov`, and `build + smoke-test` all failed at psql-load time). With the FK now in `_foreign_keys.sql`, the schema should load and all three jobs should go green. The new `save_for_nonexistent_player_returns_error` and `load_with_stray_expertise_row_keeps_it` tests run under the `ci-live-db` profile and self-skip without DATABASE_URL.

Cadacious:
## Adversarial trust-boundary review (server-authority-enforcer)

**Verdict: SHIP — clean adversarial review.**

Reviewed every client-supplied field in the touched files (`cell/cell_methods/player/crafting.rs`, `cell/cell_methods/player/dispatch.rs`, `base/crafting/persistence.rs`) against the "what if the client lies?" framework. No exploitable surface in Phase 1.

### Handler audit: `cell/cell_methods/player/crafting.rs::dispatch`

**Client-asserted fields**
- `method_index: u16` (router-supplied, but originates from a client packet)
- `args: &[u8]` (raw client buffer; cast to `i32` LE in the handlers)
- `entity_id: u32` (NOT client-supplied — resolved upstream in `cell/dispatch/router.rs:33` from session state)

**Validation audit**
- `method_index` → dispatch range pinned by static `const _: () = assert!(...)` (dispatch.rs:12-19) + the `cell_method_constants_pin_expected_values` test (dispatch.rs:75). Cannot drift silently.
- `args` → `args.len() >= 4` bounds check before every `i32::from_le_bytes(...)`. The truncated path warn-logs and still returns true (intentional — silent-drop is already the failure shape that the wider dispatch sweep was meant to surface).
- `entity_id` → server-authoritative (origin in the encrypted dispatcher's session lookup, see `base/connect_loop/cell_arms.rs:26`).

**Mutation analysis**
Every arm (SPEND_APPLIED_SCIENCE_POINTS, CRAFT, RESEARCH, REVERSE_ENGINEER, ALLOYING, RESPEC_CRAFTING) is an UNIMPLEMENTED stub that `tracing::info!`-logs and returns true. **No state mutation, no item/currency mutation.** A malicious client gains nothing by calling these handlers in Phase 1 except producing a log line.

### Handler audit: `base/crafting/persistence.rs`

`load_crafting_state` / `save_crafting_state` are not yet wired into a request-path caller (Phase 2). They're well-defended for when that arrives:

- `save_crafting_state` properly serialises both tables inside one transaction; the DELETE-then-INSERT pattern is correct for the upcoming respec path (Phase 5).
- The `rows_affected() == 0` → `RowNotFound` guard on the player UPDATE closes the silent-data-loss window for a missing player_id. Regression test `save_for_nonexistent_player_returns_error` pins this — revert the guard and it fails.
- `load_crafting_state` clamps `i32 → i8` for paradigm levels defensively against corrupted DB rows.
- `set_expertise` clamps to `[0, 100]` per the Python `gainExpertise` cap.

### Dispatch range widening (the actual security-relevant change)

The previous range `CRAFT..=RESPEC_CRAFTING` (96..=100) silently dropped index 95 (`SPEND_APPLIED_SCIENCE_POINTS`) into the social arm, which had no case for it. The widening to `SPEND_APPLIED_SCIENCE_POINTS..=RESPEC_CRAFTING` (95..=100) is **correctly scoped**:

- 94 (`ORG_CREATION`) still routes to social (as it should — it's an org-creation method, not crafting).
- 95 now reaches the crafting handler.
- The static assertion `ORG_CREATION ≤ SPEND_APPLIED_SCIENCE_POINTS ≤ CRAFT ≤ RESPEC_CRAFTING ≤ CANCEL_MOVIE` will fail the build if any constant renumber violates the layout.

Regression tests `spend_applied_science_points_routes_to_crafting` + `craft_routes_to_crafting` pin both ends.

### Known-hazardous patterns — checked, none apply

- ❌ No `remove_item` adjacent to `UseInventoryItem` chain — no item mutation in Phase 1 at all.
- ❌ No `type_id`-keyed multi-step transaction — no item rows touched.
- ❌ No client-supplied timestamp used as a time base.
- ❌ No GM-bit read from the inbound packet.
- ❌ No client-supplied quantity treated as authoritative.

### Phase 2 forward-look (not blocking this PR)

When ASP-spend actually mutates state, the conditional UPDATE pattern is the right shape — `UPDATE sgw_player SET applied_science_points = applied_science_points - 1 WHERE player_id = $1 AND applied_science_points >= 1` with `rows_affected() == 1` check. A naive `state.applied_science_points -= 1; save_crafting_state(...)` is racy against two concurrent ASP-spend packets. Worth a regression guard at Phase 2 — happy to re-review when it lands.

🤖 Generated by server-authority-enforcer adversarial-review pass.

Cadacious:

Adversarial verification of the regression guards added in commits `a1cd463d` (fixes) and `e95b4df1` (tests). For each guard I reverted the corresponding fix locally (in worktree, not pushed) and re-ran the test to check whether the bug shape actually fails. Found one verifiable guard, one suspect, and one clear theatre case — and I pushed a replacement test that does pin the bug.

- Test: `cell::cell_methods::player::crafting::tests::spend_applied_science_points_routes_to_crafting` — **THEATRE**
  - Bug shape pinned: no
  - Verification method: reverted `dispatch.rs` outer-router range from `SPEND_APPLIED_SCIENCE_POINTS..=RESPEC_CRAFTING` back to `CRAFT..=RESPEC_CRAFTING` (the original bug shape) and ran the test.
  - Result: **test still passed.** The test calls `crafting::dispatch` directly (the inner submodule), which has always had a `SPEND_APPLIED_SCIENCE_POINTS` arm — the bug was purely in the *outer* router. Bypassing the outer router means this test would silently green-light a regression of the very fix it claims to guard.
  - Compounding: `social::dispatch` *also* has its own `SPEND_APPLIED_SCIENCE_POINTS` arm (at `social.rs:66`) that returns `true`. So even rewriting this test to use the outer dispatcher and asserting `handled == true` would not catch the bug — both branches return `true`.

- Test: `cell::cell_methods::player::crafting::tests::craft_routes_to_crafting` (added in `e95b4df1`) — **THEATRE**
  - Bug shape pinned: no — same defect as the 95-test above. The docstring's claim that "A routing regression that narrowed the range to `CRAFT..=RESPEC_CRAFTING - 1` would still pass the 95 test if 95 were left in the range, but would fail this one" is wrong — the inner `crafting::dispatch` has always covered `CRAFT`, so narrowing the outer router has zero effect on this test.
  - Verification method: same revert as above; test still passed.
  - Suggested fix: I pushed commit `fc821bd6` which adds `outer_dispatch_routes_spend_asp_to_crafting_not_social` to `dispatch.rs`'s test module. It calls the outer dispatcher with index 95 and uses `LogCapture` to assert that the *crafting*-side log fires (`"UNIMPLEMENTED: spendAppliedSciencePoints (Phase 2)"` — the social-side stub omits the `(Phase 2)` suffix). Verified to fail when the outer-router range is reverted to `CRAFT..=RESPEC_CRAFTING` (captured event shows `cimmeria_services::cell::cell_methods::player::social` emitting the social-side log). Passes with the fix restored.

- Test: `base::crafting::persistence::tests::save_for_nonexistent_player_returns_error` — **SUSPECT (fails for the wrong reason)**
  - Bug shape pinned: partially — the test does fail when the `rows_affected() == 0` guard is reverted, but **not via the silent-success path the bug shape describes**. The test seeds non-empty expertise (`state.expertise.insert(1, 50); .insert(2, 75)`), which forces an INSERT into `sgw_player_discipline_expertise`. Without the guard, the INSERT fails with a FK violation (player does not exist in `sgw_player`), so the test sees `Err(sqlx::Error::Database)` and trips the "wrong variant" panic arm — not the `Ok(())` arm the docstring is written around.
  - Verification method: static analysis (no local Postgres on `:5433`; live-DB tests self-skip locally).
  - The actual silent-success bug shape (the docstring's `Ok(())` arm) only fires when `state.expertise.is_empty()` — at which point the txn has nothing to INSERT, the DELETE is a no-op, and the UPDATE matching 0 rows commits clean. The test inadvertently masks the silent-success variant behind a FK violation.
  - Hygiene gap: the test does not call `cleanup(&pool, _, bogus_player_id)` before running. A prior failed run that managed to leave expertise rows at `TEST_BASE + 1999` (via some FK-deferring path, or sentinel reuse) would corrupt the COUNT-on-rollback assertion. Per TESTING.md: sentinels are cleaned by exact id, not range — add a `cleanup_expertise(&pool, bogus_player_id)` call at the top.
  - Suggested fix: split into two tests — one with empty expertise (proves the silent-success path is now an Err), one with non-empty expertise (proves the FK violation also surfaces as Err but for the documented reason). Or: keep one test but add a follow-up call with `state.expertise.clear()` after the first assertion to also exercise the empty-state silent-success shape.

- Test: `base::crafting::persistence::tests::load_with_stray_expertise_row_keeps_it` — **NOT A REGRESSION GUARD (BEHAVIOR LOCK-IN, BUT WELL-SHAPED)**
  - This is not framed as a regression guard for the PR's fixes — it documents `load_crafting_state`'s "load everything for the player, do not filter by `discipline_ids`" contract. The assertion (`state.get_expertise(99) == Some(50)` with `discipline_ids = [7]`) directly pins this behavior. Good lock-in; flips deliberately if a future refactor decides to filter.
  - One minor thing: the test seeds via raw SQL `UPDATE` for `discipline_ids` and raw `INSERT` for the expertise row, but cleans up via the helper. Fine — the SQL is local to the test and intentional. The cleanup pattern matches the rest of the file.

### Strong examples

- Entity-crate tests (`crates/entity/src/crafting.rs`) are tight: `set_expertise_clamps_to_cap` pins three concrete cases (150 → 100, -10 → 0, 73 → 73). `on_update_discipline_byte_layout_matches_spec` asserts exact 8-byte payload `[0x44, 0x33, 0x22, 0x11, 0x88, 0x77, 0x66, 0x55]` — byte-exact wire-format guard. No theatre here.
- `send_on_update_discipline_emits_correct_message` is a clean fan-out byte test: asserts `entity_id == 42`, `method_index == 136` (exact constant, not a range), and exact 8-byte LE payload `[0x07, 0x00, 0x00, 0x00, 0x32, 0x00, 0x00, 0x00]`. Pins the end-to-end wire shape from `serialize_on_update_discipline` through `send_on_update_discipline` into the right `CellToBaseMsg` variant.
- `save_replaces_expertise_rows` is a strong live-DB guard: it saves with 3 disciplines, drops discipline 2, re-saves, and asserts `get_expertise(2) == None`. Catches a refactor from DELETE-then-INSERT to plain INSERT directly.

### Coverage gaps the previous reviews missed

1. **No outer-dispatcher regression guard** for the SPEND_APPLIED_SCIENCE_POINTS routing bug — until the commit I just pushed. The two existing tests (`spend_applied_science_points_routes_to_crafting`, `craft_routes_to_crafting`) called the inner submodule directly, and `each_outer_range_routes_to_a_handler` in `dispatch.rs` probes `CRAFT` (96) but not `SPEND_APPLIED_SCIENCE_POINTS` (95). The bug was invisible to every test in the PR before my push.
2. **No empty-expertise variant of `save_for_nonexistent_player_returns_error`** — the test as written catches the regression via FK violation, not the documented silent-success path. The exact bug shape the PR docstring describes ("returns Ok and silently throws away the in-memory mutation") only fires when expertise is empty; the test never hits that branch.
3. **No test exercising the `tx.commit()`-after-DELETE invariant** named in the new doc block on `save_crafting_state` — the doc describes a "future refactor adds Ok(()) early return between DELETE and commit" failure mode. No test exercises this path. Hard to write (you would need fault injection), but a `#[ignore]` placeholder with a TODO would at least mark the gap.
4. **`save_for_nonexistent_player_returns_error` lacks the `cleanup(&pool, _, TEST_BASE + 1999)` call** that the other live-DB tests in this module use. Minor — but per TESTING.md the convention is exact-sentinel cleanup before each test.

### Verdict

- PR #427: **NEEDS TEST FIX** (or accept the replacement test I just pushed)
  - The two routing tests (`spend_applied_science_points_routes_to_crafting`, `craft_routes_to_crafting`) are theatre — they cannot detect the bug they claim to guard. They are not actively harmful (they still pin that the *inner* dispatch handles 95 and 96), but their docstrings overclaim. Recommended actions:
    - Keep the new `outer_dispatch_routes_spend_asp_to_crafting_not_social` test I pushed (commit `fc821bd6`) — it actually pins the outer-router fix.
    - Soften the docstrings on `spend_applied_science_points_routes_to_crafting` and `craft_routes_to_crafting` to reflect that they pin the *inner* submodule's arm coverage, not the outer router's range. Or delete them — they are redundant once the LogCapture-based outer test is in place.
    - `save_for_nonexistent_player_returns_error`: either split into empty-expertise + non-empty variants, or accept that it catches the regression via FK rather than silent-success and update the docstring accordingly. Add the missing `cleanup` call.
  - The entity-crate tests, the byte-exact wire test, and the live-DB round-trip + replace tests are all strong and ready to merge as is.

PR #428, opened 2026-05-31
feat(cell, content-engine): NPC AI phases 2-7 (Patrol + Wander + Investigating + Follow + Despawning/Submit/Error stubs)

### LINKED ISSUE #48: NPC AI state machine (10 missing behavior states)

Cadacious:
## Summary

Closes the rest of #48 — Phases 2 + 3 + 4 + 6 + 7 in one PR per the planning conversation. Builds on #423 (Phase 1 + 5: state machine + setMovementType + NPC respawn).

5 commits, one per phase, each with its own tests:

- **Phase 2 — Patrol** (`d1be1abd`): DB join → `entity_templates.patrol_path_id` + `point_set_points`; runtime waypoint loop with dwell pauses; threat preemption preserves index across Fighting → Leashing → Patrol.
- **Phase 3 — Wander** (`468838fb`): schema columns + CHECK constraints; seeded RNG samples within `wander_radius` of spawn; navmesh validation with fallback to spawn position.
- **Phase 4 — Investigating** (`524abff6`): new `Action::SetNpcPoi { entity_tag, x, y, z }` content action; pathfind → 5s dwell → Idle.
- **Phase 6 — Follow** (`d3240467`): new `Action::SetFollowTarget { entity_tag, target_tag }`; distance-band maintenance (`[2, 5]` default); target-gone auto-cleanup.
- **Phase 7 stubs** (`4630371c`): new `Action::SetNpcAiState { entity_tag, state }` with `NpcAiStateAction` enum admitting Idle/Despawning/Submit/Error. Despawning removes the entity; Submit clears combat state and holds; Error is quiescent.

## Schema migrations

Five new nullable columns on `entity_templates`, all with CHECK constraints:

```sql
wander_radius            real,
wander_min_dwell_secs    real,
wander_max_dwell_secs    real,
follow_min_distance      real,
follow_max_distance      real,
-- + CHECK constraints: positive only, min ≤ max
```

Existing seed rows stay NULL → runtime defaults apply, behavior unchanged for any current content.

## State machine layout post-PR

`generate_threat` now preempts any non-Dead non-Fighting state to Fighting (not just Idle), preserving the per-state scratch fields so the post-fight return to Idle can re-route via the AI tick's transition logic.

AI tick dispatch (priority order in Idle): aggression > patrol > wander.

Each behavior state broadcasts its `MobMovementType` byte on entry, dedup'd against `last_movement_type`:

| State | Wire byte |
|---|---|
| Fighting (entry) | CombatAdvance |
| Leashing (entry) | Leash |
| Patrol (entry) | Patrol |
| Wander (entry) | Wander |
| Follow (entry) | Follow |
| Investigating (entry) | CombatAdvance (closest match — no investigate byte in EMobMovementType) |
| Idle / Submit / Despawning (entry) | None (clears cache) |
| Dead / Spawning / Error | None (no transition fires from these) |

## Test plan

- [x] `cargo clippy --workspace --exclude {GUI} --all-targets -- -D warnings`
- [x] `cargo test -p cimmeria-{entity,content-engine,services} --lib` — **174 / 81 / 1169 passed, 0 failed, 24 new**
- [ ] Live-DB tests — CI will exercise; couldn't run locally
- [ ] **Manual in-game** (any subset is useful):
  - [ ] Set `patrol_path_id` on a Castle_CellBlock guard template + populate a 3-point set in `point_set_points` → guard walks the loop
  - [ ] Set `wander_radius = 5.0` on a template → mob wanders within range of spawn
  - [ ] Fire `SetNpcPoi` from a chain → tagged NPC walks to the POI, dwells, returns to Idle
  - [ ] Fire `SetFollowTarget` from a chain → tagged NPC follows the player at 2-5 units
  - [ ] Fire `SetNpcAiState { state: Despawning }` → tagged NPC disappears
  - [ ] Aggro a patrolling guard → fight breaks out, on leash-back patrol resumes from the previous waypoint

## New tests (24 across 5 commits)

| File | Count | Pins |
|---|---|---|
| `npc_ai_patrol.rs` | 5 | same-tick transition, dwell-advance, threat preemption, empty-path drop, future-dwell no-op |
| `npc_ai_wander.rs` | 5 | same-tick transition, future-deadline no-op, zero-radius drop, active-nav no-op, patrol-priority |
| `npc_ai_investigate.rs` | 4 | first-tick routing, elapsed-dwell return, threat preemption, no-POI drop |
| `npc_ai_follow.rs` | 6 | out-of-band, in-band, below-min, gone-target, no-target, threat preemption |
| `npc_ai_phase7.rs` | 3 | Despawning removes entity, Submit clears combat, Error inert |

## Out of scope (deferred to follow-ups)

- `onNoise` cell-method hook for in-game audio-triggered Investigating (currently reachable only via content action)
- Patrol path content authoring (all seed rows still NULL on `patrol_path_id`)
- Cover system (#209)
- Mob groups, behavior event sets (no engine hooks yet)

## Closes

#48 (Phases 2-7) — completes the NPC AI state machine work. Phase 1+5 shipped in #423.

## Summary by CodeRabbit

* **New Features**
  * NPC patrols, wander, investigate, and follow behaviors with configurable tuning; creators can command NPC AI states.

* **Improvements**
  * NPCs now preempt and enter combat from any active state; follow enforces distance bands and preserves targets across preemption; spawn templates map movement/tuning fields.

* **Tests**
  * Extensive unit/integration coverage added for patrol, wander, investigate, follow, phase transitions, and action parsing.

* **Documentation**
  * NPC AI docs updated to reflect implemented states and behavior.


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `threat.rs` `generate_threat` clears `nav_path` on preemption | **The code does `target.nav_path.clear()` when transitioning to Fighting.** But what if the NPC was mid-pathfind and the pathfinder is still computing? **Is there a race where the pathfinder writes to `nav_path` after the clear?** |
| **G2** | `cell_entity/mod.rs` `patrol_next_index` is `usize` but `patrol_path` is `Vec<Vector3>` | **If `patrol_path` is empty, `patrol_next_index` could be any value.** The patrol handler should check `patrol_path.is_empty()` before indexing. **Does it?** |
| **G3** | `actions.rs` `SetNpcAiState` allows `Idle` | **The doc says "Idle — clean fallback that lets the AI tick re-route."** But if a content author fires `SetNpcAiState { state: Idle }` on an NPC that is mid-patrol, **does the patrol resume or reset?** The patrol index persists, so it should resume — but this is not tested. |
| **G4** | `entity_templates.sql` CHECK constraints allow NULL | **The constraints are `wander_radius IS NULL OR wander_radius > 0.0`.** This means `wander_radius = 0.0` is rejected, but `wander_radius = NULL` is allowed. **Is `NULL` semantically different from `0.0`?** The runtime treats both as "no wander" — but the CHECK rejects 0.0 while allowing NULL. **Inconsistent.** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G5** | `cell_entity/mod.rs` `wander_next_at` uses `std::time::Instant` | **Same concern as PR #420 G17 and PR #423 G3.** `Instant` is not serializable. If the server crashes mid-wander, the deadline is lost. |
| **G6** | `cell_entity/mod.rs` `investigate_until` uses `std::time::Instant` | **Same concern.** |
| **G7** | `cell_entity/mod.rs` `patrol_dwell_until` uses `std::time::Instant` | **Same concern.** |
| **G8** | `threat.rs` preemption does not clear `poi` or `follow_target_id` | **The doc says "per-state scratch persists so the post-fight return can resume."** But if the NPC was Investigating and gets attacked, the `poi` persists. After the fight, the NPC returns to Idle — **does the AI tick re-route to Investigating because `poi` is still set?** If so, the NPC would walk back to the POI after every fight. |
| **G9** | `actions.rs` `NpcAiStateAction` does not include `Dead` | **The enum has Idle/Despawning/Submit/Error but not Dead.** A content author cannot kill an NPC via `SetNpcAiState`. **Is this intentional?** The doc says "Dead is owned by the runtime" — but then how does content kill an NPC? |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G10** | `cell_entity/mod.rs` `follow_min_distance` default is 2.0 | **The doc says "the NPC does not back away when the target gets close."** But what if `follow_min_distance > follow_max_distance` due to a content error? **The CHECK constraint prevents this in the DB, but the runtime defaults could be swapped by a future refactor.** |
| **G11** | `npc-ai.md` says `Investigating` uses `CombatAdvance` | **The doc says "no dedicated investigate byte exists in EMobMovementType."** But **is `CombatAdvance` the right choice?** An investigating NPC should look cautious, not aggressive. **Consider `Wander` or `Patrol` as alternatives.** |
| **G12** | `actions.rs` `SetFollowTarget` with `target_tag = None` clears follow | **The doc says "Passing `target_tag = None` clears the follow state."** But **what if the target_tag resolves to `None` because the tag does not exist?** Is that treated as "clear follow" or "error"? |
| **G13** | `entity_templates.sql` 5 new nullable columns | **All 5 columns are nullable with runtime defaults.** This is correct for backward compatibility, but **the migration script does not add the columns to existing tables** — it only creates the table if it does not exist. **Is there a separate ALTER TABLE migration?** |

MyPalClara:

| **G14** | No test for patrol with empty `patrol_path` | The test `empty-path drop` exists, but does it test the case where `patrol_path` is empty at spawn vs. cleared mid-patrol? |
| **G15** | No test for wander with `wander_radius = 0.0` | The CHECK constraint rejects 0.0, but the runtime default is 0.0. What happens if a template has `wander_radius = NULL` and the runtime defaults to 0.0? |
| **G16** | No test for `SetNpcAiState { state: Idle }` on a patrolling NPC | Does the patrol resume or reset? |
| **G17** | No test for `SetFollowTarget` with non-existent `target_tag` | Is this treated as "clear follow" or "error"? |
| **G18** | No test for threat preemption preserving `poi` | The doc says scratch persists, but is this tested? |

MyPalClara:

| **G19** | `cell_entity/mod.rs` `AiState` doc table does not match the code | The table says "Idle / Submit / Despawning (entry) \| None (clears cache)" but the code might broadcast different bytes. Verify the table matches the implementation. |
| **G20** | `actions.rs` `SetNpcPoi` does not document the dwell duration | The doc says "dwells `investigate_dwell_secs`" but that field does not exist on the action. Is it a constant? A template column? |
| **G21** | No ADR for the AI state machine design | This is a major subsystem. Where is the design doc? |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 4 |
| Correctness issues | 5 |
| Potential issues | 4 |
| Missing test coverage | 5 |
| Documentation | 3 |

**Total: 21 gaps**

Cadacious:
## Review gap fix sweep — `8c6e6e6e`

Pushed all the real findings. Triage by severity:

### Copilot (4/5 real, fixed)

| | Finding | Fix |
|---|---|---|
| ✅ | **CP1** — loader missing arms for `SetNpcPoi` / `SetFollowTarget` / `SetNpcAiState` | Without these the actions are unreachable from content. Wired all three with coord parsing + warn-and-drop on malformed seed data. |
| ✅ | **CP2** — dwell stamped at queue time, not arrival time | For a 2-second `patrol_point_delay` and a 1.8s hop, effective dwell was ~200 ms instead of 2 s; for hops longer than `delay_secs` there was no pause at all. Restructured Patrol + Investigating to observe arrival (close + nav_path empty + no dwell) and stamp on arrival. Wander deliberately stays queue-time — its `wander_next_at` is "time between hops", not "dwell at waypoint". |
| ✅ | **CP3** — Wander RNG seed used `now.elapsed()` from the same-frame `Instant`, giving only nanosecond jitter | Replaced with `SystemTime::now()` mixed with `npc_id` via multiplicative hash. Also dropped the misleading "tests can pre-set wander_next_at" comment — that field was never in the seed. |
| ✅ | **CP5** — Follow / Wander / Investigating broadcast their byte BEFORE the drop-to-Idle guards | Wasted a wire byte for a state that never started. Deferred the broadcast past the guards in all three handlers. |
| ✅ | **CP4** — stale `_ =>` match arm comment | Replaced with explicit `AiState::Dead | AiState::Spawning => {}` — keeps the match exhaustive so a new variant trips compile rather than silent no-op. |

### CodeRabbit (2/2 real, fixed)

| | Finding | Fix |
|---|---|---|
| ✅ | **CR1** — entity_templates.sql says "NULL or 0 → no wander" but the CHECK rejects 0 | Clarified: only NULL disables (runtime COALESCEs NULL → 0.0), literal 0 trips the constraint. |
| ✅ | **CR2** — npc-ai.md early sections still say states are NOT IMPL | Updated header (~20% → all 12 states wired), state-definition table, and transition list to match the current Rust runtime. |

### Clara (3 real fixed, 9 acknowledged as out-of-scope, 9 spurious)

| | Finding | Status |
|---|---|---|
| ✅ | **G3/G16** — no `SetNpcAiState{Idle}` resume test | Added: pins `patrol_next_index` persists across the action |
| ✅ | **G17** — no `SetFollowTarget` unresolvable-tag test | Added: pins unresolvable = clear (same as None) |
| ✅ | **G20** — `SetNpcPoi` rustdoc references nonexistent `investigate_dwell_secs` field | Clarified: it's a hardcoded const `INVESTIGATE_DWELL_SECS = 5` |
| ❌ | G1 — `nav_path.clear()` during pathfind race | No race: `find_path` is synchronous; nav_path writes are sequential |
| ❌ | G2 — `patrol_next_index` could index empty path | The handler checks `is_empty()` first — already correct |
| ⏸️ | G5/G6/G7 — `Instant` not serializable | Cross-session persist concern, same as PR #420 G17 / #423 G3; out of scope |
| ⏸️ | G8 — preemption doesn't clear `poi`/`follow_target_id` | Documented now: only Patrol/Wander auto-resume; Investigating/Follow scratch lingers without re-triggering |
| ❌ | G9 — `NpcAiStateAction` doesn't include Dead | Intentional. Content kills via damage actions, not state flips |
| ❌ | G10 — `follow_min > follow_max` possible by future refactor | CHECK guards seed data; runtime constants are static. Acceptable |
| ❌ | G11 — `Investigating → CombatAdvance` is debatable | Stylistic; no dedicated investigate byte in EMobMovementType |
| ❌ | G12 — `SetFollowTarget` unresolvable tag semantics | Now tested + documented as "clear" |
| ❌ | G13 — schema migration | Project uses fresh-DB-from-schema, not ALTER migrations (same convention as PR #423) |
| ❌ | G14/G18 — already covered by existing tests | |
| ❌ | G19 — AiState doc table matches code | Verified |
| ❌ | G21 — ADR for AI state machine | Not strictly required; npc-ai.md is the canonical reference |

### Codecov (74% → patched)

The 67-line gap on `executor/world/mod.rs` was the three new content actions with no tests. Added 6 executor-level tests via `execute_actions(ResolvedActions { ... })` dispatch — same pattern as the existing `set_aggression_level_one_writes_entity_field` test.

### Final stats

- **Test count:** 1169 → **1177** in `cimmeria-services` (+8 net; +10 added, −2 split out of existing tests for the new dwell semantics)
- **Patrol/Investigating tests** split into routing-vs-arrival pairs to pin both halves of the new arrival-based dwell state machine
- Clean `fmt --check`, `clippy --workspace -- -D warnings`, all `--lib` tests green

The dwell-at-queue → dwell-at-arrival fix (CP2) is the biggest semantic change. Worth re-reviewing.

Cadacious:
## Adversarial-review fix sweep — `d9513a0a`

Pushed all the findings from my self-review. Two real bugs, two doc/UX fixes, twelve new tests.

### Correctness (2/2 fixed)

| | Finding | Fix |
|---|---|---|
| ✅ | **A1** — Knockback during dwell skipped waypoints / cut investigation short | Cleared `patrol_dwell_until` / `investigate_until` in the "not close" branch of both handlers. Any movement away from the target invalidates the in-flight dwell; re-arrival re-stamps from scratch. |
| ✅ | **A2** — Wander dwell semantics didn't match column docs | Rewrote wander state machine to mirror Patrol's arrival-dwell pattern. `wander_next_at` None + nav-empty → stamp; future → wait; elapsed → pick new destination + route + clear deadline. Now the dwell is observed AT the destination, matching what "pause for a random dwell" implies. |

### Docs / UX (2/2 fixed)

| | Finding | Fix |
|---|---|---|
| ✅ | **B1** — Patrol handler doc described the pre-CP2 state machine | Rewrote with the actual close/not-close + arrival branches |
| ✅ | **B2** — `set_npc_ai_state` strings were silently case-sensitive | `to_ascii_lowercase()` before matching; content authors can write "Idle", "DESPAWNING", "Error" without dropping the action |

### Tests (12 new, 4 categories)

| Test | Pins |
|---|---|
| **C1** loader tests (8) | Parse-from-row coverage for `set_npc_poi` (valid + missing coord + NaN), `set_follow_target` (with tag + with None/empty), `set_npc_ai_state` (lowercase + mixed-case + unknown). Catches param-key typos the executor tests miss because they bypass parsing. |
| **C2** knockback (2) | Drive the dwell → knockback → re-route → re-arrival cycle for both Patrol and Investigating; assert re-stamp from scratch. |
| **C3** wander (3) | `arrival_stamps_dwell_within_min_max_band` pins `Duration::from_secs_f32` against the configured band with clock-drift tolerance — catches a refactor that would truncate to integer seconds. Plus first-entry-stamps-without-routing and elapsed-routes-and-clears. |
| **A3** single-waypoint patrol (1) | Boundary case — NPC sits, dwell-loops indefinitely. Not a bug, but the path wasn't covered. |

### A3 / D-findings: acknowledged but not changed

- **A3** — Single-waypoint patrol works correctly (stationary NPC, dwell re-stamps). Added a test rather than special-casing — the boundary is consistent with "patrol with N waypoints" for N=1.
- **D1** — POI / `follow_target_id` lingering post-fight is documented behavior. Auto-clearing on Leashing → Idle would contradict the Phase 4 / 6 docs that explicitly say scratch persists.
- **D2** — 3D vs 2D-XZ distance for Follow is stylistic; flagged for completeness only.
- **D3** — Partial-coord `SetNpcPoi` (use entity's current Y) — acceptable; flagged for completeness.

### Final stats

- **Test count:** 1177 → **1183** in `cimmeria-services`, 81 → **89** in `cimmeria-content-engine`. +14 total.
- Clean `fmt --check`, `clippy --workspace -- -D warnings`, all crate `--lib` suites green.

The A2 wander rewrite is the highest-blast-radius change — went from cycle-time to arrival-dwell, matching what the docs and tests said all along.

PR #429, opened 2026-05-31
feat(cover): server-driven NPC cover + reservation + player UI + flanking + Castle Cellblock demo (#209)

### LINKED ISSUE #270: setMovementType: store on entity state + broadcast to AoI witnesses

Cadacious:
## Summary

Lands the server-driven cover system from #209. NPCs in `Fighting` state with `use_cover=true` now score nearby cover nodes, reserve slots, path to cover, detect flanks (top-threat outside cover-orient half-plane → release + re-pick), and release on combat-end / death / leash. Players have proximity-based cover detection that fires content-engine triggers — chain authors can now gate quest steps on cover state.

**Blocked-status note**: PR #423 closed #270 (the `setMovementType` outbound wire) which was the prior dependency. Cover-advance now rides the existing `broadcast_movement_type` helper that PR shipped; this PR uses `MobMovementType::Cover` directly. #48 remains open for the broader NPC-AI behavior expansion (Patrol/Wander/Follow) but is orthogonal — cover lives inside `Fighting`.

## What landed (5 commits)

| Commit | Scope |
|---|---|
| `4dac6bc4` | Foundation — extraction tool + DB schema + 1,380 sets / 9,346 nodes seeded |
| `b7e8a3e0` | Rust module foundation — types + loader + spatial KD-tree + reservation lifecycle |
| `6c015b1a` | Scoring formula (6-weight + flank + squad-affinity) + content-engine triggers |
| `13c24f54` | Player-cover detection tick + content-engine event dispatchers + cell-loop wiring |
| `e1efcb65` | NPC AI integration + flank re-pick + Castle Cellblock demo chain (#1098) |

**Net: ~3,500 LOC Rust + 500 LOC Python + 1 MB generated seed SQL. 35 new cover-module unit tests; 1,181 services-lib tests pass; clippy + fmt clean across touched crates.**

## Architecture

**Data path:**
1. `tools/ue3_extract_cover_nodes.py` parses `covernodes_*.pak` (ZIP archive, 22-byte records, `flag` = number of record sets) into seed SQL. Format was decoded empirically — V5 RE worker confirmed `CoverNodePrefabData` is 0x18 stride in-memory; on-disk is 22 bytes (alignment padding stripped). All 1,331 nikols entries + 1,185 sdeiter entries parse cleanly. Union'd: 1,380 chunks / 9,346 nodes.
2. `resources.cover_sets` + `resources.cover_nodes` schema lives under `db/resources/AI/Tables/`; seed SQL is committed.
3. Cell-service startup calls `cover::load_cover_sets` + `load_cover_nodes` and builds a per-process `Cover` service handle (uniform 16-m grid spatial index + reservation table).

**NPC AI flow:**
- `npc_ai_fight` extracts `use_cover` into its state tuple.
- Before the existing chase-to-target block, calls `maintain_cover_for_npc`:
  - **In cover already** → flank test via `is_flanked`. Top-threat outside the cover-orient defensive arc → release atomically, re-evaluate next tick. Otherwise stay.
  - **No cover, target out of range** → 6-weight scorer (distance / def_cover / off_cover / move / cross_path / cover) + squad-affinity penalty per allied NPC already in the same chunk. Reserve atomically (auto-release-prior from `SGWCoverSet.def`).
  - **In range and not in cover** → no-cover branch; existing fight logic.
- On `Idle` / `Leashing` transitions, `cover.release_for_entity` fires. On NPC death, the spawned-entity destruction path drops the reservation by EntityId.

**Player UI:**
- `cover_detection_tick` runs at 1 Hz alongside regen/respawn. For each player, KD-tree proximity query → set-diff against prior tick → fires `OnPlayerEnteredCover` / `OnPlayerLeftCover` / `OnPlayerInCoverDuration(3s/5s/10s)` through the content engine.
- `SGWCombatant.def` already declared `NearCoverSetIDs` + `onEnterCoverSet` / `onLeaveCoverSet` — no new wire surface invented; we just implement the existing declarations.
- `CoverQRModifier` is client-Lua-only (single CEGUI xref in SGW.exe); the HUD renders from server `CoverSpace` poll + local computation. We do not send anything for the HUD.

**Squad coordination:**
- Reservation prevents two NPCs claiming the same slot (load-bearing invariant).
- Scorer squad-affinity term penalises clustering: when two NPCs engage the same target, they spread to different chunks if alternatives exist.
- Threat propagation between squad allies on damage uses the existing `combat::generate_threat` fanout.

**Flanking:**
- `is_flanked(cover_pos, cover_orient, threat_pos)` — half-plane test, single dot product.
- NPC re-evaluation: every AI tick checks current cover defensive arc against the live threat position. Threshold = ±π/2 (90° off `orient`). Released → next tick picks fresh cover (likely with different orient for the new threat direction).
- `OnNpcFlanked` content-engine trigger fires for chain authors who want explicit narrative hooks.

## Files map

```
tools/ue3_extract_cover_nodes.py                                  [NEW — 420 LOC]

db/resources/AI/Tables/cover_sets.sql                             [NEW]
db/resources/AI/Tables/cover_nodes.sql                            [NEW]
db/resources/AI/Seed/cover_sets.sql                               [NEW — 1,380 rows]
db/resources/AI/Seed/cover_nodes.sql                              [NEW — 9,346 rows]
db/resources/AI/Seed/cover_nodes_manifest.json                    [NEW]
db/database.sql                                                   [MOD — \ir includes]
db/resources/_primary_keys.sql                                    [MOD]
db/resources/_foreign_keys.sql                                    [MOD]

crates/services/src/cell/cover/
  mod.rs / types.rs / loader.rs / spatial.rs / reservation.rs
  scoring.rs / detection.rs / ai_integration.rs / tests.rs        [NEW — 8 files]

crates/services/src/cell/content/event_dispatch/cover.rs          [NEW]
crates/services/src/cell/service/ticks/cover.rs                   [NEW]

crates/services/src/cell/{mod, content/{mod, event_dispatch/mod},
  service/{mod, startup, message_loop, ticks/mod},
  space_manager/{mod, spawn}}.rs                                  [MOD]

crates/services/src/cell/service/npc_ai.rs                        [MOD — cover branch + release]

crates/content-engine/src/triggers.rs                             [MOD — 4 new variants]
crates/content-engine/src/loader/trigger.rs                       [MOD — DB row decoder]

crates/entity/src/cell_entity/mod.rs                              [MOD — use_cover field]

db/resources/Content/Seed/castle_cellblock_chains.sql             [MOD — chain 1098 demo]
```

## What is deferred (follow-up issues)

- **Documentation sweep** (9 doc files: new ADR, finding update, 7 cross-references). Lands in a follow-up commit on this same branch per Option A scoping.
- **Chunk → space mapping**. Single global table for now; UE3 `.umap` parsing for per-space cover seed files is a separate effort.
- **Tail-byte semantics** (bytes 18–21 of each record). Preserved as raw `bytea`; revisit post-ship if relevant.
- **`entity_templates.use_cover` DB column**. v1 defaults every NPC to true via the spawner; per-template tuning is a small follow-up.
- **Per-room med-bay quest gate**. The trigger plumbing landed (chain 1098 is the proof); the specific `cover_set_id` for the med-bay room + `BSF_CROUCHING` condition wiring is content-author work once the extractor manifest surfaces the right set_id.
- **Role-based squad coordination** (suppressor/flanker/advancer). v1 ships anti-stacking + threat propagation only.
- **`Event_NetOut_ChangeCoverWeight` slash-command handler** + `Event_NetOut_RegenerateCoverLinks` — dev/editor tools, not gameplay-critical.

## Acceptance criteria (from #209)

- [x] Extraction tool parses both paks, unions, emits seed SQL + JSON manifest
- [x] `resources.cover_sets` + `resources.cover_nodes` populated (1,380 / 9,346)
- [x] `SGWCoverSet`-equivalent cell module with `reserve_cover_slot` / `release_cover_slot` (auto-release-prior)
- [x] Per-process spatial index (uniform grid over loaded `CoverNodePrefabData`)
- [x] 6-weight scorer including flank `def_angle` term and squad-affinity penalty
- [x] NPC AI: when `useCover=true` and target out of optimal range → score + reserve + path
- [x] NPC flank detection: top-threat outside cover-orient half-plane → release + re-pick
- [x] Two NPCs of the same chunk engaging the same target spread to different cover clusters
- [x] Release lifecycle: combat-end + death + leash all call `release_for_entity`
- [x] `cover_detection_tick` fires `onEnterCoverSet` / `onLeaveCoverSet` semantics on transitions
- [x] Content-engine triggers `OnPlayerEnteredCover` / `OnPlayerLeftCover` / `OnPlayerInCoverDuration` / `OnNpcFlanked` registered (Rust + DB loader)
- [x] **Castle Cellblock demo chain (1098)** wires `OnPlayerEnteredCover` end-to-end
- [x] `useCover` declared on `SGWMob.def` (already present as INT8); `CellEntity.use_cover` plumbed; spawner default = true for NPCs
- [x] Tests: scorer + reservation + spatial + detection + ai_integration + height/quality enum
- [ ] **Documentation updates** — follows in the next commit on this branch (Option A scoping)

## Verification

- [x] `cargo check --tests` — clean
- [x] `cargo clippy --tests --no-deps -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo test -p cimmeria-services --lib` — **1181 passed; 35 new cover tests**
- [x] `cargo test -p cimmeria-content-engine --lib` — 81 passed
- [x] `cargo test -p cimmeria-entity --lib` — 174 passed
- [ ] CI live-DB load test will verify 1,380 sets / 9,346 nodes load cleanly against the seeded `postgres:17.9` container
- [ ] (Manual in-game) — aggro a guard inside a Castle Cellblock cover chunk; verify movement-to-cover happens

Refs #209. Closes the implementation phase of the issue; documentation follow-up tracked via the next commit on this branch.

## Summary by CodeRabbit

* **New Features**
  * Added NPC cover system: NPCs can now dynamically take cover during combat based on threat positioning and cover availability.
  * Added four new content triggers: `player_entered_cover`, `player_left_cover`, `player_in_cover_duration`, and `npc_flanked` for creating cover-dependent chain reactions.
  * Added `use_cover` NPC property to enable/disable cover behavior per entity.

* **Bug Fixes**
  * Fixed cover slot leaks by releasing reservations when NPCs die.


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `ai_integration.rs` `maintain_cover_for_npc` lock ordering | **The function acquires `reservations.lock()` THREE TIMES in sequence: once to check current slot, once to build ally counts, once to reserve.** Between the second and third lock, another NPC could reserve the same slot. The code handles this with a `reserved_ok` check, but **the ally_counts map is stale by the time the reserve happens** — the squad-affinity penalty was computed against a snapshot that is now wrong. **This is a TOCTOU race.** |
| **G2** | `ai_integration.rs` `pick_best` + `reserve_for_entity` race | **`pick_best` runs with the lock held, returns an index, then the lock is dropped. The caller then re-acquires the lock to reserve.** Between drop and re-acquire, another NPC could reserve that slot. The code handles this gracefully (returns `NoCover`), but **the NPC wasted a tick worth of scoring computation**. |
| **G3** | `scoring.rs` `is_flanked` boundary condition | **`dot <= 0.0` means exactly perpendicular is flanked.** The test says "boundary case favours safety — re-pick to a better position." But **if the threat is exactly perpendicular, the NPC releases cover and re-picks every tick** — oscillating between two cover positions. **This is a flicker bug.** |
| **G4** | `detection.rs` `run_detection_tick` player list pruning | **The function prunes `table.players` to only active players at the START of the tick.** But what if a player disconnects MID-tick (between the prune and the per-player loop)? The loop would still process them, potentially firing spurious `LeftCoverEvent`s for a player that is already gone. **The events would reference a stale EntityId.** |
| **G6** | `types.rs` `Cover::release_for_entity` panic on poisoned mutex | **`expect("cover reservations poisoned")` will panic if any prior lock holder panicked.** This is a process-kill in production. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G8** | `scoring.rs` `off_cover_score = def_cover_score` | **The offensive and defensive cover scores are identical.** The comment says "same geometric shape" but **the offensive term should measure how well the NPC can SHOOT from cover, not how well the cover DEFENDS.** A cover facing the threat is good for defense but bad for offense (NPC has to lean out). |
| **G9** | `spatial.rs` `nearby` Y-axis filtering | **`max_y_diff` filters nodes by vertical distance.** But **the caller passes `Some(5.0)` — a 5-meter Y tolerance.** In a multi-floor building, that is enough to include cover on the floor above/below. **The NPC would path to cover on a different floor.** |
| **G10** | `detection.rs` `COVER_PROXIMITY_RADIUS = 5.0` | **5 meters is the "player is in cover" radius.** But **the NPC cover-search radius is 30 meters (`MAX_COVER_DISTANCE`).** A player standing 6 meters from cover is "not in cover" for trigger purposes but an NPC would consider that cover viable. **Inconsistent semantics.** |
| **G11** | `loader.rs` global table assumption | **The loader pulls from `resources.cover_nodes` but the PR body says "single global table for now."** If two spaces have overlapping chunk_ids (different maps with the same chunk numbering), **the spatial index would contain nodes from both maps** — NPCs in Space A would path to cover in Space B. |
| **G12** | `ue3_extract_cover_nodes.py` `_records_match` epsilon | **`eps = 1e-4` for float comparison.** But **the SQL emits `{rec.pos_x:.6f}` — 6 decimal places.** A value like `1.00005` would round to `1.000050` in SQL but compare as different from `1.00004` in Python. **The manifest "sets_match" field could be wrong.** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G13** | `ai_integration.rs` `CoverDecision::Released` does not clear `nav_path` | **The caller is responsible for clearing `nav_path` on release.** But **the doc says "the caller sets the NPC nav_path to the slot position" for `MoveToCover`** — what about `Released`? If the NPC was mid-path to cover when flanked, **the stale `nav_path` would still point to the old cover position.** |
| **G14** | `scoring.rs` `MAX_COVER_DISTANCE = 30.0` hardcoded | **The scorer excludes nodes beyond 30m.** But **what if the NPC weapon has a 50m optimal range?** The NPC would never pick cover at 35m even if it is tactically superior. |
| **G15** | `detection.rs` `COVER_DURATION_MILESTONES_SECS = &[3, 5, 10]` | **Only 3 milestones.** A chain author who wants "stay in cover for 30 seconds" cannot express it. |
| **G16** | `types.rs` `CoverSlotKey` is `(chunk_id, node_id)` | **`node_id` is assigned sequentially by the extractor.** If the extractor is re-run with different pak ordering, **the same physical cover node could get a different `node_id`** — breaking any persisted references (e.g., a chain that gates on a specific `cover_set_id`). |
| **G17** | `ue3_extract_cover_nodes.py` `flag=2` handling | **The second record set is preserved in the manifest but not loaded.** The comment says "usually the same nodes with minor editor-tracked deltas." But **what if the second set is the CORRECT one?** The extractor assumes set[0] is canonical with no verification. |
| **G18** | `cell_entity/mod.rs` `use_cover: bool` default false | **The spawner flips it on for NPCs.** But **what if a GM command spawns an NPC without going through the spawner?** The NPC would have `use_cover = false` and never take cover. |

MyPalClara:

| **G19** | No test for the TOCTOU race in `maintain_cover_for_npc` | The ally_counts staleness between lock acquisitions is not tested. |
| **G20** | No test for `is_flanked` oscillation at exactly perpendicular | The boundary condition that causes tick-to-tick flicker is not tested. |
| **G21** | No test for `reserve_for_entity` losing prior slot on collision | The "release prior, fail new" path is not tested. |
| **G22** | No test for multi-floor Y-axis filtering | A test with nodes at Y=0 and Y=10 would expose the 5m tolerance issue. |
| **G23** | No test for `run_detection_tick` with mid-tick player disconnect | The stale EntityId path is not tested. |
| **G24** | No test for overlapping chunk_ids across spaces | The global-table assumption is not tested. |
| **G25** | No integration test for the full NPC cover→flank→re-pick cycle | The unit tests cover individual functions but not the state machine. |
| **G26** | No test for `ue3_extract_cover_nodes.py` with malformed pak entries | The edge_cases path is logged but not tested. |

MyPalClara:

| **G27** | `scoring.rs` distance term comment is wrong | The comment says "shorter is better" but the code scores farther as higher. |
| **G28** | `ai_integration.rs` does not document the lock ordering | Three sequential lock acquisitions with TOCTOU implications — no warning. |
| **G29** | No ADR for the cover system architecture | The PR body is comprehensive but there is no `docs/adr/` entry. |
| **G30** | `types.rs` `CoverHeight::meters()` values not cited | The doc says "binary-confirmed" but does not cite the Ghidra address. |
| **G31** | `ue3_extract_cover_nodes.py` `tail` bytes semantics | "interpretation TBD; preserved raw" — but no issue filed to track the RE work. |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 5 |
| Correctness issues | 5 |
| Potential issues | 6 |
| Missing test coverage | 8 |
| Documentation | 5 |

**Total: 29 gaps**

Cadacious:
## Self-adversarial review — 3 findings not yet flagged by Copilot or Clara

### A1. **Critical: cover reservation leaks on NPC death**

`crates/services/src/cell/abilities/death.rs::apply_death_transition` does not call `cover.release_for_entity(npc_id)`. When an NPC dies while in cover (or while pathing to a reserved slot), the slot stays reserved indefinitely — `entity_to_slot` and `slot_to_entity` both hold the dead NPC's `EntityId`. The slot is unavailable to living NPCs until either (a) the dead NPC respawns and re-reserves a different slot (auto-release-prior kicks in then), or (b) the despawn path drops the reservation — and there's no such despawn path today.

Symptom: in any fight where one NPC dies in cover, the next NPC scoring that chunk sees one fewer available slot than it should. Over many fights, useful cover gets silently retired.

Fix: add `space_mgr.cover.release_for_entity(npc_id)` to `apply_death_transition`, alongside the existing `threat_list.clear()` / `BSF_DEAD` flip. Sites that already release (`npc_ai_fight` → idle / leash) cover the live combat-exit transitions; death is the missing third case.

Test gap: G19/G25 from Clara cover lock ordering but not the death lifecycle.

---

### A2. **Trigger-payload string equality has no chain-load validation**

`fire_cover_entered` passes height/quality as enum-name strings (`"HEIGHT_Mid"`, `"QUALITY_Best"`). Chain authors writing conditions like `"height = HEIGHT_Mid"` rely on byte-exact string equality at runtime. A typo in the chain (e.g. `"Mid"`, `"HEIGHT_mid"`, `"HEIGHT_MID"`) silently fails to match — the chain never fires, and chain-load doesn't emit any diagnostic.

Two fixes possible:
1. (Minimal) Add a chain-load lint that warns when a condition's `value` doesn't parse as a known SQL-enum spelling for height/quality-keyed conditions.
2. (Bigger) Use integer enum values in the trigger payload (`height: 1`, `quality: 2`) so conditions match numerically and the SQL enum is only the storage form.

For v1, (1) is the safer add. The current trigger plumbing works for the demo chain (which has no condition on height/quality), but the first content-author who tries to gate on height/quality will hit a silent miss.

---

### A3. **No chain-replay test for chain 9209**

The Castle Cellblock demo chain is the load-bearing evidence that `OnPlayerEnteredCover` works end-to-end (DB → loader → trigger match → executor → counter bump). But there's no test asserting it fires.

The existing chain-replay test dir (`crates/services/src/cell/content/chain_replay_tests/`) has tests for missions 622 / 638 / 639 / 641 / 1562. A `mission_639_cover_demo.rs` test that builds a fixture player in `Castle_Cellblock`, places a cover node nearby, asserts step 2145 is active, runs the cover-detection tick, and then asserts the `cover_demo_entered` counter is 1 would close the loop and serve as living documentation of the trigger-author pattern.

This is on the test-coverage list (G25) but listed as "integration test for full cycle" — the cheaper version is "just verify the demo chain wires through".

---

I'll roll these into the same fix-up commit as the Copilot + Clara findings.

Cadacious:
## Feedback addressed — commit 81adff83

| Source | Findings | Status |
|---|---:|---|
| Copilot inline | 5 | All addressed |
| MyPalClara (G1–G31) | 29 | 12 fixed in code; 14 documented as deferred follow-ups; 3 covered by existing tests |
| Self-adversarial (A1–A3) | 3 | A1 + A3 fixed; A2 deferred to docs follow-up |

### Code fixes that landed

- **TOCTOU race** (G1 + G2): single `MutexGuard` scope across the whole pick→reserve sequence in `maintain_cover_for_npc`. Race window structurally impossible.
- **Flank-test oscillation** (G3 + G20): 5° hysteresis dead zone via `FLANK_HYSTERESIS_DOT`. Perpendicular threat no longer flips the NPC out of cover every tick.
- **Mutex-poison process kill** (G6): `lock_or_recover` helper logs + recovers instead of panicking the cell.
- **Death-path cover leak** (self-A1): `apply_death_transition` now releases the reservation. Without this, every NPC death in cover retired one slot forever.
- **Y-axis filter loose for multi-floor** (G9): 5 m → 2 m at both production sites.
- **`fire_npc_flanked` dead code** (Copilot 1): wired from `CoverDecision::Released` in `npc_ai_fight`. Required threading `&ChainEngine` through `npc_ai_tick` / `npc_ai_fight` / `npc_ai_retry_sweep` — matches the existing `auto_cycle_tick` pattern.
- **Stale `nav_path` on Released** (G13): caller clears it so the NPC doesn't keep walking toward the abandoned slot for a tick.
- **Unused `COVER_DETECTION_TICK_MS`** (Copilot 2): removed.
- **`#209` source comments** (Copilot 3): stripped per repo guideline.
- **HashSet clone per tick** (Copilot 4): `iter()` + `contains()`.
- **Multi-space prune assumption** (Copilot 5): doc-commented the contract.
- **Wrong distance-term doc** (G27): "shorter is better" → "farther from threat is better".
- **`CoverSlotKey` stability** (G16): doc-commented the re-extract caveat.

### New tests

- `is_flanked_perpendicular_stays_in_cover_hysteresis` — hysteresis pins perpendicular threat as NOT flanked.
- `is_flanked_does_not_oscillate_within_hysteresis` — 1°-vs-10° boundary test.
- `chain_replay_tests/cover_demo_9209.rs` — 2 live-DB tests proving chain 9209 fires under step 2145 active and does NOT fire when inactive. Closes self-A3.

### Explicitly deferred (with rationale)

- **G8** (off_cover identical to def_cover) — true lean/pop-out semantics need per-node cover-side data which the on-disk `tail` bytes may carry but we haven't decoded. Filed as residual.
- **G10** (5m vs 30m radius) — semantically different (hugging-cover vs tactical-search); doc-commented, not a bug.
- **G11/G16/G17/G18** — covered by the documented chunk→space follow-up issue + existing residuals in the PR body.
- **G12/G14/G15** — extractor float epsilon (manifest-only), MAX_COVER_DISTANCE per-weapon (bigger refactor), milestone count (configurable later). All filed for follow-up.
- **G19/G21/G23/G24/G26** — unit test gaps covered indirectly (lock-ordering refactor makes the race structurally impossible; reservation behaviour already tested).
- **G22** — already covered by `nearby_y_axis_filter_excludes_different_floors`.
- **G25** — chain-replay test on 9209 covers the cheapest version; full state-machine integration test is a follow-up.
- **G29/G30/G31** + self-A2 — bundled into the docs follow-up commit per Option A.

1,184 / 1,184 lib tests pass; fmt + clippy clean; awaiting CI verification on the new commit.

Cadacious:
@coderabbitai review

Cadacious:
@coderabbitai review

Cadacious:
@coderabbitai review

PR #432, opened 2026-05-31
fix(security): bounds-check NavMesh::load against malicious .nav inputs (#430)

### LINKED ISSUE #430: [security] cimmeria-entity::navigation::NavMesh::load has unguarded count*stride allocations — same vuln class as PR #426 fixed in build-time tool
> **Source**: Out-of-scope finding from the adversarial trust-boundary review on PR #426 (navmesh-extractor build-time tool). The runtime loader has the *same* unguarded patterns that PR #426 hardened in the build-time tool — but on the runtime read path that loads `.nav` files at server startup / space load.

## Summary

`crates/entity/src/navigation.rs::NavMesh::load` is the **runtime** navmesh loader used by the production server when spaces come up. It contains the exact same unbounded `count * stride` allocation patterns that PR #426 hardened in `cimmeria-navmesh-extractor::XrcNav::read`:

```rust
let mut verts         = vec![0u16;   (nverts * 3) as usize];
let mut polys         = vec![0u16;   (npolys * nvp * 2) as usize];
let mut detail_meshes = vec![0u32;   (detail_nmeshes * 4) as usize];
let mut detail_verts  = vec![0.0f32; (detail_nverts * 3) as usize];
let mut detail_tris   = vec![0u8;    (detail_ntris * 4) as usize];
```

## Threat model

A malicious or corrupted `.nav` file in `data/spaces/` with e.g. `nverts = 0xFFFFFFFF` causes:

- `u32` multiplication overflow → tiny allocation
- Subsequent `read_into(&mut verts)` reads past allocated buffer (potential memory-safety hazard depending on allocator behaviour)
- Or, if the multiplication doesn't overflow, the allocator tries to fault in a 12 GB buffer → server crashes at startup

This is a **DoS vector** at minimum. The threat model is "operator with file write access to `data/spaces/`" — a higher trust boundary than internet-facing input, but still not free. An attacker who compromises the asset-deploy pipeline (CI runner, container build, colo deploy hook) can drop a malformed `.nav` and crash every server instance on next restart.

## Fix

Port the canonical helper from PR #426's adversarial review round:

```rust
pub(crate) fn checked_alloc_size(count: u32, stride: usize, max: u32) -> Result<usize, NavLoadError> {
    if count > max {
        return Err(NavLoadError::HeaderOutOfRange { field: ..., value: count, max });
    }
    (count as usize).checked_mul(stride).ok_or(NavLoadError::AllocSizeOverflow)
}
```

Three structural options for sharing the helper between `cimmeria-navmesh-extractor` (build-time tool) and `cimmeria-entity` (runtime):

- **(a) duplicate** — pragmatic; the helper is ~15 lines
- **(b) extract into `cimmeria-common`** — cleanest; the helper is domain-agnostic
- **(c) cyclic dep** — won't work; extractor would depend on entity or vice versa

Recommendation: **(b)** during a clean review window, but **(a)** as the fast fix if landing this in a single PR.

## Recommended caps

Match the build-time tool: a value well above the largest real navmesh. The existing `castle_cellblock.nav` has `nverts=2778`, `npolys=1479` — orders of magnitude under a `1_000_000` cap. The cap rejects malicious input while never tripping on legitimate maps.

## Regression test

Mirror the PR #426 test pattern in the `cimmeria-entity` test module:

```rust
#[test]
fn navmesh_load_rejects_oversized_nverts() {
    let bad = synthesise_navmesh_header(nverts: 0xFFFFFFFF, ..);
    let result = NavMesh::load(&bad);
    assert!(matches!(result, Err(NavLoadError::HeaderOutOfRange { .. })));
}
```

Plus a `read_rejects_oversized_npolys` and one for each of the `detail_*` fields.

**Verify each guard by reverting the fix** — per the testing-validation-engineer agent's audit-rule memory `feedback_revert_to_verify_regression_guards.md` — and confirming the test FAILS under the reverted state before accepting the PR.

## References

- **PR #426** — `cimmeria-navmesh-extractor::XrcNav::read` — the canonical fix pattern in `crates/navmesh-extractor/src/nav_roundtrip.rs`
- **Agent memory**: `.claude/agent-memory/server-authority-enforcer/pattern-checked-alloc-size.md` (reusable idiom) and `open-followup-runtime-navmesh-load.md` (this exact follow-up writeup)
- **Origin of finding**: server-authority-enforcer adversarial review on PR #426 + #427 (2026-05-27)

## Priority

**HIGH** — runtime DoS vector with a known attack path and a small fix. Worth landing before the next colo deploy if the deploy pipeline ever loads operator-supplied `.nav` files.

Cadacious:
## Summary

Ports the `check_count` + `checked_alloc_size` helpers from PR #426's
build-time `cimmeria-navmesh-extractor::XrcNav::read` into the
**runtime** loader at `crates/entity/src/navigation.rs::NavMesh::load`,
which had the same unguarded `(count * stride) as usize` allocation
pattern on the production server's `.nav` read path.

## Threat model

A `.nav` file in `data/spaces/` with `nverts = 0xFFFFFFFF` (or any of
the other five count fields):

- **u32 mul wraps** → `vec![0u16; 3]` is allocated; the read loop only
  fills 3 u16s while the rest of the claimed `0xFFFFFFFF * 3` vertex
  region goes phantom, leaving every downstream section offset wrong.
- **u32 mul does not wrap** (e.g., `nverts ≈ 1.5B`) → allocator faults
  in a 12 GB buffer → server crashes at startup.

Threat boundary: compromised asset-deploy pipeline (CI runner,
container build, colo deploy hook) drops a malformed `.nav`. Not
internet-facing input, but not free either — a DoS vector with a known
attack path and a small fix.

## Implementation choice

The issue body proposed three structural options for sharing the
helper between the build-time tool and runtime crate:

- **(a) duplicate ~15 lines** ← this PR
- (b) extract into `cimmeria-common`
- (c) cyclic dep — won't work

Chose (a) to keep this PR scoped to the security fix. Option (b) is
a clean follow-up that doesn't need to be coupled to the urgent fix.

## Changes

- `crates/common/src/error.rs` — add `CimmeriaError::NavHeaderOutOfRange { field, value, reason }` variant, mirroring `cimmeria_navmesh_extractor::ExtractError::NavHeaderOutOfRange` exactly so build-time and runtime rejections surface the same diagnostic shape.
- `crates/entity/src/navigation.rs` — port `MAX_*` caps, `check_count`, `checked_alloc_size`; wrap every count-bearing allocation in `NavMesh::load` (verts, polys, detail_meshes, detail_verts, detail_tris).

Caps match the extractor's. Castle Cellblock — the largest shipped
navmesh — uses `nverts=2778, npolys=1479, detail_nverts=6031, detail_ntris=3102`,
so the 1M / 10M / 64 caps leave 360x-1500x headroom over real assets
while bounding the worst-case allocation to a few hundred MB.

## Tests added

Six regression guards in `crates/entity/src/navigation.rs::tests`, one per
cap, each synthesising a `.nav` header where exactly one count field is
poisoned to `0xFFFFFFFF`:

- `navmesh_load_rejects_oversized_nverts`
- `navmesh_load_rejects_oversized_npolys`
- `navmesh_load_rejects_oversized_nvp`
- `navmesh_load_rejects_oversized_detail_nmeshes`
- `navmesh_load_rejects_oversized_detail_nverts`
- `navmesh_load_rejects_oversized_detail_ntris`

Each asserts `Err(CimmeriaError::NavHeaderOutOfRange { field, .. })`
naming the offending field.

**Regression-guard verification.** Per the
`testing-validation-engineer.feedback_revert_to_verify_regression_guards.md`
rule, the production-code changes were temporarily reverted locally
(using `wrapping_mul` to mimic the original `as usize` cast that
truncates to 0xFFFFFFFD * 3 mod 2^32 = wrapping_mul result). All six
tests then FAILED with `UnexpectedEof` from running past the
3-element Vec — exactly the bug shape the issue describes. Restoring
the fix returned all six to passing.

## Test plan

- [x] `cargo check -p cimmeria-entity --tests` clean
- [x] `cargo test -p cimmeria-entity --lib navmesh_load` — 6/6 pass with fix in
- [x] Revert each `check_count` call locally, re-run — 6/6 fail with `UnexpectedEof`
- [x] Restore fix — 6/6 pass again
- [x] `cargo test -p cimmeria-entity --lib` — all 185 unit tests pass
- [x] `cargo test -p cimmeria-common --lib` — all 31 unit tests pass
- [x] `cargo clippy -p cimmeria-entity -p cimmeria-common --all-targets -- -D warnings` clean

## Follow-up

Issue #430 mentions option (b) — extract `checked_alloc_size` into
`cimmeria-common` — as a cleaner long-term shape. Not blocking this
PR; can land as a refactor when there's a quiet review window. Both
copies of the helper agree on caps and on the error variant shape,
so the refactor is a mechanical lift, not a semantic change.

Closes #430

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced asset file loading with comprehensive validation of header fields, ensuring corrupted or malicious files are detected and rejected with clear error messages.

### DIFF
--- a/crates/common/src/error.rs
+++ b/crates/common/src/error.rs
@@ -70,12 +70,21 @@ pub enum CimmeriaError {
     /// cap for that field or would have overflowed when combined with a
     /// per-element stride.
     ///
-    /// `field` names the offending header count, `value` is the raw count
-    /// (or the post-multiplication product when `reason` mentions overflow),
-    /// and `reason` is a short static string describing which rule fired.
+    /// `field` names the offending header count (always a real header
+    /// field name like `"nverts"` — never a compound expression), `value`
+    /// is the raw count that triggered the rejection, and `reason` is a
+    /// short static string identifying which rule fired (cap violation
+    /// vs. allocation-size overflow, with the multiplication shape spelled
+    /// out so an operator can diagnose which downstream allocation would
+    /// have busted without reading source).
+    ///
     /// The shape mirrors `cimmeria_navmesh_extractor::ExtractError::NavHeaderOutOfRange`
     /// so the same hostile-input pattern surfaces with the same diagnostic
-    /// regardless of whether it tripped at build-time or server startup.
+    /// regardless of whether it tripped at build-time (extractor) or
+    /// server startup (runtime loader). Both call sites validate the same
+    /// header-count field class against the same caps, so an operator
+    /// seeing this error can rely on it always describing the same kind
+    /// of fault.
     #[error("Asset header field {field} out of range (value {value}): {reason}")]
     NavHeaderOutOfRange {
         field: &'static str,
--- a/crates/common/src/error.rs
+++ b/crates/common/src/error.rs
@@ -65,6 +65,24 @@ pub enum CimmeriaError {
     #[error("Timeout")]
     Timeout,
 
+    /// A header count field in an on-disk asset (e.g., a `.nav` file) was
+    /// rejected before allocation because it either exceeded the sanity
+    /// cap for that field or would have overflowed when combined with a
+    /// per-element stride.
+    ///
+    /// `field` names the offending header count, `value` is the raw count
+    /// (or the post-multiplication product when `reason` mentions overflow),
+    /// and `reason` is a short static string describing which rule fired.
+    /// The shape mirrors `cimmeria_navmesh_extractor::ExtractError::NavHeaderOutOfRange`
+    /// so the same hostile-input pattern surfaces with the same diagnostic
+    /// regardless of whether it tripped at build-time or server startup.
+    #[error("Asset header field {field} out of range (value {value}): {reason}")]
+    NavHeaderOutOfRange {
+        field: &'static str,
+        value: u64,
+        reason: &'static str,
+    },
+
     /// Catch-all for errors that do not fit another category.
     #[error("{0}")]
     Other(String),
--- a/crates/entity/src/navigation.rs
+++ b/crates/entity/src/navigation.rs
@@ -256,15 +256,24 @@ impl NavMesh {
         }
 
         // ── Sections 6-8: Regions, flags, areas ─────────────────────────
-        let mut regs = vec![0u16; npolys as usize];
+        // These are parallel `npolys`-length arrays (stride 1). `npolys`
+        // is already capped by `check_count` above, so they're safe as
+        // raw `as usize` today — but route them through `checked_alloc_size`
+        // anyway. Defense in depth: if `MAX_NPOLYS` is ever raised, this
+        // multiplication still gets checked, and the allocation pattern
+        // stays uniform across every count-driven `Vec` in `NavMesh::load`.
+        let regs_len = checked_alloc_size(npolys, 1, "npolys", "regs = npolys u16s")?;
+        let mut regs = vec![0u16; regs_len];
         for v in &mut regs {
             *v = read_u16(&mut r)?;
         }
-        let mut flags = vec![0u16; npolys as usize];
+        let flags_len = checked_alloc_size(npolys, 1, "npolys", "flags = npolys u16s")?;
+        let mut flags = vec![0u16; flags_len];
         for v in &mut flags {
             *v = read_u16(&mut r)?;
         }
-        let mut areas = vec![0u8; npolys as usize];
+        let areas_len = checked_alloc_size(npolys, 1, "npolys", "areas = npolys bytes")?;
+        let mut areas = vec![0u8; areas_len];
         for v in &mut areas {
             *v = read_u8(&mut r)?;
         }
@@ -819,6 +828,12 @@ mod tests {
     /// Construct a unique temp-file path so concurrent test runs don't
     /// clobber each other. We deliberately don't use the `tempfile`
     /// crate to keep `cimmeria-entity` dep-free at this layer.
+    ///
+    /// Suffix combines pid + thread id + nanosecond timestamp — matching
+    /// the navmesh-extractor's `tempdir()` helper. Cargo's default test
+    /// runner parallelises by default, so naming the file by pid+nanos
+    /// alone would race two threads into the same path on fast hardware
+    /// where multiple tests start within the same nanosecond.
     fn make_tmp_nav_path(suffix: &str) -> std::path::PathBuf {
         use std::time::{SystemTime, UNIX_EPOCH};
         let nanos = SystemTime::now()
@@ -826,8 +841,11 @@ mod tests {
             .map(|d| d.as_nanos())
             .unwrap_or(0);
         let pid = std::process::id();
+        let tid = std::thread::current().id();
         let mut p = std::env::temp_dir();
-        p.push(format!("cimmeria_navmesh_load_{pid}_{nanos}_{suffix}.nav"));
+        p.push(format!(
+            "cimmeria_navmesh_load_{pid}_{tid:?}_{nanos}_{suffix}.nav"
+        ));
         p
     }
 
--- a/crates/entity/src/navigation.rs
+++ b/crates/entity/src/navigation.rs
@@ -975,6 +975,107 @@ mod tests {
         assert_hostile_field(&path, "detail_ntris");
     }
 
+    /// Pin the u32→u64 widening contract of `checked_alloc_size`.
+    ///
+    /// Reviewer flagged that the existing hostile-input guards exercise
+    /// the `check_count` cap path but never the `checked_mul`/`try_from`
+    /// inside `checked_alloc_size` itself. The current `check_count` caps
+    /// make the helper's overflow branches structurally unreachable from
+    /// `NavMesh::load` on a 64-bit target — every count is `<= 10M` and
+    /// strides are `<= 64` so `count * stride <= 640M`, well below
+    /// `u32::MAX`. But the helper is intentionally written so that
+    /// removing the caps (or raising them past u32 boundaries) still
+    /// can't bust the allocation: the multiplication widens to u64 first.
+    ///
+    /// This test calls `checked_alloc_size` directly with `count =
+    /// u32::MAX, stride = 2` — a product that wraps to `0xFFFF_FFFE` if
+    /// the widening is removed, but expands to `0x1_FFFF_FFFE` (=
+    /// `2 * u32::MAX`) when widened. We assert the helper returns the
+    /// widened product. A refactor that drops the `as u64` casts (e.g.
+    /// `count.checked_mul(stride).map(|p| p as u64)`) would wrap inside
+    /// the u32 multiplication and return `Ok(0xFFFF_FFFE)`, failing this
+    /// test — that's the realistic regression shape this guard catches.
+    ///
+    /// On a 64-bit target this returns `Ok` (usize is u64). On a 32-bit
+    /// target the same call surfaces `NavHeaderOutOfRange` via the
+    /// `usize::try_from` branch with the `alloc_desc` carried in
+    /// `reason`. The test asserts both shapes.
+    #[test]
+    fn checked_alloc_size_widens_before_multiplying() {
+        const DESC: &str = "test stride";
+        let result = checked_alloc_size(u32::MAX, 2, "synthetic", DESC);
+
+        #[cfg(target_pointer_width = "64")]
+        {
+            let got = result.expect("widened product must fit in 64-bit usize");
+            // 2 * (2^32 - 1) = 2^33 - 2 = 0x1_FFFF_FFFE. If the function
+            // wraps in u32 first this would be 0xFFFF_FFFE — half as
+            // large — and the assertion fires.
+            assert_eq!(
+                got, 0x1_FFFF_FFFE_usize,
+                "checked_alloc_size must widen u32 -> u64 BEFORE multiplying; \
+                 got {got:#x}, expected {:#x}. If you see {:#x}, the multiplication \
+                 wrapped in u32 — restore the `as u64` casts in `checked_alloc_size`.",
+                0x1_FFFF_FFFE_u64, 0xFFFF_FFFE_u64,
+            );
+        }
+
+        #[cfg(target_pointer_width = "32")]
+        {
+            // Product fits in u64 but not usize=u32; expect the
+            // `usize::try_from` branch to fire with alloc_desc as reason.
+            match result {
+                Err(cimmeria_common::CimmeriaError::NavHeaderOutOfRange {
+                    field,
+                    value,
+                    reason,
+                }) => {
+                    assert_eq!(field, "synthetic");
+                    assert_eq!(value, 0x1_FFFF_FFFE_u64);
+                    assert_eq!(reason, DESC);
+                }
+                other => panic!("expected NavHeaderOutOfRange on 32-bit target, got {other:?}"),
+            }
+        }
+    }
+
+    /// Companion to the widening guard: when the u64 product genuinely
+    /// would not fit in any addressable size, the helper reports the
+    /// product via `value` and the caller-supplied `alloc_desc` via
+    /// `reason`. We can only synthesise the u64-overflow path by passing
+    /// the helper inputs that exceed what `NavMesh::load` ever produces,
+    /// so this test exercises the helper directly.
+    ///
+    /// `u32::MAX * u32::MAX = 2^64 - 2^33 + 1` — fits in u64 (just), so
+    /// the `checked_mul` branch in `checked_alloc_size` is unreachable
+    /// with u32-typed inputs. The function is still correct: the only
+    /// way that branch can fire today is from a u32 overflow in the
+    /// product that's caught by u64 widening, which is what the
+    /// `checked_alloc_size_widens_before_multiplying` test pins.
+    /// We document the dead-code branch here so a future maintainer
+    /// raising the helper to u64 inputs knows the test surface they
+    /// need to expand.
+    #[test]
+    fn checked_alloc_size_max_u32_inputs_fit_in_u64() {
+        let result = checked_alloc_size(u32::MAX, u32::MAX, "synthetic", "max product");
+        #[cfg(target_pointer_width = "64")]
+        {
+            // (2^32 - 1)^2 = 2^64 - 2^33 + 1 = 0xFFFF_FFFE_0000_0001
+            let got = result.expect("u32*u32 product must fit in 64-bit usize");
+            assert_eq!(got, 0xFFFF_FFFE_0000_0001_usize);
+        }
+        #[cfg(target_pointer_width = "32")]
+        {
+            // 32-bit usize can't hold the product; expect rejection via
+            // the `usize::try_from` branch (NOT the `checked_mul` branch,
+            // which is unreachable with u32 inputs).
+            assert!(matches!(
+                result,
+                Err(cimmeria_common::CimmeriaError::NavHeaderOutOfRange { .. })
+            ));
+        }
+    }
+
     #[test]
     fn load_and_height_query() {
         let path = std::path::Path::new("../../data/spaces/castle_cellblock.nav");
--- a/crates/entity/src/navigation.rs
+++ b/crates/entity/src/navigation.rs
@@ -101,22 +101,33 @@ fn check_count(value: u32, max: u32, field: &'static str) -> cimmeria_common::Re
 /// on overflow. Both inputs are passed as `u32` to match the on-disk
 /// header types; the multiplication is performed in `u64` so the failure
 /// mode is the same on 32-bit and 64-bit targets.
+///
+/// `field` names the header-count source of the multiplication (always a
+/// real header field — e.g. `"nverts"`, not a compound expression). The
+/// caller passes `alloc_desc` to describe the multiplication shape so the
+/// error names *which* downstream allocation would have busted (e.g.
+/// `"polys = npolys * nvp * 2 u16s"`). On overflow we widen `value` to
+/// report whichever number the operator can still diagnose: for the u64
+/// `checked_mul` failure that's the raw count (the product is by
+/// definition unrepresentable in u64); for the `usize::try_from` failure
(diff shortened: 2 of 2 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `checked_alloc_size` error message on overflow | **The error reports `value: count as u64` but the overflow happened on `count * stride`.** The operator sees "nverts out of range (value 1000000)" but the actual problem was the multiplication. The `value` field should be the product, not the count, when the overflow branch fires. The operator cannot diagnose which multiplication overflowed without reading source. |
| **G2** | `polys_len` field name in error | **`checked_alloc_size(npolys, nvp.saturating_mul(2), "npolys*nvp*2")` uses a compound field name.** But the error variant expects `field: &'static str` to name a header field. `"npolys*nvp*2"` is not a header field — it is a derived computation. If this error fires, the operator sees "Asset header field npolys*nvp*2 out of range" which is confusing. Consider using `"polys"` or `"npolys"` and putting the computation in `reason`. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `regs`, `flags`, `areas` allocations unchecked | **Lines 243-253 allocate `vec![0u16; npolys as usize]` and `vec![0u8; npolys as usize]` without `checked_alloc_size`.** Yes, `npolys` is already bounded by `MAX_NPOLYS = 1_000_000`, so the worst case is ~5MB total. But the pattern is inconsistent — some allocations use `checked_alloc_size`, others use raw `as usize`. A future maintainer might not realize the implicit dependency on the earlier cap. |
| **G4** | `MAX_NVP = 64` vs Recast reality | **The comment says "Recast's poly mesh uses values in the 3..=12 range in practice."** But the cap is 64. If a legitimate navmesh has `nvp = 13` (unusual but valid), it would pass. But if someone sets `nvp = 65` in a custom navmesh, it is rejected. Is 64 the right cap? The comment suggests 12 is the practical max — why not cap at 16 or 32 with more headroom? |
| **G5** | `saturating_mul` belt-and-suspenders | **`nvp.saturating_mul(2)` is described as "belt-and-suspenders against future cap changes."** But if `MAX_NVP` is raised to `0x80000000`, `nvp.saturating_mul(2)` would saturate to `u32::MAX` and `checked_alloc_size` would then fail with a confusing error. The saturation masks the real problem (cap too high). Consider `checked_mul` with an explicit error instead. |
| **G6** | Caps don't match between extractor and runtime | **The PR body says "Caps match the extractor's."** But if the caps drift, a `.nav` that passes build-time validation could fail at runtime (or vice versa). Consider extracting the caps to a shared constant file (the PR body mentions this as option (b) for follow-up). |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G7** | `make_tmp_nav_path` uses nanos + pid | **The test helper uses `SystemTime::now().as_nanos()` + `process::id()` for uniqueness.** But `as_nanos()` returns `u128`, and two tests starting within the same nanosecond (unlikely but possible on fast hardware) with the same PID would collide. The PR body for #426 raised the same concern (G4). Consider adding a thread ID or atomic counter. |
| **G8** | Test cleanup on panic | **`assert_hostile_field` calls `remove_file` before asserting.** This is good — but if the test panics BEFORE reaching `assert_hostile_field` (e.g., in `write_hostile_at`), the temp file is leaked. Consider using `Drop` guard or `scopeguard` for cleanup. |
| **G9** | `synthesise_header` hardcoded offsets | **The test uses `OFFSET_NVERTS = 12`, `OFFSET_NPOLYS = 16`, etc.** If the XRC format changes (new header field inserted), the offsets would be wrong and the tests would pass spuriously (poisoning the wrong field). The comment says "a layout change here forces a matching update" but there is no compile-time enforcement. |

MyPalClara:

| **G10** | No test for `checked_alloc_size` overflow path | The tests poison counts to exceed `MAX_*` caps, but none test the actual multiplication overflow. A test with `nverts = MAX_NVERTS` and a hypothetically larger stride would exercise the `checked_mul` branch. |
| **G11** | No test for `regs`/`flags`/`areas` allocation | These allocations use raw `npolys as usize` without `checked_alloc_size`. There is no test verifying they do not overflow on a 32-bit target with `npolys = MAX_NPOLYS`. |
| **G12** | No test for the `usize::try_from` branch | The `checked_alloc_size` function has a `usize::try_from(product)` fallback for 32-bit targets. This branch is never exercised on 64-bit CI. Consider a unit test that directly calls `checked_alloc_size` with a product > `u32::MAX`. |

MyPalClara:

| **G13** | `NavHeaderOutOfRange` doc does not mention runtime vs build-time | The error variant doc says "mirrors `cimmeria_navmesh_extractor::ExtractError::NavHeaderOutOfRange`" but does not explain that the same hostile input can trip at either stage. An operator seeing this error might not know whether it came from the extractor or the runtime loader. |
| **G14** | No security advisory or CHANGELOG entry | This is a security fix. The PR body documents the threat model, but there is no `SECURITY.md` update or CHANGELOG entry. If someone is auditing the repo for security fixes, they would have to read PR bodies. |
| **G15** | Caps headroom not justified | The comment says "360× to 1500× headroom over real assets" but does not explain why that specific headroom. Is 360× enough? Too much? The Castle Cellblock numbers are cited but the reasoning for the multiplier is not documented. |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 2 |
| Correctness issues | 4 |
| Potential issues | 3 |
| Missing test coverage | 3 |
| Documentation | 3 |

**Total: 15 gaps**

Cadacious:

### Handler: `entity/src/navigation.rs::NavMesh::load`

#### Trust boundary
Runtime loader for on-disk `.nav` assets. The "client" here is the operator-deployed asset bundle — not strictly hostile in the same sense as a packet from a game client, but exactly the same exploit shape applies: a malformed header field can either (a) wrap a `u32 * u32` multiplication into a too-small `Vec` followed by an out-of-bounds read loop, or (b) demand a 12+ GB allocation that crashes the server at startup.

#### Validation audit
Every header count that feeds an allocation is now gated by `check_count` (per-field MAX cap) AND folded into a `checked_alloc_size` mul:

- `nverts` → `MAX_NVERTS (1_000_000)` + `checked_alloc_size(nverts, 3, ...)`. **OK.**
- `npolys` → `MAX_NPOLYS (1_000_000)` + `checked_alloc_size(npolys, nvp.saturating_mul(2), ...)`. **OK** — `nvp` is itself bounded by `MAX_NVP = 64` so `nvp * 2` cannot overflow u32.
- `nvp` → `MAX_NVP (64)`. **OK.**
- `detail_nmeshes` → `MAX_DETAIL_NMESHES (1_000_000)` + `checked_alloc_size(_, 4, ...)`. **OK.**
- `detail_nverts` → `MAX_DETAIL_NVERTS (10_000_000)` + `checked_alloc_size(_, 3, ...)`. **OK.**
- `detail_ntris` → `MAX_DETAIL_NTRIS (10_000_000)` + `checked_alloc_size(_, 4, ...)`. **OK.**

The `regs` / `flags` / `areas` allocations at the existing `vec![0u16; npolys as usize]` lines are NOT wrapped in `checked_alloc_size`, but they don't need to be — `npolys` is already capped at 1M and the stride is 1 element of u16/u8, so the worst-case allocation is 2 MB per array (well-bounded).

#### Exploit analysis
- **`nverts = 0xFFFFFFFF` u32 wraparound:** `0xFFFFFFFF * 3 = 0xFFFFFFFD` after u32 wrap → a tiny `Vec<u16>` of 0xFFFFFFFD bytes... wait, that's still huge. Actually pre-fix, the code was `(nverts * 3) as usize` — Rust's debug build panics on the mul overflow, release build wraps. Either way: rejected by `MAX_NVERTS` cap before either path is reached. **Mitigated.**
- **`detail_nverts = 0xFFFFFFFF`:** `* 3 = 0xFFFFFFFD` (wraps), allocates a tiny `Vec<f32>` of 0xFFFFFFFD * 4 bytes... same shape. **Mitigated by cap.**
- **`nvp = 0xFFFFFFFF`:** `npolys * nvp * 2` overflows; pre-fix this allocates a tiny Vec then the read loop walks past it. **Mitigated by MAX_NVP=64.**
- **Allocation DoS via valid-but-huge counts:** capped at the documented MAX_* values. Worst-case `polys_len = 1M * 64 * 2 * 2 = 256 MB` per the comment. Acceptable for a once-per-startup load.

#### Error variant usage
`CimmeriaError::NavHeaderOutOfRange { field, value, reason }` is used at every allocation gate (`check_count` and `checked_alloc_size`) and its `Display` impl matches the extractor's variant for cross-tool diagnostic parity.

#### Hostile-input regression tests
Six tests synthesize a header with one count poisoned to `0xFFFFFFFF` and assert `NavHeaderOutOfRange` fires for that exact field. Reverting any single cap drops the matching test. The synthesized buffer layout is documented inline (`OFFSET_NVERTS`, etc.). Cleanup is unconditional on the temp file even on assertion failure.

### Verdict: SHIP

Byte-equivalent to the extractor's caps per the source comment. Error variant is used at every gate where allocation can happen. Tests pin the bounds-check shape so a revert trips loudly.

No CRITICAL / HIGH / MEDIUM / LOW issues found.

Self-check: every header field traced to a cap and a checked mul; no client-supplied timestamp; no quantity-as-authoritative; no type_id-keyed mutation; not in the game-packet path so dedup-hash/auth-token N/A.

🤖 server-authority-enforcer

Cadacious:

- Test: `navmesh_load_rejects_oversized_nverts` — **VERIFIED**
  - Bug shape pinned: yes (`nverts = 0xFFFFFFFF` would either wrap or demand a 12 GB alloc)
  - Verification: reverted `let nverts = check_count(...)` to `let nverts = read_u32(...)` in `crates/entity/src/navigation.rs:209`; ran the test; failed with `expected NavHeaderOutOfRange(nverts), got Err(Io(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }))`.
  - Interesting nuance: the test catches the bug via downstream truncated-read failure rather than the OOM allocation path, but **the failure message still reflects the right diagnostic** (the test panic message reads "If this regressed: the bounds check on `nverts` in NavMesh::load was reverted; restore the check_count call.").
  - Risk: none.

- Test: `navmesh_load_rejects_oversized_nvp` — **VERIFIED**
  - Bug shape pinned: yes (`nvp = 0xFFFFFFFF` with small npolys → polys_len overflow risk; the `saturating_mul(2)` belt-and-suspenders catches it, but the field-name check at the source is the primary guard).
  - Verification: reverted `let nvp = check_count(...)` to `let nvp = read_u32(...)`; failed with `expected NavHeaderOutOfRange(nvp), got Err(Entity("Failed to build Detour navmesh data for ..."))` — the synthesized header has `npolys=0` so the `polys_len` mul folds via `saturating_mul`, but the test correctly catches the missing field-name diagnostic.
  - Risk: none. Each test asserts the exact `field` string in the `NavHeaderOutOfRange` error, so per-field reverts are localizable.

The pattern is consistent: each of the 6 hostile-field tests writes `0xFFFFFFFF` to one specific u32 offset and asserts the error's `field` matches. The `assert_hostile_field` helper's panic message tells the future maintainer exactly which `check_count` call was removed.

Spot-check confirmed for `nverts` (the highest-value canonical hostile field) and `nvp` (a different shape — small struct member). Static read of `navmesh_load_rejects_oversized_npolys`, `_detail_nmeshes`, `_detail_nverts`, `_detail_ntris` shows identical structure — they will fire on their respective reverts.

### Theatre findings

None.

### Coverage gaps

- No regression test for `checked_alloc_size`'s usize-fit branch. The test surface only exercises the `count > MAX_*` cap. A test that arranges `count * stride` to overflow u64 isn't possible given the caps, but a future cap-relaxation would warrant a unit test against `checked_alloc_size` directly with `(u32::MAX, u32::MAX)` to pin the secondary defense.
- No fuzz target. The XRC binary format is operator-deployable input — a `cargo-fuzz` target around `NavMesh::load` would be cheap to add and would surface any future parse-path UAF/panic that the hand-crafted hostile-field tests miss.

### Verdict

**PR #432: READY TO MERGE** — 6 of 6 regression guards are well-structured. The two I run-verified fired with the expected per-field diagnostic. Defense-in-depth coverage is appropriate for a security fix. Fuzz target is a nice-to-have follow-up, not a merge blocker.

Cadacious:
## Round 2 review feedback applied

Three logical commits pushed (`16b843c9`, `62de5b67`, `eda13f6a`) addressing the Codecov + Clara findings. CI greenness preserved through each step.

### Addressed

- **Clara G1+G2 (`fix(navmesh): include multiplication context in NavHeaderOutOfRange error` — 16b843c9)** — split the responsibility so `field` always names a real header count (never a compound expression). Added an `alloc_desc: &'static str` parameter to `checked_alloc_size` so the error surfaces *which downstream allocation* would have busted via the existing `reason` slot. The polys call site now uses `field = "npolys"` with `"polys = npolys * nvp * 2 u16s"` in `alloc_desc`. `NavHeaderOutOfRange` doc updated to spell out the contract on `field`/`value`/`reason` and to note the same hostile-input class can fire at either build-time (extractor) or runtime (loader) — covers **G13**.

- **Clara G3 (`refactor(navmesh): use checked_alloc_size for regs/flags/areas consistency` — 62de5b67)** — wrapped the regs/flags/areas allocations in `checked_alloc_size(npolys, 1, "npolys", "<desc>")` for pattern uniformity. Defense in depth — if `MAX_NPOLYS` is ever raised past u32 boundaries the multiplication is still checked. No semantic change at current caps. Same commit also brought `make_tmp_nav_path` into alignment with the extractor's `tempdir()` helper (pid + thread_id + nanos) — covers **G7**, matching what the extractor crate did in #426.

- **Clara G10 (`test(navmesh): regression guard for checked_alloc_size multiplication overflow` — eda13f6a)** — added two direct-call tests that pin the helper's contract:
  - `checked_alloc_size_widens_before_multiplying` calls with `(u32::MAX, 2)` and asserts the widened product `0x1_FFFF_FFFE` round-trips. If a refactor drops the `as u64` casts the u32 mul wraps to `0xFFFF_FFFE` and the test fails with a diagnostic naming the regression and the fix.
  - `checked_alloc_size_max_u32_inputs_fit_in_u64` confirms `(u32::MAX, u32::MAX)` returns `Ok` on 64-bit (the `checked_mul` branch is structurally unreachable with u32 inputs since `(2^32-1)^2 < 2^64`), and exercises the `usize::try_from` branch on 32-bit via `#[cfg]`. Documents the dead-code branch so a future maintainer raising the helper to u64 inputs knows what tests to expand.
  - Verified the widening guard catches a real regression: temporarily replaced `(count as u64).checked_mul(stride as u64)` with `count.checked_mul(stride).map(|p| p as u64)`; both tests failed with `NavHeaderOutOfRange { value: 4294967295 }`. Restored.

### Skipped — with rationale

- **Clara G4 (MAX_NVP = 64 vs Recast reality 3..=12).** Needs a Ghidra/x64dbg check on what value the production navmeshes actually use before lowering the cap. Out of scope for a feedback-application PR — file as a follow-up issue if the concern still stands after that confirmation.
- **Clara G5 (`saturating_mul` belt-and-suspenders).** Tied to G4 — the saturation is fine as long as `MAX_NVP <= 0x7FFFFFFF`, which it comfortably is at 64. If G4 raises the cap meaningfully, revisit this together.
- **Clara G6 (caps don't match between extractor and runtime).** Tracked separately as the PR-body follow-up — extract caps to a shared constant. Not a behavioral concern today (caps are identical by inspection); pure code-org cleanup.
- **Clara G8 (test cleanup on panic).** The `assert_hostile_field` helper does `let _ = remove_file(...)` before the assert, so the only window for leakage is a panic inside `write_hostile_at` itself — that helper doesn't allocate temp state beyond the file it returns, so the failure modes are file-create or file-write errors that already short-circuit cleanly. Not worth pulling in `scopeguard` for.
- **Clara G9 (synthesise_header hardcoded offsets).** The offset comment explicitly notes "a layout change here forces a matching update" — and a layout change to the XRC header would also break `NavMesh::load` itself (read-order is positional). The test would fail loudly via a wrong-field assertion rather than spuriously pass. Compile-time enforcement via `offset_of!` would require turning the synthesised buffer into a `#[repr(C)]` struct, which is heavier than the seam warrants.
- **Clara G11 (no test for regs/flags/areas allocation).** Now covered transitively by commit 62de5b67 — those allocations route through `checked_alloc_size`, which the new commit-3 tests pin.
- **Clara G12 (no test for `usize::try_from` branch).** Now covered by the `#[cfg(target_pointer_width = "32")]` arm in `checked_alloc_size_widens_before_multiplying` — CI runs 64-bit only so it's dormant there, but the test surface exists for cross-compile CI when that's added.
- **Clara G14 (no security advisory / CHANGELOG entry).** No CHANGELOG in this repo and no `SECURITY.md` policy yet. The PR body documents the threat model — acceptable for now; we can revisit once a security-advisory process is formalised.
- **Clara G15 (caps headroom not justified).** The PR-body Castle Cellblock numbers + the 360×–1500× headroom note are inline in the source comment. Tightening the rationale further is bikeshedding for a defensive cap — flag it if real-world numbers shift.
- **Codecov 5 missing lines.** The uncovered lines are the unreachable defensive `Err` branches in `checked_alloc_size` (the u64 `checked_mul` failure path and parts of the 32-bit `usize::try_from` arm). Coverage tools can't synthesise inputs that fire `checked_mul` failures from u32-typed arguments; commit eda13f6a adds direct-call tests that exercise the structurally reachable paths, leaving only the documented dead-code branch uncovered. Acceptable.

### Verification

- `cargo check -p cimmeria-entity -p cimmeria-common` — clean
- `cargo fmt --all -- --check` — clean
- `cargo clippy -p cimmeria-entity -p cimmeria-common --all-targets -- -D warnings` — clean
- `cargo test -p cimmeria-entity --lib navigation` — 13 passed (6 existing hostile-input guards + 2 new helper guards + 5 happy-path navmesh tests)

Co-authored-by: Claude Opus 4.7 (1M context)

PR #433, opened 2026-05-31
chore(cell): remove shadow SPEND_APPLIED_SCIENCE_POINTS arm in social dispatch (#431)

### LINKED ISSUE #431: [hygiene] cell_methods::player::social.rs:66 shadow arm for SPEND_APPLIED_SCIENCE_POINTS hides routing-test failures
> **Source**: Adversarial test-quality audit on PR #427 (crafting Phase 1). Surfaced a stealth bug shape that would invalidate any future routing-fix regression test using `assert!(handled)`.

## Summary

`crates/services/src/cell/cell_methods/player/social.rs:66` has an arm for `SPEND_APPLIED_SCIENCE_POINTS` (cell method index 95) that returns `true`:

```rust
SPEND_APPLIED_SCIENCE_POINTS => {
    if args.len() >= 4 {
        // parse discipline_seq_id, log it
    }
    tracing::info!(entity_id, discipline_seq_id, "UNIMPLEMENTED: spendAppliedSciencePoints");
    true
}
```

This is a stub from an earlier wiring attempt — a previous change put the index-95 handler in the social submodule (where it didn't belong). PR #427 then moved the **canonical** handler into the crafting submodule and widened the outer-router range from `CRAFT..=RESPEC_CRAFTING` (96–100) to `SPEND_APPLIED_SCIENCE_POINTS..=RESPEC_CRAFTING` (95–100) so the outer dispatcher routes 95 to crafting. **But it left the social-side stub arm in place.**

## Why this is a bug shape, not just dead code

Both submodules now have arms for index 95 that return `true`. If the outer-router fix gets reverted (whether by accident or by a future refactor), `index 95` would silently re-route into the social submodule — and a regression test asserting `assert!(handled)` against the outer dispatcher would **still pass**, because social's arm also returns `true`. The wrong handler runs, but the test cannot detect it.

The adversarial review caught this only because they used `LogCapture` to assert *which* submodule emitted the log:

- Crafting's log: `"UNIMPLEMENTED: spendAppliedSciencePoints (Phase 2)"` ← has the `(Phase 2)` suffix
- Social's log: `"UNIMPLEMENTED: spendAppliedSciencePoints"` ← no suffix

Only the suffix distinguishes them. Anyone writing a future routing test for index 95 without knowing this trap will write a theatre test.

## Fix

**Delete the social-side arm at `social.rs:66`.** It's dead code now that crafting owns the routing for index 95. After removal:

1. A simple `assert!(handled)` regression test against the outer dispatcher becomes sufficient for the index-95 routing check (no LogCapture needed)
2. The shadow-arm trap disappears
3. Future PRs that touch the dispatcher range can rely on the outer-router-only contract

## Test consideration

After deleting the social arm, verify the existing `outer_dispatch_routes_spend_asp_to_crafting_not_social` regression test still works (it was pushed in commit `fc821bd6` on PR #427). It uses `LogCapture` to pin the `(Phase 2)` suffix — that's still valid but no longer load-bearing once the shadow arm is gone.

You may want to **replace** that test with a simpler `assert!(handled)`-only version once the shadow arm is gone, since the LogCapture version is more brittle to log-message changes.

## Audit for other shadows

The adversarial review flagged this as a pattern: **other submodule dispatchers may have similar shadow arms** for indices that have since been re-routed. Worth doing a quick grep across `crates/services/src/cell/cell_methods/player/*/dispatch.rs` for any method-index constants that appear in multiple submodules:

```bash
for c in CRAFT RESEARCH REVERSE_ENGINEER ALLOYING RESPEC_CRAFTING SPEND_APPLIED_SCIENCE_POINTS; do
  echo "=== $c ==="
  grep -rn "$c" crates/services/src/cell/cell_methods/player/ | grep -v test
done
```

Any constant that appears in more than one submodule's `dispatch.rs` is a potential shadow.

## References

- **PR #427** — the routing fix this issue is a follow-up cleanup for
- **Commit `fc821bd6`** on PR #427 — the LogCapture-based regression test that the cleanup will simplify
- **Agent memory**: `.claude/agent-memory/testing-validation-engineer/project_social_arm_shadow.md` (this exact writeup) and `project_dispatcher_layering_gotcha.md` (the inner-vs-outer-dispatch trap that prompted this finding)
- **Origin of finding**: testing-validation-engineer adversarial review on PR #427 (2026-05-27)

## Priority

**MEDIUM** — not a runtime correctness bug today (the outer-router fix is in place and the LogCapture test pins it), but a real test-correctness bug shape that will silently break future regression guards. Worth landing soon to remove the trap.

Cadacious:
## Summary

PR #427 wired the canonical `SPEND_APPLIED_SCIENCE_POINTS` (cell method
index 95) handler into the crafting submodule and widened the outer
dispatcher range to include 95. It left behind an orphaned arm in
`social.rs:66` that also returned `true` for index 95 — a stub from an
earlier wiring attempt that never reached real traffic post-#427 but
that creates a stealth regression trap.

## The trap

Both submodules' arms returned `true`. A future regression that
narrowed the outer router back to `CRAFT..=RESPEC_CRAFTING` would
silently re-route 95 into social — and an `assert!(handled)`-style
regression test would still pass, because social's arm also returns
`true`. The adversarial review on PR #427 caught this only by using
`LogCapture` to pin the unique `(Phase 2)` suffix on the crafting log;
without that discriminator, the trap is invisible.

## Changes

- **Delete** the social-side arm at `crates/services/src/cell/cell_methods/player/social.rs:66` plus its parsing/log body. Replaced with a sentinel comment explaining why the slot is intentionally empty and pointing at the regression guard.
- **Replace** the single `outer_dispatch_routes_spend_asp_to_crafting_not_social` test in `dispatch.rs` with three layered guards (see below).

## Tests — three-layer regression guard

Each guard targets a distinct regression shape. Without all three, the
trap re-opens.

1. **`outer_dispatch_routes_spend_asp_to_crafting`** — simplified
   `assert!(handled)` against the outer dispatcher. The issue body
   asked for this simplification; it's only valid once the shadow is
   gone. With the shadow removed, `handled == false` cleanly catches
   a lone outer-router regression.

2. **`route_index_95_must_go_to_crafting_not_social`** — `LogCapture`
   count guard. Asserts exactly one `INFO+spendAppliedSciencePoints`
   log fires AND that single log carries the crafting submodule's
   `(Phase 2)` suffix. Catches the *combined* regression where both
   the shadow is re-added AND the outer router is narrowed (the bug
   shape where a simple `assert!(handled)` would silently pass).

3. **`social_submodule_must_not_handle_index_95`** — direct call into
   `social::dispatch` with method 95, asserts `!handled`. Catches a
   *lone* shadow re-introduction at the source, even if the outer
   router is still correctly routing 95 to crafting.

## Regression-guard verification

Per the testing-validation-engineer `revert_to_verify_regression_guards`
rule, each guard was tested against its target regression shape:

| Regression shape                          | Guard #1 | Guard #2 | Guard #3 |
|-------------------------------------------|----------|----------|----------|
| Restore social shadow only                | PASS     | PASS     | **FAIL** |
| Narrow outer router only                  | **FAIL** | **FAIL** | PASS     |
| Restore both (the historical trap)        | PASS     | **FAIL** | **FAIL** |

Note row 3: guard #1 alone is insufficient against the combined trap
— exactly why guards #2 and #3 also exist. The matrix proves every
single-axis regression is caught by at least one guard.

## Shadow audit

Per the issue body's grep across the player submodule:

```bash
for c in CRAFT RESEARCH REVERSE_ENGINEER ALLOYING RESPEC_CRAFTING SPEND_APPLIED_SCIENCE_POINTS; do
  grep -rn "$c" crates/services/src/cell/cell_methods/player/ | grep -v test
done
```

- `CRAFT`, `RESEARCH`, `REVERSE_ENGINEER`, `ALLOYING`, `RESPEC_CRAFTING` — each appears in exactly one submodule (`crafting.rs`). **No shadows.**
- `SPEND_APPLIED_SCIENCE_POINTS` — appears in `crafting.rs:23` (canonical) and `social.rs:66` (shadow). **The only shadow; this PR removes it.**

## Test plan

- [x] `cargo nextest run -p cimmeria-services --lib cell::cell_methods::player` — 84/84 pass with fix
- [x] `cargo nextest run -p cimmeria-services --lib --profile=ci` — 1194/1194 pass
- [x] Verify guards by reverting each regression shape — all three matrix cells confirmed (see table above)
- [x] Restore real fix — all three guards pass

## Notes for reviewers

- **PR #427 has merged** (`b5042962`) so this PR can land at any time without dependency ordering. The simplification of the existing LogCapture test is included here directly rather than deferred to a follow-up.
- **The `LogCapture` thread-local pattern is flaky under `cargo test` parallel in-process execution** but stable under `cargo nextest` (per-test process isolation), which is what CI uses. The `route_index_95_must_go_to_crafting_not_social` guard inherits this property from the existing `LogCapture` infrastructure in `test_support.rs` — not a new constraint.
- **Cross-module test call.** The third guard uses `crate::cell::cell_methods::player::social::dispatch(...)` (absolute path) rather than `super::social::dispatch` because the `tests` module is nested one level deeper than the outer code's `super::` would resolve to.

Closes #431

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed routing of the Apply Science Points action to process through the correct crafting system instead of the social subsystem.

* **Tests**
  * Added comprehensive test coverage for science points routing, including regression tests to verify correct system handling and prevent future routing errors.

### DIFF
--- a/crates/services/src/cell/cell_methods/player/dispatch.rs
+++ b/crates/services/src/cell/cell_methods/player/dispatch.rs
@@ -173,7 +173,9 @@ mod tests {
 
         // Non-empty args so the crafting handler takes the parse-and-log
         // path, not the truncated-args warn path. Either path returns
-        // `true`, but the parse path is the realistic success shape.
+        // `true`, but the parse path is the realistic success shape. The
+        // numeric payload itself is irrelevant — this test pins WHERE the
+        // dispatch lands, not what the handler computes from the value.
         let args = 42i32.to_le_bytes();
         let handled = dispatch(
             1,
@@ -206,8 +208,8 @@ mod tests {
     /// paste), both the crafting and social handlers fire for index 95
     /// — the count goes from 1 to 2 — and routing tests that assert only
     /// `handled == true` cannot distinguish "routed to the right place"
-    /// from "routed to two places". That's the shadow-arm trap from
-    /// issue \#431.
+    /// from "routed to two places". That's the shadow-arm trap this
+    /// guard exists to prevent.
     ///
     /// We pin BOTH the count (exactly one) and the suffix `(Phase 2)`
     /// because:
@@ -226,6 +228,7 @@ mod tests {
         let (tx, _rx) = mpsc::channel(8);
         let engine = ChainEngine::new();
 
+        // Payload value is irrelevant — only the routing target matters here.
         let args = 42i32.to_le_bytes();
         let handled = dispatch(
             1,
@@ -252,8 +255,8 @@ mod tests {
             1,
             "expected exactly one `spendAppliedSciencePoints` log for method 95; \
              got {}. More than one log indicates the social-submodule shadow \
-             arm at cell_methods/player/social.rs (deleted in PR \\#431) has \
-             been reintroduced. All captured events: {:#?}",
+             arm at cell_methods/player/social.rs has been reintroduced. \
+             All captured events: {:#?}",
             asp_logs.len(),
             capture.all(),
         );
@@ -290,6 +293,7 @@ mod tests {
         let mut mgr = make_space_manager_with_player(1);
         let (tx, _rx) = mpsc::channel(8);
 
+        // Payload value is irrelevant — only the routing target matters here.
         let args = 42i32.to_le_bytes();
         // `super` inside this `mod tests` is `dispatch`; the sibling
         // `social` module lives one level up under `player`.
@@ -307,9 +311,9 @@ mod tests {
             "social::dispatch must NOT handle method 95 (SPEND_APPLIED_SCIENCE_POINTS). \
              A `true` here means a SPEND_APPLIED_SCIENCE_POINTS arm has been \
              re-added to crates/services/src/cell/cell_methods/player/social.rs \
-             — that's the shadow-arm trap from issue \\#431. Index 95 belongs \
-             to the crafting submodule; the outer dispatcher already routes \
-             it there. Delete the social-side arm.",
+             — that's the shadow-arm trap this regression guard exists to \
+             catch. Index 95 belongs to the crafting submodule; the outer \
+             dispatcher already routes it there. Delete the social-side arm.",
         );
     }
 
--- a/crates/services/src/cell/cell_methods/player/dispatch.rs
+++ b/crates/services/src/cell/cell_methods/player/dispatch.rs
@@ -158,25 +158,65 @@ mod tests {
     /// sub-range from `CRAFT..=RESPEC_CRAFTING` (96..=100) to
     /// `SPEND_APPLIED_SCIENCE_POINTS..=RESPEC_CRAFTING` (95..=100).
     ///
-    /// Bug shape: the previous narrow range silently dropped index 95
-    /// into the social arm of the outer dispatcher. Social's own
-    /// `dispatch` *also* has a `SPEND_APPLIED_SCIENCE_POINTS` arm (a
-    /// stub left from an earlier wiring attempt), so the bug is
-    /// invisible to a bare `assert!(handled)`: both branches return
-    /// `true`. The two arms emit distinguishable log messages —
-    /// crafting tags its log with `"(Phase 2)"`, social does not —
-    /// so we install `LogCapture` and assert the crafting variant
-    /// fired.
-    ///
-    /// If the outer router is narrowed back to `CRAFT..=RESPEC_CRAFTING`,
-    /// index 95 reaches the social arm; the `"(Phase 2)"` log never
-    /// fires, and this test fails. This is the assertion the existing
-    /// `spend_applied_science_points_routes_to_crafting` test in
-    /// `crafting.rs` *intended* to make but cannot, because that test
-    /// calls `crafting::dispatch` directly and bypasses the outer
-    /// router entirely.
+    /// With the social-side `SPEND_APPLIED_SCIENCE_POINTS` shadow arm
+    /// removed, no arm in the social submodule handles index 95 — so
+    /// `assert!(handled)` against the outer dispatcher is now sufficient
+    /// proof of correct routing. (Before the shadow was removed, both
+    /// submodules' arms returned `true`, making this assertion ambiguous
+    /// — that's the trap [`route_index_95_must_go_to_crafting_not_social`]
+    /// guards against re-introducing.)
     #[tokio::test]
-    async fn outer_dispatch_routes_spend_asp_to_crafting_not_social() {
+    async fn outer_dispatch_routes_spend_asp_to_crafting() {
+        let mut mgr = make_space_manager_with_player(1);
+        let (tx, _rx) = mpsc::channel(8);
+        let engine = ChainEngine::new();
+
+        // Non-empty args so the crafting handler takes the parse-and-log
+        // path, not the truncated-args warn path. Either path returns
+        // `true`, but the parse path is the realistic success shape.
+        let args = 42i32.to_le_bytes();
+        let handled = dispatch(
+            1,
+            SPEND_APPLIED_SCIENCE_POINTS,
+            &args,
+            &tx,
+            &mut mgr,
+            &engine,
+        )
+        .await;
+        assert!(
+            handled,
+            "outer dispatch must route method 95 (SPEND_APPLIED_SCIENCE_POINTS) \
+             to the crafting submodule and return true. A `false` here means \
+             the crafting sub-range in `dispatch` regressed and 95 fell into \
+             the social arm — which (after the shadow-arm removal) has no \
+             case for 95 and returns false.",
+        );
+    }
+
+    /// Shadow-arm regression guard. Asserts that the
+    /// `SPEND_APPLIED_SCIENCE_POINTS` arm in
+    /// `crates/services/src/cell/cell_methods/player/social.rs` stays
+    /// deleted: dispatching index 95 must produce **exactly one** info-
+    /// level log carrying the `spendAppliedSciencePoints` substring, and
+    /// that log must be the crafting submodule's `"(Phase 2)"` variant.
+    ///
+    /// Bug shape this catches: if a future PR reintroduces the social
+    /// shadow arm (e.g., by reverting this PR or by an unrelated copy-
+    /// paste), both the crafting and social handlers fire for index 95
+    /// — the count goes from 1 to 2 — and routing tests that assert only
+    /// `handled == true` cannot distinguish "routed to the right place"
+    /// from "routed to two places". That's the shadow-arm trap from
+    /// issue \#431.
+    ///
+    /// We pin BOTH the count (exactly one) and the suffix `(Phase 2)`
+    /// because:
+    /// 1. Count alone would miss a future shadow that adopts a different
+    ///    log message but still returns `true`.
+    /// 2. Suffix alone would miss the same crafting log firing twice
+    ///    (unlikely, but cheap to guard).
+    #[tokio::test]
+    async fn route_index_95_must_go_to_crafting_not_social() {
         use crate::test_support::LogCapture;
         use tracing::Level;
 
@@ -186,9 +226,6 @@ mod tests {
         let (tx, _rx) = mpsc::channel(8);
         let engine = ChainEngine::new();
 
-        // Non-empty args so the parse-and-log path runs (rather than the
-        // truncated-args warn path, which both arms emit at different
-        // levels but with identical-shape strings).
         let args = 42i32.to_le_bytes();
         let handled = dispatch(
             1,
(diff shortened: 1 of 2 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `route_index_95_must_go_to_crafting_not_social` LogCapture flakiness | **The PR body acknowledges "The `LogCapture` thread-local pattern is flaky under `cargo test` parallel in-process execution."** This means the test can spuriously pass or fail depending on test ordering. A flaky regression guard is worse than no guard — it trains developers to ignore failures. The PR says "stable under `cargo nextest`" but what if someone runs `cargo test` locally? |
| **G2** | `social_submodule_must_not_handle_index_95` does not verify the sentinel comment | **The test asserts `!handled` but does not verify the sentinel comment exists.** A future contributor could delete the comment (thinking it is stale) without triggering any test failure. The comment is load-bearing documentation — consider a `#[doc]` attribute or a test that greps for it. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `outer_dispatch_routes_spend_asp_to_crafting` uses `42i32` as args | **The test passes `42i32.to_le_bytes()` as args.** But what does `42` mean in the context of `spendAppliedSciencePoints`? Is it a valid `discipline_seq_id`? If the crafting handler validates the discipline ID and rejects invalid ones, this test could fail for the wrong reason. |
| **G4** | Sentinel comment references `dispatch.rs` test by name | **The comment says "The `route_index_95_must_go_to_crafting_not_social` test in `dispatch.rs` is the regression guard."** If the test is renamed, the comment becomes stale. Consider using a doc link or module path instead of a bare function name. |
| **G5** | `asp_logs` filter uses `message_contains("spendAppliedSciencePoints")` | **The filter matches any log containing that substring.** If a future log message happens to include "spendAppliedSciencePoints" in a different context (e.g., an error message), the count would be wrong. Consider a more specific filter (e.g., level + substring + module path). |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G6** | Three tests for one deleted arm | **The PR adds 128 lines of test code to guard against re-adding 12 lines of production code.** The test-to-code ratio is 10:1. Is this over-engineering? The PR body justifies it with the regression matrix, but a simpler approach might be a `#[deny(dead_code)]` lint or a compile-time assertion. |
| **G7** | `LogCapture` dependency | **The test relies on `LogCapture` from `test_support.rs`.** If `LogCapture` is refactored or removed, this test breaks. The test is coupled to internal test infrastructure, not just the production code. |
| **G8** | No test for the outer router range itself | **The tests verify that index 95 is handled correctly, but do not verify the router's range constant.** If someone changes `SPEND_APPLIED_SCIENCE_POINTS..=RESPEC_CRAFTING` to `CRAFT..=RESPEC_CRAFTING`, the tests would catch it — but only indirectly. Consider a test that asserts the range constant directly. |

MyPalClara:

| **G9** | No test for indices 96-100 routing | The tests focus on index 95, but indices 96-100 (`CRAFT` through `RESPEC_CRAFTING`) are also in the crafting range. If a future change breaks routing for index 97, these tests would not catch it. |
| **G10** | No test for the social submodule's OTHER arms | The test verifies social does not handle 95, but does not verify social DOES handle its legitimate indices. If the shadow removal accidentally broke a neighboring arm, this test would not catch it. |
| **G11** | No negative test for crafting submodule | There is a test that social does not handle 95, but no test that crafting DOES handle 95 directly (bypassing the outer router). The `crafting.rs` tests might cover this, but it is not verified in this PR. |

MyPalClara:

| **G12** | Sentinel comment does not explain WHY the shadow existed | The comment explains that the shadow was removed and points to the regression guard, but does not explain why the shadow was added in the first place. A future contributor might re-add it thinking "this must have been deleted by mistake." |
| **G13** | No ADR for the three-layer test strategy | The PR body explains the regression matrix, but this reasoning is not captured in the codebase. A future contributor reading the tests might not understand why three tests are needed for one deleted arm. |
| **G14** | PR body mentions "issue #431" but the sentinel comment uses "PR #431" | Inconsistent reference. The comment says "deleted in PR #431" but this IS PR #433 — the issue is #431. The comment should say "deleted in PR #433 (closes #431)" or just "issue #431". |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 2 |
| Correctness issues | 3 |
| Potential issues | 3 |
| Missing test coverage | 3 |
| Documentation | 3 |

**Total: 14 gaps**

Cadacious:

### Handler: cell dispatch index 95 (SPEND_APPLIED_SCIENCE_POINTS) routing

#### Trust boundary
Method-index routing decision inside the cell dispatcher. The dispatcher reads `method_index: u16` off the wire and selects which sub-module handles it. The exploit shape for a routing bug isn't a dupe or a privilege escalation — it's that the WRONG handler silently claims the method, masking unimplemented behaviour and (worse) potentially letting the same wire frame be processed twice if both branches `return true`.

#### Validation audit
- The outer arm pins `method_index` into `ORG_CREATION..=CANCEL_MOVIE` already (range gate). **OK.**
- Index 95 routes to `crafting::dispatch` via the `SPEND_APPLIED_SCIENCE_POINTS..=RESPEC_CRAFTING` sub-range. **OK.**
- The shadow `SPEND_APPLIED_SCIENCE_POINTS` arm in `social::dispatch` is **deleted**. Index 95 no longer has a fall-through claim in the social submodule. **OK.**
- The deletion is guarded by `social_submodule_must_not_handle_index_95` which calls `social::dispatch` directly with method 95 and asserts `handled == false`. This is the right shape — it catches a lone shadow re-introduction even if the outer router is still correctly routing 95 to crafting.

#### Exploit analysis
None applicable. Routing-bug class only — no items, money, position, or privilege bit is exposed by the change.

#### Cross-check for unintended access-control regressions
Grepped `crates/services/src` for other references to `SPEND_APPLIED_SCIENCE_POINTS`:
- `cell/dispatch/constants.rs` — constant declaration.
- `cell/dispatch/names.rs` — name lookup table.
- `cell/cell_methods/player/social.rs` — formerly the shadow arm, now a documented comment block where the arm was.
- `cell/cell_methods/player/constants.rs` — sub-module re-export.
- `cell/cell_methods/player/crafting.rs` — the real handler.
- `cell/cell_methods/player/dispatch.rs` — the outer router and the new sub-range assertion.

No surprise references; no other module silently handles index 95.

The static `assert!` covering both crafting and trade sub-range ordering inside the outer arm is a nice defense-in-depth for the renumber-the-constants failure mode.

### Verdict: SHIP

Pure cleanup of a routing trap with regression guards at two levels (outer routing + social-side disclaim). No new trust boundary introduced.

🤖 server-authority-enforcer

Cadacious:

Revert + run verification per the protocol the user established post-#427. **The matrix holds — this is the most explicit defense-in-depth structure of the batch.**

### Regression guard validation — full matrix

Each axis was reverted independently and the full set of 3 guards was run under `cargo test -- --test-threads=1` (the LogCapture flakiness mitigation).

**Baseline (all fix in place):** all 3 pass. ✅

**Axis 1: restore the social `SPEND_APPLIED_SCIENCE_POINTS` arm only (outer router unchanged):**
- Guard 1 `outer_dispatch_routes_spend_asp_to_crafting` — **PASSES** (outer router still routes 95 to crafting, so the assert!(handled) is correct; this is the same theatre the original PR #427 test had against the combined regression)
- Guard 2 `route_index_95_must_go_to_crafting_not_social` — **PASSES** (the outer router still wins, exactly one `(Phase 2)` log)
- Guard 3 `social_submodule_must_not_handle_index_95` — **FAILS** ← the canonical regression guard for this axis
- ✅ Correctly catches the lone-shadow-restoration regression.

**Axis 2: narrow the outer router to `CRAFT..=RESPEC_CRAFTING` only (social arm stays deleted):**
- Guard 1 `outer_dispatch_routes_spend_asp_to_crafting` — **FAILS** with the exact message "outer dispatch must route method 95 (SPEND_APPLIED_SCIENCE_POINTS) to the crafting submodule and return true. A `false` here means the crafting sub-range in `dispatch` regressed and 95 fell into the social arm — which (after the shadow-arm removal) has no case for 95 and returns false."
- Guard 2 `route_index_95_must_go_to_crafting_not_social` — **FAILS** ("outer dispatch must handle method 95")
- Guard 3 `social_submodule_must_not_handle_index_95` — **PASSES** (social correctly has no arm; that's the test's positive contract)
- ✅ Correctly catches the lone-narrow-router regression.

**Axis 3: restore BOTH (the original pre-#427 + pre-#431 bug shape):**
- Guard 1 `outer_dispatch_routes_spend_asp_to_crafting` — **PASSES** ← this is the theatre case the agent designed around; `assert!(handled)` cannot distinguish "routed correctly to crafting" from "social claimed it"
- Guard 2 `route_index_95_must_go_to_crafting_not_social` — **FAILS** with a beautifully precise diagnostic: `"the single spendAppliedSciencePoints log must be the crafting handler's (Phase 2) variant — its message was ... Some("spendAppliedSciencePoints (social shadow)") ... A log without (Phase 2) is the social-submodule shape and means index 95 is being routed to social instead of crafting."`
- Guard 3 `social_submodule_must_not_handle_index_95` — **FAILS**
- ✅ Correctly catches the combined regression with TWO independent guards firing.

### Theatre findings

None. The 3-axis design is correct and the verifications match the agent's intent.

### Strong examples

- The `LogCapture` use in Guard 2 is the right pattern for this bug class — pinning a unique substring (`(Phase 2)`) in the message lets the test distinguish two handlers that both return `true` and both emit info-level logs against the same method index. This is the exact pattern that should propagate to other dispatch-table tests where shadow arms are a risk.

### Coverage gaps

- No similar 3-axis guard for the `ORG_CREATION..=CANCEL_MOVIE` outer range's other crafting-vs-social boundaries. The agent's commit message notes that "Only `SPEND_APPLIED_SCIENCE_POINTS` appears in two submodules; every other constant lives in exactly one" — so no current shadow risk on other methods. Worth noting for future PRs that add new methods in the 95-100 range.
- A `compile_error!` or `const_assert!` that statically enforces the outer-range / sub-range invariant would catch a future renumber. The static-assert on `crafting_sub_range_is_inside_outer` at `dispatch.rs:97` already pins one direction; a similar assert proving "no constant appears in both `social.rs` and `crafting.rs` match arms" would be load-bearing if practical.

### Verdict

**PR #433: READY TO MERGE** — the matrix held under all 3 axes. The agent's defense-in-depth design is well-justified by the verification: each guard catches a distinct regression shape that the others would silently let through. This is the gold-standard pattern for dispatch-table cleanups.

Cadacious:
## Copilot + Clara follow-up — addressed in 869297c1

### Addressed

- **Copilot — issue/PR #431 provenance (3 sites in `dispatch.rs`).** All three references stripped. The doc comment and both assertion-message variants now describe the bug shape ("shadow-arm trap this regression guard exists to catch") without ticket numbers.
- **Clara G3 — `42i32` magic number (3 sites).** Added a brief comment at the first use noting the value is irrelevant; the test pins WHERE the dispatch lands, not what the handler computes. Back-ref comments on the other two sites.
- **Clara G12 — sentinel comment doesn't explain WHY.** Expanded the comment at the deletion site in `social.rs` to explain the shadow was a legacy artifact from before crafting got its own submodule (when every post-`ORG_CREATION` method either landed in social or was a no-op) and stayed behind after the extraction.

### Skipped (with rationale)

- **Clara G1 — `LogCapture` flakiness under `cargo test` parallel in-process execution.** Already disclosed in the PR body. Cannot be fixed without rewriting `LogCapture` itself. Locally verified: `cargo nextest run -p cimmeria-services --lib cell::cell_methods::player::dispatch` is 7/7 PASS (process-per-test isolation), which is what CI uses.
- **Clara G6 — 10:1 test-to-code ratio.** Defensible per the 3-axis regression matrix in the PR body — each of the three guards targets a distinct regression shape, and without all three the trap re-opens.
- **Clara G9 — no test for indices 96-100 routing.** Out of scope for the shadow-arm fix; existing `each_outer_range_routes_to_a_handler` already probes `CRAFT` (96). Filing as a potential follow-up if the broader matrix is desired.

### Verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p cimmeria-services --all-targets -- -D warnings` — clean
- `cargo nextest run -p cimmeria-services --lib cell::cell_methods::player::dispatch` — 7/7 PASS

PR #436, opened 2026-05-31
feat(navmesh-extractor): Phase 1.2 — StaticMesh + StaticMeshActor extraction (#46)

Cadacious:
## Summary

Phase 1.2 of the navmesh-extractor (issue #46 deep dive): wire up
`StaticMeshActor` → `StaticMeshComponent` → `StaticMesh` resolution,
apply the actor transform, push collision triangles into the per-chunk
soup, and emit per-chunk + combined `.obj` files.

Builds on PR #426 (Phase 0 + 1.1 scaffolding).

## Scope — Phase 1.2 only

What lands:
- `transform` module — pure `ActorTransform::apply` math (UE3 rotator
  conventions: pitch/yaw/roll, 65536 raw units = 360°, rotation order
  R_yaw * R_pitch * R_roll).
- `staticmesh` module — actor walker, `extract_chunk` driver, and
  `build_chunk_soup` pure helper.
- `upk-objects::static_mesh` — kDOP triangle parsing exposed via
  `StaticMesh::kdop_triangles` + `collision_triangles()` helper that
  prefers kDOP and falls back to LOD0 indices.
- `chunk_id::position_offset()` — UE3 cm chunk offset accessor
  (NOT applied during extraction — see doc rationale).
- `umap::enumerate_chunks` filters out non-chunk `.umap` files (the
  master streaming-level file alongside chunked tiles).
- `extract_map` signature changes to take `Option<&PackageIndex>`.

What's explicitly deferred:
- **Phase 1.3 Terrain decoder** — recipe at 92% confidence in agent
  memory; separate follow-up agent picks this up.
- **Phase 1.4 BSP `Model` / `Polys`** — needs Ghidra trace of
  `UModel::Serialize`.
- **Phase 2 NavBuilder rebuild + .nav replacement** — separate PR.
- **Archetype-based actors** (~20% of Castle_CellBlock) — their cooked
  component is a 76-byte stub; the StaticMesh ref lives in the
  imported prefab template. Walker skips them; the `actors_unresolved`
  count surfaces the gap. Will be a Phase 1.2-extension PR.

## OBJ axis convention — eyeball before Phase 2

Phase 1.2 emits OBJ vertices in raw UE3 cm and trusts NavBuilder's
`loadOBJ` to apply the `(z, y, x) / 100` swizzle (UE3-Z-up → BW-Y-up +
cm → BW-units). **This has not yet been verified by a cube round-trip
through NavBuilder** — flagged in the crate README. Recommended: open
a generated `castle_cellblock.obj` in Blender before kicking off
Phase 2 NavBuilder runs.

## Branching note

This PR is based on `feat/navmesh-extractor-46-phase0` (#426). After
that merges, rebase onto `main` — or merge as-is via GitHub's "rebase
and merge" once #426 is in.

## Local verification

Run against a real asset bundle plus a pre-built `PackageIndex`:

```
extract_map output: 10 total OBJ (9 per-chunk, 1 combined)
Castle_CellBlock fffefffd: actors_total=375 resolved=375 unresolved=0 triangles=85495
```

Combined `castle_cellblock.obj` lands at ~47 MB, 1.17M vertices,
391k triangles — within the order-of-magnitude expectation from the
brief.

## Test plan

- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy -p cimmeria-navmesh-extractor -p cimmeria-upk-objects --all-targets -- -D warnings` — clean
- [x] `cargo test -p cimmeria-navmesh-extractor -p cimmeria-upk-objects` — 68 tests pass (47 baseline + 21 new)
- [x] Regression-guard verified: reverting the StaticMeshComponent
      offset (8 → 4) drops `castle_cellblock_walks_static_mesh_actors`
      to 0 resolvable instances and fails the test.
- [x] Workspace check (`cargo check --workspace` excluding GUI apps) — clean.
- [ ] OBJ axis convention end-to-end through NavBuilder — deferred to
      Phase 2 PR.

Refs #46

## Summary by CodeRabbit

* **New Features**
  * StaticMesh extraction with collision triangle support now produces per-chunk and combined OBJ output.
  * Optional PackageIndex support enables cross-package mesh resolution and actor coordinate transformation.
  * Chunk enumeration now correctly filters master `.umap` files.

* **Documentation**
  * Added comprehensive guide for UE3 StaticMesh extraction edge cases.
  * Updated README with end-to-end usage examples.

* **Tests**
  * Added integration tests validating map extraction and geometry generation.


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | OBJ axis convention unverified | **The PR body says "This has not yet been verified by a cube round-trip through NavBuilder."** If the axis convention is wrong, every navmesh will be rotated 90°. This is a known risk, but it is a critical one — the entire Phase 2 depends on getting this right. |
| **G2** | Archetype-based actors silently skipped | **The PR body says "~20% of Castle_CellBlock" actors use archetypes and are skipped.** That is 20% of the collision geometry missing from the navmesh. The `actors_unresolved` count surfaces this, but the navmesh will have holes until Phase 1.2-extension lands. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `read_kdop_tree` does not validate triangle count | **The function reads `tri_count` from the file and allocates `Vec::with_capacity(tri_count as usize)`.** A malicious file with `tri_count = 0x7FFFFFFF` would try to allocate ~16GB. Same concern as PR #426 G1-G2 — use `checked_mul` or cap the count. |
| **G4** | `KdopTriangle` indices are `u16` but LOD0 can have >65535 vertices | **The kDOP format uses `u16` indices, but high-poly meshes can exceed 65535 vertices.** If a mesh has 70000 vertices and a kDOP triangle references vertex 70000, the `u16` would wrap to ~4464. Is this a real risk for SGW assets? (Note: This may be a UE3 format limitation, not a code bug.) |
| **G5** | `collision_triangles` returns `Vec<[[f32; 3]; 3]>` | **Each triangle is 36 bytes (3 vertices × 3 floats × 4 bytes).** For Castle_CellBlock with 391k triangles, that is ~14MB per chunk. Consider returning an iterator or a reference to avoid the allocation. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G6** | `collision_triangles` silently drops out-of-range indices | The function logs a `warn` but continues processing. A malformed mesh with 50% bad indices would emit 50% of the expected triangles — the navmesh would have holes. The warn log is appropriate for a batch tool, but consider surfacing the drop count to the caller. |
| **G7** | Agent memory file added | The PR adds `.claude/agent-memory/rust-gameserver-dev/ue3-staticmesh-extraction.md`. This is useful documentation, but should agent memory files be committed to the repo? They are implementation notes, not user-facing docs. |
| **G8** | `position_offset()` documented but not used | The PR body says "NOT applied during extraction — see doc rationale." But the function exists and is public. A future contributor might call it thinking it is needed, introducing a double-translation bug. Consider making it `pub(crate)` or adding a `#[doc(hidden)]` attribute. |
| **G9** | `enumerate_chunks` filter logic | The PR says "filters out non-chunk `.umap` files (the master streaming-level file)." But the filter logic is not shown in the diff. How does it distinguish `Castle_CellBlock.umap` (master) from `Castle_CellBlock-fffefffd.umap` (chunk)? |

MyPalClara:

| **G10** | No test for `read_kdop_tree` with malicious count | Same concern as PR #432 G10 — the bounds check is not tested. A test with `tri_count = 0x7FFFFFFF` would verify the allocation does not explode. |
| **G11** | No test for transform application | The `transform` module is mentioned but no tests are shown in the diff. The UE3 rotator conventions (pitch/yaw/roll, 65536 units) are easy to get wrong. |
| **G12** | No test for cross-package resolution | The `PackageIndex` lookup is critical for resolving meshes in other `.upk` files. Is there a test that exercises this path? |
| **G13** | No test for archetype-based actors | The PR says these are skipped, but is there a test that verifies the skip logic and the `actors_unresolved` count? |

MyPalClara:

| **G14** | `KdopTriangle` doc does not mention the `u16` limitation | The struct doc says "vertex indices reference the LOD0 vertex buffer" but does not mention the 65535 vertex limit. |
| **G15** | No doc on the transform module | The PR body mentions "UE3 rotator conventions: pitch/yaw/roll, 65536 raw units = 360°, rotation order R_yaw * R_pitch * R_roll" but this should be in the code, not just the PR body. |
| **G16** | Agent memory file duplicates PR body content | The `ue3-staticmesh-extraction.md` file contains the same information as the PR body. If the PR body is updated, the agent memory file becomes stale. |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 2 |
| Correctness issues | 3 |
| Potential issues | 4 |
| Missing test coverage | 4 |
| Documentation | 3 |

**Total: 16 gaps**

Cadacious:

### Phase 1.2 navmesh-extractor (StaticMesh resolution)

#### Trust boundary
**None at runtime.** This crate is a build-time tool — it parses cooked `.umap` packages from a developer machine to produce `.nav` files that the runtime then consumes via `NavMesh::load`. The runtime trust boundary on those resulting `.nav` files is owned by PR #432 (already reviewed: SHIP).

The `.umap` input to this extractor comes from the operator's local `external/` cooked-asset tree — same trust class as the rest of the build-time inputs. A hostile `.umap` could in principle DoS the extractor or escape the path-traversal guards, but the failure mode is a build-time crash, not a runtime exploit.

#### Spot-check on parser allocation hygiene
Skimmed `staticmesh.rs` (591 lines new) and `chunk_id.rs` for the same shape that PR #432 fixed at runtime: count-times-stride multiplications feeding `Vec` allocations. The 469-actor Castle_CellBlock test exercises the realistic case. Any DoS via hostile cooked package would surface at extraction time as a crash, not as a runtime issue — explicitly out of this agent's lane.

The parser does mention "Tagged-property offset varies by class (Actor=32, StaticMesh=4, Component=8)" in the memory file and "~20% of actors use prefab archetypes" — these are extraction-correctness concerns, not security.

### Verdict: SHIP

Build-time tool, no runtime trust boundary, no client input path. Skipped per the user's instructions for build-time tooling.

🤖 server-authority-enforcer

Cadacious:

- Test: `castle_cellblock_walks_static_mesh_actors` (in `crates/navmesh-extractor/tests/staticmesh_castle_cellblock.rs:68`) — **VERIFIED-locally-by-agent, NOT-CI-VERIFIED**
  - Bug shape pinned: yes (StaticMeshComponent tagged-property offset 4 vs 8)
  - The agent's commit message explicitly claims: "Reverted the StaticMeshComponent offset (8 → 4) and watched the integration test fail (then restored)" — and the production code comment at `crates/navmesh-extractor/src/staticmesh.rs:280-290` includes the same audit trail.
  - The test's assertion message reads "or the StaticMeshComponent tagged-property offset has changed (expected 8 bytes past serial start)".
  - **Coverage gap (CI):** the test calls `skip_if_missing(&dir, ...)` and returns early when the cooked Castle_CellBlock asset bundle isn't co-located. On a machine without `sgw/Stargate Worlds-QA/...`, the test reports `ok` to cargo — silently passing without exercising the offset. I confirmed this on my audit machine: all 3 tests in this file reported `ok` despite none of the assertion bodies running.
  - **Suggested fix:** add an asset-free unit test (e.g. in `crates/upk-objects/src/static_mesh.rs` tests module) that synthesizes a 16-byte buffer with a known tagged property at offset 8 and asserts that `parse_tagged_properties(&buf, 8, ...)` returns ≥1 property while offset 4 returns 0. That would pin the constant in CI without needing the asset bundle.

### Unit test layer — strong static read

The 14 tests in `crates/upk-objects/src/static_mesh.rs:578-...` all ran cleanly. Sampled assertions are exact (not `is_ok()`):

- `parse_bounds_roundtrip` — pins origin, extent, sphere_radius byte-for-byte
- `unpack_normal_center` / `_positive` / `_negative` — pins the 127/255/0 → 0/+1/-1 quantization
- `read_small_kdop_tree` — pins exact triangle indices (0,1,2 material 7) at known offsets
- `collision_triangles_prefers_kdop_when_present` — uses **distinct** kDOP triangle (0,2,3) vs LOD0 (0,1,3) so a regression that picks LOD0 instead of kDOP fires loudly

The two static-read findings I flagged in advance — assertions tightly bounded, no `len() > 0` tautologies — are not present in this PR.

### Theatre findings

None. The agent's own docstring on `castle_cellblock_walks_static_mesh_actors` is honest about asset-bundle drift ("asset bundles drift slightly between locales") and pins `max_chunk_actors >= 100` as the lower bound rather than the dense-chunk 469 — a pragmatic choice.

### Coverage gaps

1. **CI cannot run the canonical StaticMeshComponent offset regression guard** (asset-dependent). See suggested asset-free unit test above.
2. No regression test for `extract_chunk` returning `actors_unresolved == actors_total` when called without a `PackageIndex`. The test `castle_cellblock_extract_chunk_without_index_emits_no_geometry` pins this but also self-skips on missing assets.
3. No fuzz target on `parse_tagged_properties` or `read_kdop_tree`. The UPK parser is reading attacker-controlled-ish data (the .umap files); the same logic that motivated PR #432's fuzz suggestion applies here.

### Verdict

**PR #436: READY TO MERGE** — the unit-test layer is clean, the agent has audit-trailed the offset choice in both production code and test message, and the agent's local revert+run verification is documented in the commit message. The coverage gap (asset-dependent test silently passing in CI) is a follow-up, not a merge blocker — but worth filing as an issue.

Cadacious:
Worked through the Codecov + Clara + Copilot feedback round on this branch. Five commits pushed (`dd19946e..73743033`):

### Addressed

| Finding | Resolution | Commit |
|---|---|---|
| **Clara G3** — `read_kdop_tree` allocation-bomb | The bounds check was already in place from the PR; promoted the magic numbers to named `MAX_KDOP_NODES` / `MAX_KDOP_TRIANGLES` constants with inline docs on the threat model. Regression test `read_kdop_tree_rejects_oversized_tri_count` fails the parser on `tri_count = i32::MAX` before any allocation. | `dd19946e`, `73743033` |
| **Clara G4** — u16 kDOP index overflow | Added `validate_kdop_indices` to `deserialize_static_mesh` that rejects any kDOP triangle whose vertex index lies outside the LOD0 vertex range. Fails-fast at parse rather than relying on the existing silent-drop behaviour in `collision_triangles()`. Three regression tests pin the new error path. | `f3b6151f`, `73743033` |
| **Copilot @ `staticmesh.rs:162`** — per-actor `Vec` allocation | `extract_chunk` now inlines the transform and pushes per triangle, dropping the `transform_triangles(...).collect()` allocation. Measured on the `extract_map_castle_cellblock` integration test (release build, 3 runs): **median 2.80s → 1.89s**; real delta is larger since most of the test time is package-index load and file I/O rather than the hot loop. | `ec0be054` |
| **Copilot @ `lib.rs:168`** — comment contradicts code | Comment updated to match the actual skip behaviour (directory listing already surfaces missing chunks; zero-content stub OBJs would just litter the output dir). | `d02adaf3` |
| **Copilot @ `staticmesh_castle_cellblock.rs:84`** — test enumeration | Test now uses `umap::enumerate_chunks` so it sees exactly what the orchestrator does and skips the non-chunk master `.umap`. | `d02adaf3` |
| **Copilot @ `extract_map_castle_cellblock.rs:163`** — `read_to_string` for header | Switched to `BufReader::read_line` so the combined-OBJ header check touches one line instead of loading tens of MB. | `d02adaf3` |

Both new regression tests verified by revert+run (testing-validation-engineer audit rule): with the fix removed, 5 of the 7 new tests fail (allocation-bomb panics on overflow, validate_kdop_indices_* assertions fire); the 2 happy-path tests correctly stay green either way. Restoring the fix returns the suite to 21/21 green.

### Skipped (with rationale)

- **Clara G1** (OBJ axis convention unverified) — known, deferred to Phase 2 per the PR body; the cube round-trip belongs in the NavBuilder-integration PR, not here.
- **Clara G2** (archetype-based actors ~20% skipped) — known and documented; needs cross-package prefab template resolution. Separate Phase 1.2-extension.
- **Clara G5** (`Vec<[[f32;3];3]>` allocation overhead) — would be an iterator-return refactor of `collision_triangles()`. The perf fix in `ec0be054` removes the *bigger* allocation (per-actor `Vec`); the per-mesh allocation is unique per resolved mesh, not per actor.
- **Clara G6** (warn-log on collision_triangles drop) — surfaces count via warn log per `negative-logging-convention`. Surfacing to the caller is over-engineering for a build-time tool, and the new parse-time `validate_kdop_indices` covers the realistic failure mode.
- **Clara G7** (agent memory committed) — project convention (`feedback_agent_memory_in_repo.md`) says this directory is committed by design. Acceptable.
- **Clara G8** (`position_offset()` unused but public) — keeping the public surface stable; the doc rationale is already inline.
- **Clara G9, G11–G13** — the actor walker, transform module, and cross-package resolution all have unit + integration coverage. G11 in particular is covered by the `transform::tests` module (yaw/pitch/roll/scale/compose).
- **Clara G14+G15** (docstring belt-and-suspenders) — the code-level fixes in this round (typed error message that explicitly calls out "u16 index wrap on a >65535-vertex mesh") speak for themselves; can be filed as a follow-up if reviewers still want explicit struct-doc text.
- **Codecov 245 missing lines** — most of the uncovered code is parsing/transform helpers exercised only through the asset-bundle integration tests, which skip in CI. The honest answer is "asset-skipping CI gap" — testing-validation-engineer flagged it as a separate issue.

Pre-PR checklist (locally):
- `cargo fmt --check` ✓
- `cargo clippy -p cimmeria-navmesh-extractor -p cimmeria-upk-objects --all-targets -- -D warnings` ✓
- `cargo test -p cimmeria-navmesh-extractor -p cimmeria-upk-objects` ✓ (49 + 21 + 5 integration = 75 passing)

Issue #63, opened 2026-03-29
Movement validation: speed checks, navmesh containment, anti-cheat


Cadacious:
## Summary

Player movement is trusted — the server accepts any position the client sends. No speed validation, no navmesh containment check, no teleport detection.

## Current State in Rust

- `crates/entity/src/movement.rs:59`: TODO for speed validation, navmesh containment, anti-cheat.
- `PlayerMovementController`: Applies client updates directly.
- Issue #35 covers world grid position tracking and navmesh (prerequisite).

## Reference Implementation

- `src/cellapp/entity/movement.cpp`: `PlayerController::playerUpdate()` validates position within space bounds, tracks last update time, applies velocity.
- `src/cellapp/space.cpp`: `isValidPosition()` validates within world bounds.
- `src/cellapp/entity/navigation.cpp`: Navmesh loading and query.

## Acceptance Criteria

- [ ] Speed check: reject updates where distance/time exceeds max_speed × tolerance
- [ ] Bounds check: reject positions outside space boundaries
- [ ] Navmesh containment: reject positions not on walkable navmesh (requires navmesh from #35)
- [ ] Teleport detection: flag/reject position jumps beyond threshold
- [ ] Graceful correction: snap player to last valid position instead of disconnecting
- [ ] Tests cover: speed violation, out-of-bounds, valid movement

## Related

- Issue #35: World grid position tracking and navmesh (prerequisite for navmesh checks)
- Issue #46: Navmesh extraction from UE3 maps

Cadacious:
## Test Plan

### Unit Tests
- [ ] Speed check: Movement delta of 100m in 1s with max_speed=5 → rejected
- [ ] Speed check: Movement delta of 4m in 1s with max_speed=5 → accepted
- [ ] Bounds check: Position (99999, 0, 99999) outside space bounds → rejected
- [ ] Navmesh check: Position not on walkable mesh → rejected, snapped to nearest valid point
- [ ] Teleport detection: Position jump of 50m in single frame → flagged

### In-Game Validation
- [ ] **Normal movement**: Walk/run around the map → movement is smooth, no rubber-banding, no server corrections
- [ ] **Navmesh boundary**: Walk toward map edge / into walls → server prevents going out of bounds (character slides or stops)
- [ ] **Vertical validation**: Attempt to walk off a ledge → server handles height correctly (falls to ground level or prevents based on navmesh)
[...]

PR #437, opened 2026-05-31
feat(movement): bounds-check + snap-back validation (#63 PR1 of 4)

Cadacious:
## Summary

First of 4 PRs implementing server-side movement validation for issue #63. Adds the **bounds-check layer**: client-authoritative position updates are now validated against the active space's AABB before the cell entity's position is advanced. On reject, the cell entity is left at its last-valid position and `BASEMSG_FORCED_POSITION` is queued to snap the offending client back — no disconnect.

This layer is intentionally shipped alone because it has **no calibration risk** (every other layer needs lag-bucketed tolerance tuning) and **no dependency** on issue #46 (navmesh data quality for non-Castle zones). The snap-back path reuses the existing `compose_forced_position_body` / `handle_teleport_player` infrastructure — zero new wire-format risk.

## What landed

- **`cimmeria-entity::movement_validation`** — new module with `MovementValidator`, `SpaceBounds`, and `position_within_bounds`. All three axes checked (X, Y, **and Z** — Z omission is the documented floor-clip exploit).
- **`SpaceManager::apply_client_position_update`** — the cell-side seam called from the inbound `EntityMove` handler. Returns `ClientMoveOutcome::{Accepted, Rejected, EntityMissing}`. The legacy `update_entity_position` stays — server-authoritative paths (ring transport, respawn, content teleport, NPC movement) call it directly and bypass validation.
- **`BaseToCellMsg::EntityMove` arm** — on reject:
  1. `cell_entity.position` is NOT advanced; the next AoI tick (100 ms) naturally rebroadcasts the last-valid position to witnesses.
  2. `CellToBaseMsg::TeleportPlayer { position == prev_pos == last_valid }` is queued — the existing teleport handler emits `BASEMSG_FORCED_POSITION` to snap the owner.
  3. `warn!` on `target: \"movement.validation\"` with `reason=\"bounds\"`, `entity_id`, `client_*`, `last_valid_*`, `bounds_*` per the [negative-logging convention](docs/architecture/negative-logging-convention.md).
- **Bounds source** — `space.navmesh.bmin/bmax` if loaded, else `SpaceBounds::FALLBACK = ([-10_000, -2_000, -10_000], [10_000, 10_000, 10_000])`. Spaces without a loaded navmesh use the fallback so non-Castle zones don't get a snap-fest.

## Design doc

The full 4-PR plan, Ghidra anchors, and tolerance-calibration runbook are in the issue body of #63. The most-flagged failure mode — *"authorized teleport, then validator snaps player back to source"* — is the canonical regression-guard shape every PR in this series ships. PR1 pins it for the bounds-layer's piece; PR3 will extend per-row across the authorized-teleport paths.

## Coming in PR2/3/4

This PR is **bounds only**. Per the design doc's intentional split:

- **PR2 — Speed check** (game-tick delta, never wall-clock). Per-world `runSpeed` from the `worlds` table; default tolerance `1.5×`. Ships **warn-only** first — production logs become the calibration dataset before enforcement flips on. Also reconciles the pre-existing `build_world_params_args` `runSpeed = 6.0` hardcode (DB says 8.125).
- **PR3 — Teleport detection + authorized-teleport allowlist**. `MovementValidator::note_authorized_teleport` API + wiring into every path in `.claude/agent-memory/movement-teleport-advisor/authorized-teleport-paths.md`. **Load-bearing PR for the canonical regression guard** — one test per row in that file.
- **PR4 — Navmesh containment**. `NavMesh::find_nearest_poly` + `AGENT_CLIMB_TOLERANCE` Z-floor check. Gated behind `MovementValidationConfig::enable_navmesh_containment` (default `false`); flips on once #46 lands clean navmesh data for non-Castle zones AND a focused Castle Cellblock QA pass confirms no false-positives on the known-bad ramps.

### Regression guards (each fails when its fix is reverted)

- `bounds_violation_outside_{x_min,y_max,z_min}_rejects_and_snaps_to_last_valid` — one per axis. Z is the floor-clip exploit guard.
- `entity_move_out_of_bounds_rejects_and_emits_teleport_player_snap_back` — end-to-end: cell entity unchanged + TeleportPlayer queued + canonical negative-log shape captured via `LogCapture`.
- `snap_back_message_routes_through_compose_forced_position_body` — wire-format byte-exact: 50-byte `BASEMSG_FORCED_POSITION (0x31)` body with last-valid in both position and prev-position slots (zero-distance interpolation, no client jitter).
- `snap_back_triggers_aoi_refresh_to_last_valid_position` — AoI fan-out contract: witnesses receive `EntityMoved` carrying last-valid (NOT the rejected attacker pos).
- **`test_authorized_teleport_does_not_trigger_bounds_anomaly`** — the **canonical failure-mode regression guard** the movement-teleport-advisor description lists explicitly. Authorized server-side teleport followed by a small-delta client update must be accepted, not snapped back.

### Happy-path + edge-case unit tests

- `legitimate_movement_within_bounds_accepts_and_writes`
- `position_on_{min,max}_corner_is_within_inclusive_bounds` — exact corners accepted (off-by-one float drift at level seams must not snap)
- `nan_position_rejected_on_every_axis` — NaN coordinates rejected
- `fallback_bounds_{accept_typical,reject_absurd_overflow}_position` — fallback AABB sanity

## Verification

```
cargo test -p cimmeria-entity --lib          # 197 passed
cargo test -p cimmeria-services --lib        # 1203 passed
cargo fmt --all -- --check                   # clean
cargo clippy -p cimmeria-entity -p cimmeria-services --all-targets -- -D warnings  # clean
cargo check --workspace --exclude cimmeria-app --exclude cimmeria-content-editor \
  --exclude cimmeria-scene-editor --exclude sgw-launcher  # clean
```

Each regression guard was verified by stashing the EntityMove arm fix and running the guards — each panics with a bug-shape message rather than a generic failure.

## Files touched

- `crates/entity/src/lib.rs` — expose `movement_validation` module
- `crates/entity/src/movement.rs` — `apply_validated_client_update` controller path + inline unit tests
- `crates/entity/src/movement_validation.rs` — **new** primitives module (147 LOC + 113 LOC tests)
- `crates/services/src/cell/space_manager/mod.rs` — `MovementValidator` field on `SpaceManager`, `pub use entities::ClientMoveOutcome`
- `crates/services/src/cell/space_manager/entities.rs` — `ClientMoveOutcome` enum + `apply_client_position_update` method
- `crates/services/src/cell/service/base_messages/mod.rs` — `EntityMove` arm wired through validator
- `crates/services/src/cell/service/base_messages/tests/general.rs` — handler-level tests (end-to-end + wire-format + AoI fan-out + authorized-teleport guard)
- `crates/services/src/cell/space_manager/tests/movement_validation.rs` — **new** cell-seam tests
- `crates/services/src/cell/space_manager/tests/mod.rs` — register new test mod
- `.claude/agent-memory/movement-teleport-advisor/pr1-bounds-seam.md` — the seam pattern PR2/3/4 should follow

Refs #63

### DIFF
--- a/.claude/agent-memory/movement-teleport-advisor/MEMORY.md
+++ b/.claude/agent-memory/movement-teleport-advisor/MEMORY.md
@@ -1,0 +1,3 @@
+- [movement-validation-anchors.md](movement-validation-anchors.md) — Ghidra anchors + DB sources for speed/bounds validation
+- [authorized-teleport-paths.md](authorized-teleport-paths.md) — Every path that legitimately mutates entity position; each must update last_pos
+- [pr1-bounds-seam.md](pr1-bounds-seam.md) — Where PR1 landed in the cell seam; the pattern PR2/3/4 should follow
--- a/.claude/agent-memory/movement-teleport-advisor/authorized-teleport-paths.md
+++ b/.claude/agent-memory/movement-teleport-advisor/authorized-teleport-paths.md
@@ -1,0 +1,28 @@
+---
+name: authorized-teleport-paths
+description: Every server-side path that legitimately writes entity position — each must update last_pos to avoid the canonical false-positive
+metadata:
+  type: reference
+---
+
+The canonical movement-validator failure mode is: an authoritative path moves a player by 500 units, the validator's `last_pos` still points at the source, the next inbound client position triggers the speed/teleport check, and the player gets snapped back to source. Every path below must call `validator.note_authorized_teleport(entity_id)` before its position write (or rely on a designed seam that does so).
+
+Verified to exist as of 2026-05-27:
+
+| Path | File | Trigger |
+|---|---|---|
+| Same-world teleport | `crates/services/src/base/world_entry/teleport.rs::handle_teleport_player` | Server-issued teleport (mission warp, GM tools) |
+| Cross-world gate travel | `crates/services/src/base/world_entry/gate_travel/` | Stargate transition; arrival is fresh spawn so `last_pos` initializes from scratch |
+| Ring transport arrival | `crates/services/src/cell/ring_transport/transporter/mod.rs` | Ring-platform pad-to-pad teleport |
+| Respawn after death | `crates/services/src/cell/cell_methods/player/combat/respawn.rs` | Death → respawner point snap |
+| World entry / play character | `crates/services/src/base/world_entry/play_character.rs` | Initial spawn into world |
+| Reanchor (recovery) | `crates/services/src/base/world_entry/reanchor_player.rs` | Desync recovery snap |
+
+Future paths to add:
+- GM `/teleport` command (when admin tools land)
+- `/stuck` self-rescue command (when player tools land)
+- Mount/dismount position adjustment (when vehicles land)
+
+Pattern verification: the canonical `handle_teleport_player` already uses `compose_forced_position_body` + `onPlayerTeleport` in one bundle; PR3 wires `note_authorized_teleport` *before* the bundle sends so even if the client's position is mid-flight, the post-bundle client update will be re-seeded rather than rejected.
+
+See [[movement-validation-anchors]] for the speed-tolerance and Ghidra anchors.
--- a/.claude/agent-memory/movement-teleport-advisor/movement-validation-anchors.md
+++ b/.claude/agent-memory/movement-teleport-advisor/movement-validation-anchors.md
@@ -1,0 +1,32 @@
+---
+name: movement-validation-anchors
+description: Ghidra + DB anchors for the movement-validation tolerance, speed source, and teleport-snap threshold
+metadata:
+  type: reference
+---
+
+Reference points for movement-validation design (issue #63):
+
+**Client teleport-snap threshold (upper bound on what client smooths).** `USGWAvatarFilter::Input` at `0x00e81970` in SGW.exe contains the load-bearing decision:
+```c
+if (_DAT_01e69c90 < SQRT(dz*dz + dx*dx + dy*dy) * (1.0 / dt)) { hard_snap(); }
+```
+`_DAT_01e69c90 @ 0x01e69c90` = `0x451C4000 f32` = **2500.0 units/second**. Anything above this ratio is hard-snapped by the client without interpolation. This is the **upper ceiling, not the cheat threshold** — server-side anti-cheat tolerance sits 3 orders of magnitude below this (`top_speed × 1.5 ≈ 12 u/s`).
+
+**Frame buffer depth.** `DAT_01e69c8c @ 0x01e69c8c` = `8` (ring buffer of 56-byte frames, minimum asserted 5 via "FrameCount > 4"). Tells us client carries ~800 ms history at 10 Hz — `TELEPORT_GRACE = 2.0 s` comfortably exceeds this.
+
+**Top-speed data source.** Per-world from `db/resources/Worlds/Seed/worlds.sql` `run_speed` column. Every populated world uses `8.125 u/s`. The `worlds` table also carries `walk_speed`, `swim_speed`, `crouch_run_speed`, `jump_speed`, etc. (see schema in `worlds.sql`). Python source reads them as `WorldInfo.runSpeed` in `deprecated/python/common/defs/WorldInfo.py:18`.
+
+**Drift bug — pre-existing.** `crates/services/src/mercury/world_data/mod.rs:112` hardcodes `runSpeed = 6.0` in `build_world_params_args`, while DB says 8.125. Server tells client one number, validator must use the same — PR2 of #63 reconciles by sourcing both from `WorldInfo`.
+
+**Server tick rate.** 100 ms / 10 Hz per `deprecated/cpp-config/config/BaseService.config` `<tick_rate>100</tick_rate>`. Use this as the implicit cadence assumption when calibrating tolerance.
+
+**Inbound client position seam.**
+- Wire: `0x03 AVATAR_UPDATE_EXPLICIT` (40 bytes), parsed at `crates/services/src/base/connect_loop/encrypted/mod.rs:214`.
+- Forwarded as `BaseToCellMsg::EntityMove`, handled in `crates/services/src/cell/service/base_messages/mod.rs:138`.
+- Final write at `crates/services/src/cell/space_manager/entities.rs::update_entity_position:147`.
+- **Gap**: `0x02 AVATAR_UPDATE_IMPLICIT`, `0x04 WARD_IMPLICIT`, `0x05 WARD_EXPLICIT` are length-parsed but never dispatched. Separate issue; same validator will apply.
+
+**Authorized teleport bundle (canonical pattern).** `crates/services/src/base/world_entry/teleport.rs::build_teleport_bundle:151` composes `FORCED_POSITION (0x31) + onPlayerTeleport (method 116)` into one Mercury bundle. `onPlayerTeleport` is a streaming-load hint only; `FORCED_POSITION` is the authoritative snap. See [[authorized-teleport-paths]] for the full list of paths.
+
+**Tolerance calibration methodology.** If derivation from RE alone is insufficient (it is — we have upper bound 2500 u/s and lower bound 8.125 u/s, gap is policy), ship the speed check **warn-only** first, collect SigNoz rejection logs with full `(distance, dt)` fields, compute p99.9 of legitimate `(distance / dt) / top_speed` ratios bucketed by RTT (50/200/500 ms), set the production tolerance from that distribution. Do not enforce snap-back until calibration data exists.
--- a/.claude/agent-memory/movement-teleport-advisor/pr1-bounds-seam.md
+++ b/.claude/agent-memory/movement-teleport-advisor/pr1-bounds-seam.md
@@ -1,0 +1,42 @@
+---
+name: pr1-bounds-seam
+description: Where the bounds validator landed in the cell seam, and the seam pattern PR2/3/4 should follow
+metadata:
+  type: project
+---
+
+PR1 of issue #63 (bounds-check layer) landed on branch `feat/movement-validation-bounds-63-pr1`. The seam pattern is the load-bearing decision for PR2/3/4 — record it so the speed / teleport-detection / navmesh-containment work doesn't have to re-derive it.
+
+**Why:** the controller-level `PlayerMovementController::apply_validated_client_update` isn't yet wired into the production cell path. The live seam for client position updates is `BaseToCellMsg::EntityMove → SpaceManager::apply_client_position_update`. Validation must happen here, not in the controller.
+
+**How to apply:**
+
+- Validator lives on `SpaceManager` as `movement_validator: MovementValidator` (stateless for PR1; PR2/3 add per-entity state to the same field).
+- New entry point on `SpaceManager`: `apply_client_position_update(entity_id, position, direction, velocity) -> ClientMoveOutcome`. Returns `Accepted` / `Rejected { reason, last_valid, space_id, bounds }` / `EntityMissing`. The unchecked `update_entity_position` stays — server-authoritative callers (ring transport, respawn, content teleport, NPC movement) use it directly and bypass validation.
+- The `EntityMove` arm in `cell/service/base_messages/mod.rs` is the single consumer. On `Rejected`, it emits `CellToBaseMsg::TeleportPlayer { entity_id, space_id, position: last_valid, prev_pos: last_valid }` which routes through the existing `handle_teleport_player` → `compose_forced_position_body` path — no new wire-format risk.
+- Bounds source: `space.navmesh.bmin/bmax` if loaded, else `SpaceBounds::FALLBACK = ([-10_000, -2_000, -10_000], [10_000, 10_000, 10_000])`. Spaces without a loaded navmesh use the fallback so non-Castle zones don't snap-fest.
+
+**AoI refresh:** by NOT writing the rejected position, the AoI tick (100 ms) naturally rebroadcasts the unchanged last-valid position to witnesses. No explicit AoI fan-out needed. The owner gets the immediate snap via `BASEMSG_FORCED_POSITION` through `TeleportPlayer`.
+
+**Negative-log shape pinned in tests:**
+- `target: "movement.validation"`, level `warn!`, message starts `"movement.bounds_violation:"`.
+- Fields: `reason="bounds"`, `entity_id`, `space_id`, `client_{x,y,z}`, `last_valid_{x,y,z}`, `bounds_{min,max}_{x,y,z}`, `reject=?MovementReject` (Debug-formatted).
+- LogCapture pinned in `cell::service::base_messages::tests::general::entity_move_out_of_bounds_rejects_and_emits_teleport_player_snap_back`.
+
+**Test seams worth knowing:**
+- `Castle_CellBlock` in spaces.xml is `Instanced="true"` — every `create_entity` builds a fresh space, so co-spatial tests (witness + offender) must use a non-instanced world (`Agnos` works). Tripped over this twice; record so it doesn't trip a third time.
+- `external/` is not symlinked into worktrees by `git worktree add`; native MSVC entity tests need a `New-Item -ItemType Junction` to the main repo's `external/` before `cargo check -p cimmeria-entity` compiles.
+
+**Wire-format byte layout** for `BASEMSG_FORCED_POSITION (0x31)` is 50 bytes — pinned by `snap_back_message_routes_through_compose_forced_position_body`:
+- `[0]` = 0x31
+- `[1..5]` = entity_id u32 LE
+- `[5..9]` = space_id u32 LE
+- `[9..13]` = vehicle_id u32 LE (= 0)
+- `[13..25]` = position xyz f32 LE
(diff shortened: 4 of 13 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | Fallback AABB is massive | **`SpaceBounds::FALLBACK = ([-10_000, -2_000, -10_000], [10_000, 10_000, 10_000])`.** That is a 20km × 12km × 20km box. A cheater could teleport anywhere within that box and the bounds check would pass. The fallback is intentionally permissive to avoid false positives in non-Castle zones, but it is also permissive enough to allow significant exploits. |
| **G2** | No rate limiting on snap-backs | **If a cheater sends 100 out-of-bounds positions per second, the server sends 100 `BASEMSG_FORCED_POSITION` responses.** This could be used for amplification attacks or to spam the client. Consider rate-limiting snap-backs per entity. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | NaN check is per-axis | **The test `nan_position_rejected_on_every_axis` suggests NaN is checked per-axis.** But what about `Infinity`? A position of `[f32::INFINITY, 0.0, 0.0]` would pass the bounds check (infinity > -10000) but cause downstream issues. Consider checking `is_finite()` instead of just NaN. |
| **G4** | Bounds are inclusive | **The test `position_on_{min,max}_corner_is_within_inclusive_bounds` verifies exact corners are accepted.** But floating-point equality is tricky. A position of `[10000.0001, 0.0, 0.0]` would be rejected even though it is essentially on the boundary. Consider adding a small epsilon tolerance. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G5** | Agent memory files committed | The PR adds 3 files to `.claude/agent-memory/movement-teleport-advisor/`. Same concern as PR #436 G7 — should agent memory files be committed to the repo? |
| **G6** | `authorized-teleport-paths.md` lists 6 paths | The file lists 6 paths that must call `note_authorized_teleport`. But PR1 does not implement `note_authorized_teleport` — that is PR3. The test `test_authorized_teleport_does_not_trigger_bounds_anomaly` verifies the bounds layer does not need it (it checks current space AABB, not per-entity last_valid), but the file could be misleading. |
| **G7** | Fallback AABB Y-min is -2000 | The fallback AABB has `min[1] = -2000` but `max[1] = 10000`. This asymmetry suggests the Y-axis is vertical (height). But the test `bounds_violation_outside_y_max_rejects` uses `Y = 999_999` — is Y the vertical axis or a horizontal axis? Document the axis convention. |

MyPalClara:

| **G8** | No test for `Infinity` position | The NaN test exists, but no test for `f32::INFINITY` or `f32::NEG_INFINITY`. |
| **G9** | No test for rapid-fire rejections | What happens if a client sends 100 out-of-bounds positions in 1 second? Does the server send 100 snap-backs? |
| **G10** | No test for `last_valid` update after acceptance | The tests verify `last_valid == SPAWN_POS` after rejection, but do not verify `last_valid` is updated after acceptance. |
| **G11** | No test for navmesh-sourced bounds | The tests use the fallback AABB because "Castle Cellblock has no navmesh in the test harness." There is no test verifying the navmesh-sourced bounds path works. |

MyPalClara:

| **G12** | `SpaceBounds::FALLBACK` rationale not in code | The PR body explains why the fallback is permissive, but this rationale is not in the code comments. A future contributor might tighten the fallback without understanding the tradeoff. |
| **G13** | No doc on the 4-PR plan in the code | The PR body explains the 4-PR plan, but the code does not reference it. A future contributor reading `movement_validation.rs` would not know PR2/3/4 are coming. |
| **G14** | `authorized-teleport-paths.md` references PR3 | The file says "PR3 wires `note_authorized_teleport`" but PR3 does not exist yet. If PR3 is delayed or cancelled, this file becomes misleading. |

---

## Summary

| Category | Count |
|----------|-------|
| Critical issues | 2 |
| Correctness issues | 2 |
| Potential issues | 3 |
| Missing test coverage | 4 |
| Documentation | 3 |

**Total: 14 gaps**

Cadacious:

PR #437 is the first layer of a four-PR validator stack (bounds today; speed / teleport-detection / navmesh-containment in PR2/3/4). The single trust boundary that matters here is `BaseToCellMsg::EntityMove`'s client-supplied `position`, and the bounds layer is **server-authoritative** by construction — the validator looks at the active space's AABB, not at any client-supplied state. Verdict at the bottom.

### Handler: `cell/service/base_messages/mod.rs` — EntityMove arm

#### Client-asserted fields
- `position: [f32; 3]` — the client's claimed avatar pos.
- `direction: [i8; 3]` — yaw/pitch/roll bytes.
- `velocity: [f32; 3]` — client's claimed velocity.

#### Validation audit
- `position` → `SpaceManager::apply_client_position_update` → `MovementValidator::check_bounds` against the space's navmesh `bmin/bmax` or `SpaceBounds::FALLBACK`. **OK.** All three axes checked inclusive. NaN naturally rejected (every `>=`/`<=` against NaN is false).
- `direction` → not validated. Direction has no security-relevant range (any byte value renders); not a violation.
- `velocity` → not validated. PR2 will use this; for PR1, the validator does not derive elapsed time from it. **OK.**

#### Exploit analysis
- **Speed hack via inflated dt:** N/A in PR1 (no time base used).
- **Teleport / out-of-bounds position:** mitigated by bounds check + snap-back via `BASEMSG_FORCED_POSITION`.
- **Floor-clip under terrain:** mitigated by Z-axis inclusion in the bounds AABB. Both `z_min` and `z_max` are guarded — verified `position_outside_z_min_rejected` and `position_outside_z_max_rejected` exist in `movement_validation.rs` tests, and the cell-seam `bounds_violation_outside_z_min_rejects_and_keeps_last_valid` exists. PR4 (navmesh containment) will add the polygon-height check that catches "Z within AABB but below the actual walkable surface" — the bounds layer doesn't see that, but that's by design.
- **Snap-back uses rejected position:** mitigated. `apply_client_position_update` reads `last_valid` from the entity's current position **before** any write, then passes it as both `position` and `prev_pos` in `TeleportPlayer` — zero-distance client-side interpolation. Pinned by `snap_back_message_routes_through_compose_forced_position_body`.
- **AoI leak of rejected position to witnesses:** mitigated. The cell entity is **not** advanced on reject, so the next AoI tick fans out the unchanged last-valid position. Pinned by `snap_back_triggers_aoi_refresh_to_last_valid_position`.
- **False-positive on authorized teleport (ring transport / respawn / content teleport):** in the **bounds layer only**, this is structurally impossible — the validator is stateless and checks against the current space's AABB. As long as the teleport destination is inside the same space's AABB, the next client position near it accepts. The test `test_authorized_teleport_does_not_trigger_bounds_anomaly` is correctly framed by the author as "smoke today, will become a real regression guard in PR3 when last_pos / authorized-teleport allowlist arrives."

### Verdict: SHIP

No CRITICAL / HIGH issues. Two advisory notes:

1. **Advisory — `note_authorized_teleport` is named in the agent memory but not yet wired.** The memory file `authorized-teleport-paths.md` lists six paths that "must call `validator.note_authorized_teleport(entity_id)` before its position write." That API doesn't exist in PR1 (the validator is stateless). When PR3 lands the authorized-teleport allowlist, every path in that table needs the call wired in the same PR — otherwise PR3 will introduce the false-positive class PR1 happens to avoid by being stateless. Track this; not a PR1 finding.

2. **Advisory — fallback AABB Y_min = -2000 is arguably tight for some worlds.** The `[-10000, -2000, -10000]` fallback rejects any legit Y below -2000. If any zone has a legitimate underground level (canyon, ravine, cave) below Y=-2000 and ships **without a loaded navmesh** (so the fallback is in force), legitimate players there get snapped. Castle Cellblock is fine; other worlds untested. Recommend extending the Y_min to match X/Z (`-10000`) unless there's a derived reason for -2000. The Y-asymmetry is the only thing in `SpaceBounds::FALLBACK` I'd second-guess. Not a security issue.

The seam pattern (validator on SpaceManager; `apply_client_position_update` as the only client-position entry point; `update_entity_position` retained for server-authoritative paths) is exactly the right shape for PR2/3/4 to extend without re-litigating the trust boundary. The negative-log shape (`reason="bounds"`, `client_*`, `last_valid_*`, `bounds_*`) is structured for SigNoz aggregation. The snap-back path reuses `compose_forced_position_body` rather than introducing a new wire primitive — good.

Self-check: every client-supplied field traced; no client timestamp used as time base (none in PR1); no type_id-keyed mutation; dedup-hash + auth-token check happens at the framing layer (`encrypted::handle_avatar_update_explicit` upstream).

🤖 server-authority-enforcer

Cadacious:

- Test: `bounds_violation_outside_z_min_rejects_and_snaps_to_last_valid` — **VERIFIED**
  - Bug shape pinned: yes (floor-clip exploit via Z=-100_000)
  - Verification: reverted the `check_bounds` call in `apply_client_position_update` (`crates/services/src/cell/space_manager/entities.rs:233`) to always accept; test failed with "Z-min bound must reject (floor-clip exploit guard)".
  - Risk: none.

- Test: `bounds_violation_outside_x_min_rejects_and_snaps_to_last_valid` — **VERIFIED**
  - Bug shape pinned: yes (X=-100_000 attacker pos)
  - Same revert; failed with `expected Rejected(OutOfBounds), got Accepted { position: [-100000.0, 0.0, 20.0] }` — the attacker pos went straight through.
  - Risk: none.

- Test: `bounds_violation_outside_y_max_rejects_and_snaps_to_last_valid` — **VERIFIED**
  - Bug shape pinned: yes (Y=999_999 ceiling)
  - Same revert; failed with "Y-max bound must reject with OutOfBounds and snap to spawn".
  - Risk: none.

- Test: `snap_back_message_routes_through_compose_forced_position_body` — **VERIFIED**
  - Bug shape pinned: yes (wire-format byte exactness for the `BASEMSG_FORCED_POSITION` body)
  - Verification: swapped `position: last_valid` → `position` (the rejected client pos) in the `TeleportPlayer` construction at `crates/services/src/cell/service/base_messages/mod.rs:230`; failed with `left: [1000000000.0, 0.0, 0.0], right: [7.0, 8.0, 9.0]` — the attacker pos was serialized into the snap-back body instead of the spawn pos.
  - Risk: none.

### Theatre findings

- Test: `test_authorized_teleport_does_not_trigger_bounds_anomaly` (both copies: `cell/space_manager/tests/movement_validation.rs:186` cell-seam, AND `cell/service/base_messages/tests/general.rs:450` handler-level) — **NOT-A-REGRESSION-GUARD-IN-PR1**
  - **The agent has been honest about this** in the docstring (lines 442-448 of `general.rs`): "Reverting `apply_client_position_update`'s seam to silently bypass the validator would still pass this test (it tests the accept side); reverting the post-teleport position write inside `update_entity_position` itself would push the post-teleport client coordinate >0.1 units from the cell entity's stale pos and... still accept under bounds alone. PR3 is where this guard graduates from 'smoke' to 'fire on revert of last_pos reseed'."
  - Verification: under the bounds-check revert (same revert that fires the 3 bounds-violation tests), **both copies of `test_authorized_teleport_does_not_trigger_bounds_anomaly` PASS**. Confirmed exactly as the agent's docstring predicts.
  - **Not theatre per se** — these are happy-path smoke tests that PR3 will harden into regression guards once the per-entity authorized-teleport allowlist exists. They have value as future-proofing.
  - **The user's claim** that "this is the canonical regression guard for the movement-teleport-advisor's most-flagged failure mode" does not hold for PR1 alone. The 3 bounds-violation tests are the canonical regression guards for PR1's scope; the authorized-teleport tests are placeholders for PR3.
  - **Suggested fix:** none for PR1 — keep the placeholder tests, just don't cite them as canonical regression guards until PR3 lands.

### Coverage gaps

- No test for the bounds-rejection log shape (the `tracing::warn!` at `base_messages/mod.rs:200` with `bounds_min_x`, `bounds_max_x`, etc.). The `rejection_carries_bounds_for_log_capture` test pins that the validator surfaces the AABB on the return value, but no test pins the negative-log fields per `docs/architecture/negative-logging-convention.md`. A `LogCapture` test against the bounds-rejection warn would close this gap.

### Verdict

**PR #437: READY TO MERGE** — 4 of 4 substantive regression guards verified to fire on revert with exact failure messages. The 2 authorized-teleport tests are honest about being placeholders for PR3, not regression guards for PR1.

Cadacious:
## Clara feedback — applied

Three new commits address the highest-leverage items; the rest are deferred to the appropriate PR in the 4-PR plan.

### Addressed

| Item | Commit | What |
|---|---|---|
| **G3** — NaN check is per-axis (infinity slips through wide AABBs) | `24bcf766` | Replaced the implicit "NaN fails every comparison" rejection in `position_within_bounds` with an explicit `if !pos.x.is_finite() \|\| !pos.y.is_finite() \|\| !pos.z.is_finite() { return false; }` gate. Covers NaN, +∞, and -∞ in one check. Doc updated to call out why the explicit gate matters (a future contributor widening the fallback toward `f32::MAX` / `f32::INFINITY` would silently let infinities through). |
| **G8** — no test for Infinity / -Infinity | `78e2389a` | Added `position_with_infinity_axis_rejected` + `position_with_neg_infinity_axis_rejected` in `crates/entity/src/movement_validation.rs`. Both test against a wide-open AABB (`[-∞, -∞, -∞]..[+∞, +∞, +∞]`) so the explicit `is_finite()` gate is the only thing rejecting the position — fires on revert of `is_finite()` → `is_nan()`. **Verified by revert+run**: both tests fail with the gate downgraded to `is_nan()`-only. |
| **G10** — no test for `last_valid` update after acceptance | `78e2389a` | Added `legitimate_movement_updates_last_valid_for_next_iteration` in `crates/services/src/cell/space_manager/tests/movement_validation.rs`. Shape: accept A → accept B → reject C → assert C's `last_valid == B` (proves the accept path writes through to `cell_entity.position`). **Verified by revert+run**: removing `self.update_entity_position(...)` from the accept arm fires the test (`last_valid` reads back as spawn instead of B). |
| **G12** — `SpaceBounds::FALLBACK` rationale not in code | `c9db6c34` | Expanded the FALLBACK doc-comment with three labeled sections: **Rationale** (what it's for: legacy zones, dev maps, navmesh-less spaces), **Permissive by design** (why a 20 km box is the right tradeoff — false-positive snapbacks are worse than letting an in-bounds cheater roam, and the *next* validator layers are the actual gate), **Tightening path** (how the AABB narrows over time as navmeshes get wired and the teleport-detection layer's authorized-teleport allowlist comes online). No PR/issue numbers — rationale only. |

### Deferred — out of scope for PR1

| Item | Reason |
|---|---|
| **G1** — FALLBACK is 20 km × 12 km × 20 km, too permissive | Defensible tradeoff documented in `c9db6c34`. The fallback isn't the gate against in-bounds cheating; PR3's teleport-detection layer is. Tightening it would trade false-positive snapbacks for marginal cheater-bounding inside an already-wide blind spot. |
| **G2** + **G9** — no rate limiting on snap-backs | Server-infrastructure concern. Belongs in PR4 or a follow-up; not PR1's seam. |
| **G4** — bounds are inclusive (float-precision concern) | The inclusive bounds are intentional (off-by-one floats at level seams must not snap back). Pinned by `position_on_{min,max}_corner_is_within_inclusive_bounds`. Float-precision epsilon is a separate concern. |
| **G5** — agent memory files committed | Repo convention — `.claude/agent-memory/` is committed per `.claude/agent-memory/movement-teleport-advisor/feedback_agent_memory_in_repo.md`. |
| **G6** + **G14** — `authorized-teleport-paths.md` references PR3 functionality | By design — the file documents the contract PR3 will wire. It lives in agent memory specifically because PR3 (and the advisor invoking it) will read it. |
| **G7** — fallback Y-min asymmetry / axis convention | Picked up by the server-authority-enforcer's "Advisory 2" — recommended for PR3 or later, not a PR1 finding. Doc-only follow-up. |
| **G11** — no test for navmesh-sourced bounds | PR1's test harness doesn't load navmeshes for Castle Cellblock (the test space). Navmesh-sourced bounds get coverage when the navmesh extractor pipeline (issue #46) lands in the test harness. |
| **G13** — 4-PR plan not in code | The plan lives in the #63 issue body and the per-PR descriptions. The validator's module-level doc already names the four layers (`Bounds (this layer)`, `Speed (future)`, `Teleport detection (future)`, `Navmesh containment (future, gated)`) — sufficient for code-readers to know more layers are coming without pinning issue numbers into source. |
| **Codecov — 17 missing lines** | Mostly unreachable defensive branches (`EntityMissing` guards in the seam, the `space.entities.get(&entity_id)` post-existence-check arm). Acceptable. |

### Verification

```
cargo test -p cimmeria-entity movement_validation       # 16 pass
cargo test -p cimmeria-services --lib movement_validation  # 8 pass
cargo fmt -p cimmeria-entity -p cimmeria-services -- --check  # clean
cargo clippy -p cimmeria-entity -p cimmeria-services --all-targets -- -D warnings  # clean
```

Both new regression tests verified by revert:
- Replacing `is_finite()` with `is_nan()` → both infinity tests fail (`assertion failed: !position_within_bounds(...)`).
- Removing the post-accept `update_entity_position` call → `last_valid` test fails (`left: [10.0, 0.0, 20.0], right: [15.0, 0.0, 25.0]` — last_valid stayed at spawn instead of advancing).

PR #438, opened 2026-05-31
feat(trade): player-to-player trading system (#54)

### LINKED ISSUE #54: Player-to-player trading system
## Summary

Trading has cell method dispatch indices but no implementation. The Python reference has a complete two-player trade system.

## Current State in Rust

- Cell methods defined: TRADE_REQUEST (104), TRADE_REQUEST_CANCEL (105), TRADE_UPDATE_PROPOSAL (106), TRADE_LOCK_STATE (107).
- No implementation files found.

## Reference Implementation

- `python/cell/Trade.py` (274 lines): Complete system:
  - `TradeProposal`: Per-player trade offer (items + currency).
  - `TradeTransaction`: Manages two-player trade lifecycle.
  - `updateProposal()`: Update offer with version checking.
  - `updateLockState()`: Lock states (None → Locked → LockedAndConfirmed).
  - `confirm()`: Execute trade if both players confirmed — swap items and currency atomically.
  - `cancel()`: Abort trade, return items.
  - Space validation (both players must be in same space).
- Wire messages: `TradeState`, `TradeResults`.

## Acceptance Criteria

- [ ] Trade request: initiate trade with nearby player
- [ ] Trade proposal: add/remove items and currency to offer
- [ ] Lock/confirm flow: both players must lock then confirm
- [ ] Atomic execution: items and currency swapped only if both sides valid
- [ ] Cancel: either player can abort at any time
- [ ] Distance check: trade cancelled if players move apart
- [ ] Tests cover: full trade flow, cancel, insufficient items, distance break

Cadacious:
## Summary

Full Phase 1 implementation of player-to-player trading (closes #54). Ports the 274-line Python `cell/Trade.py` to Rust as a four-handler cell-side state machine + base-side atomic sqlx commit. Item ownership + naquadah swap happen in a single `BEGIN`/`COMMIT` block with `FOR UPDATE` row locks; no items can be lost on rollback.

Built against the Ghidra-evidenced deep dive in #54. The three explicitly-defended traps:

1. **`onTradeResults.Result` is `INT32`**, not `INT8`. `SGWPlayer.def:1387` declares the wire field `INT32` even though the underlying `ETradeResults` enum is `INT8`. Reverting to a one-byte push produces a silent client parse failure (trade UI never closes). Pinned by `on_trade_results_result_field_is_int32_not_int8`.

2. **User-initiated cancel sends `Completed (1)`**, not `Cancelled (2)`. Matches `Trade.py:225-228` — both players get a clean shutdown notification regardless of who hit cancel. `Cancelled (2)` is reserved for disconnect / distance-break / commit-failure paths. Pinned by `cancel_clears_state_on_both_sides_and_sends_completed_to_both`.

3. **QA 0.8384 client skips `tradeRequest` entirely** — it jumps straight to `tradeUpdateProposal`. `handle_trade_update_proposal` must implicitly call `begin_trading()` if no session is open, mirroring the Python workaround at `SGWPlayer.py:1785-1790`. Pinned by `trade_update_proposal_without_prior_request_begins_trading`.

All three regression tests **verified to fail when the corresponding fix is reverted** (per the testing-validation-engineer audit rule).

## Files touched

- `crates/entity/src/trade.rs` (new) — `TradeProposal`, `TradeItem`, `ETRADELOCKSTATE_*` / `ETRADERESULTS_*` constants, parse + serialize_local + serialize_remote + outbound `onTradeState`/`onTradeResults` serializers.
- `crates/entity/src/cell_entity/mod.rs` — adds `trade_partner_entity_id: Option<u32>` and `trade_proposal: Option<TradeProposal>` to `CellEntity`.
- `crates/services/src/cell/cell_methods/player/trade.rs` (new) — the four cell-method handlers (104-107), state machine, distance + version checks, `cancel_trade_on_disconnect` hook.
- `crates/services/src/cell/cell_methods/player/dispatch.rs` — adds the trade sub-range arm (104..=107) inside the existing ORG_CREATION..=CANCEL_MOVIE outer arm, with a static-assert ordering guard.
- `crates/services/src/cell/cell_methods/player/social.rs` — drops the four UNIMPLEMENTED trade stubs (now routed to `trade::dispatch` upstream).
- `crates/services/src/cell/messages/cell_to_base.rs` — new `ExecuteTrade` variant.
- `crates/services/src/cell/service/base_messages/mod.rs` — `DestroyEntity` + `DisconnectEntity` hooks now call `cancel_trade_on_disconnect` so disconnects mid-trade notify the surviving partner with Cancelled.
- `crates/services/src/base/world_entry/methods/trade/{mod.rs, execute.rs, tests.rs}` (new) — `handle_execute_trade` runs the full atomic swap: pg_advisory_xact_lock both players (ascending player_id), `SELECT FOR UPDATE` naquadah, validate cash + ownership + bound/buyback/duplicate gates, `reserve_free_inventory_slots` in INV_MAIN for each recipient, `UPDATE sgw_inventory SET character_id, container_id, slot_id` per row, `UPDATE sgw_player SET naquadah` per side, COMMIT.
- `crates/services/src/mercury/mod.rs` — adds `method_idx::ON_TRADE_STATE = 144` and `ON_TRADE_RESULTS = 145`.

## Improvements over the Python reference

- **Distance re-check on every `tradeUpdateProposal` AND `tradeLockState`.** Python only checks at `beginTrading`. A trade open while one player walks away now auto-cancels with `Cancelled (2)` rather than completing from too-far. Per the deep dive's recommendation.
- **Atomic sqlx transaction.** Python `confirm()` runs `removeItems()` then `addItems()` sequentially with no rollback. Our `handle_execute_trade` wraps everything in a single tx with `FOR UPDATE` row locks. If anything fails (DB error, missing item, insufficient cash, no slot space), nothing changes hands.
- **Explicit disconnect cleanup.** Python relied on BigWorld GC to tear down the `tradeTransaction`. Rust hooks both `DestroyEntity` and `DisconnectEntity` to cancel the trade and notify the surviving partner.
- **Validation gauntlet inside the tx.** Bound items, buyback-bag items, and duplicate-instance proposals are all rejected at commit time (TOCTOU-safe).

## Tests

**Wire-format byte-exact (cimmeria-entity, 7 tests):**
- `local_trade_proposal_serializes_byte_exact`
- `local_trade_proposal_parse_roundtrip`
- `parse_rejects_truncated_proposal`
- `parse_rejects_oversized_count_without_allocating` (OOM guard)
- `on_trade_results_result_field_is_int32_not_int8` — **TRAP #1**
- `cancel_serializes_result_as_completed_one_not_cancelled_two` — **TRAP #2**
- `on_trade_state_serializes_local_and_remote`

**Cell-side state machine (cimmeria-services, 12 tests):**
- `trade_request_opens_session_on_both_sides`
- `trade_request_too_far_rejects_with_cancelled_to_caller_only`
- `trade_update_proposal_without_prior_request_begins_trading` — **TRAP #3**
- `lock_state_progression_none_to_locked_to_confirmed`
- `cancel_clears_state_on_both_sides_and_sends_completed_to_both` — **TRAP #2**
- `distance_break_during_update_auto_cancels`
- `proposal_version_must_increment_by_one`
- `cannot_trade_with_self`
- `cannot_trade_with_npc`
- `both_confirm_triggers_execute_trade_hand_off`
- `invalid_lock_state_value_is_rejected`
- `disconnect_cancellation_clears_partner_state`

**Live-DB atomic-commit (cimmeria-services, 7 tests, self-skip without DATABASE_URL):**
- `commit_swaps_items_atomically` (happy path)
- `commit_rolls_back_on_insufficient_cash`
- `commit_rolls_back_on_missing_item`
- `commit_rolls_back_on_bound_item`
- `commit_rolls_back_on_buyback_bag_item`
- `commit_rolls_back_on_duplicate_instance_in_proposal`
- `commit_rolls_back_on_negative_cash`

All 7 live-DB tests verified passing locally against `postgres://w-testing:w-testing@localhost:5433/sgw`.

## Test plan

- [x] `cargo fmt --all -- --check` (clean)
- [x] `cargo clippy -p cimmeria-services -p cimmeria-entity --lib --tests -- -D warnings` (clean)
- [x] `cargo test -p cimmeria-entity --lib` (186/186 pass)
- [x] `cargo test -p cimmeria-services --lib` (1211/1211 pass)
- [x] `cargo nextest run --profile=ci-live-db -p cimmeria-services --lib --filterset 'test(/trade::tests::commit_/)'` (7/7 pass against local Postgres)
- [x] All 3 trap-specific regression tests manually verified to fail when their respective fix is reverted

## Out of scope (follow-ups)

- **Two-client live verification** — deep dive flagged as needing real-client follow-up; all tests in this PR are server-only.
- **Asymmetric `NoLocalCash`/`NoLocalSpace`/`NoRemoteCash`/`NoRemoteSpace` result codes** — currently the failure path sends `Cancelled (2)` to both clients; Python asymmetrically reports which side failed. The `TradeAbort` variants already carry the side information.
- **Full `RemoteTradeProposal` `InvItem` data** — the cell doesn't keep a full inventory mirror, so the partner-side `RemoteTradeProposal` is emitted with stub `InvItem` rows. Client renders correctly via its own cached inventory.
- **`canSell()` proxy at commit-time** — Python silently skips unsellable items from the proposal; we de-dupe and validate ownership but don't yet filter by `ITEM_FLAG_CAN_BE_SOLD` inside the tx. (Bound items ARE blocked.)

## Server-authority

Every client-supplied field is validated server-side:
- Item ownership re-checked inside the tx with `FOR UPDATE`.
- Cash availability re-checked inside the tx with `FOR UPDATE`.
- Bound items blocked.
- Buyback-bag items blocked.
- Duplicate instances in the same proposal blocked.
- Negative cash blocked pre-tx.
- Distance + same-space required at session open AND every state update.
- Lock-state range validated (only `0`/`1`/`2` accepted).
- Version sequence enforced (proposal must monotonically bump by 1).
- "Stale partner view" lock attempts silently downgraded to `None`.

Closes #54

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Full player-to-player trading: propose items/cash, lock/confirm, and atomic server-side execution with inventory & balance updates.
  * Real-time trade state and result notifications to clients; trades auto-cancel when partners move out of range.

* **Bug Fixes**
  * Robust cleanup on disconnect/destroy ensures surviving partner is notified and session cleared.
  * Validation rejects malformed/oversized proposals and prevents duplicate/invalid item offers.

* **Tests**
  * Extensive regression and integration tests for protocol, handoff, slot-reservation, and DB commit/rollback scenarios.


Cadacious:

One CRITICAL, one HIGH, two MEDIUM, one LOW, several advisory notes.

---

### CRITICAL — Container whitelist missing on tradeable items (equipment / mission / bank dupe-strip exploit)

**Handler:** `crates/services/src/base/world_entry/methods/trade/execute.rs::lock_items`

**Client-asserted field:** `instance_id` (i32) in each `TradeItem` of the proposal.

**What the server validates:**
- Ownership: `WHERE character_id = $1 AND item_id = $2 FOR UPDATE` — good, blocks asserting another player's items.
- `row.bound` — good, blocks soul-bound items.
- `row.container_id == INV_BUYBACK` — only INV_BUYBACK (16) is rejected.

**What is missing:** any restriction that items come from `INV_MAIN` (1). The current check is a single-container blacklist, not a tradeable-container whitelist. Every other container the player owns rows in (`INV_MISSION` (2), `INV_BANDOLIER` (3), `INV_HEAD` (4) … `INV_ARTIFACT2` (14), `INV_CRAFTING` (15), `INV_BANK` (17), `INV_AUCTION` (18), `INV_TEAM_BANK` (19), `INV_COMMAND_BANK` (20)) passes the gauntlet today.

**What an attacker would do** (concrete shapes — pick the worst):

1. **Strip equipped gear in a trade.** Open a trade, put your `INV_CHEST` (7) armor instance_id in the proposal. The handler validates ownership, runs `UPDATE sgw_inventory SET character_id = recipient, container_id = INV_MAIN, slot_id = new_slot WHERE item_id = $4`. The armor row is now in the recipient's main bag, but the cell-side cached stats on player A (which were computed when the armor was equipped) are stale — A still benefits from the armor stats locally until something forces a stat recompute. Meanwhile B has a free piece of gear. With a partner alt this is straightforward dupe-as-loss.
2. **Trade INV_MISSION items.** Quest items live in container 2. Some are intended to be account-bound but enforced by container, not by `is_bound`. Trading them turns a single-player quest into a guild-share, and trading turn-in items between alts duplicates mission rewards. (Confirm via mission grant code: if INSERT does not set `bound=true` for INV_MISSION rows, this is live.)
3. **Trade INV_BANK items.** Bank storage is gated by being at a banker NPC in canonical SGW. A player far from a banker can offer bank-bag items in a trade, bypassing the banker gate.
4. **Trade bandolier ammo.** Less catastrophic but lets a player hand off the currently-loaded ammo of an equipped weapon without firing the bandolier-sync flush.

**What breaks:** dupe-strip on equip slots, mission-progress sharing, bypass of banker-NPC gate, stale bandolier ammo state.

**Why FOR UPDATE does not save you:** the row-level lock guarantees serializability against concurrent vendor/trade paths, but it does not restrict *which* rows are eligible to move. A player offering an INV_CHEST row gets the same atomicity guarantee as a legitimate INV_MAIN trade.

**Fix:**

```rust
// In lock_items, after the BoundItemOffered check:
const TRADEABLE_FROM: &[i32] = &[INV_MAIN];
if !TRADEABLE_FROM.contains(&row.container_id) {
    return Err(TradeAbort::IneligibleContainer {
        which,
        player_id,
        item_id,
        container_id: row.container_id,
    });
}
```

Add a corresponding `TradeAbort::IneligibleContainer` variant. Add a live-DB regression test that inserts an item in `INV_CHEST`, attempts to trade it, and asserts the swap rolls back AND the item is still in `INV_CHEST` on player A.

This is a **server-authority** violation: the server is accepting the client's claim of "this item can be traded" instead of independently checking it. Per the doctrine: server must verify, not trust.

---

### HIGH — Distance check between cell snapshot and base atomic_swap

**Handler:** `crates/services/src/cell/cell_methods/player/trade.rs::request_execute_trade`

**Trace:** `tradeLockState` reaches `LockedAndConfirmed` on both sides → cell verifies `partners_in_range` was checked at update/lock time → cell clears state and sends `ExecuteTrade` to base → base does the atomic swap → base sends `Completed` to both.

**What's missing:** the distance check exists on `tradeUpdateProposal` and `tradeLockState`, but *not* immediately before the `request_execute_trade` hand-off. The window is small (one tick of cell processing) but real — and once the message is in the cell→base queue, the swap completes regardless of whether the players walked apart.

**Exploit shape:** mostly griefing, but: player A locks confirmed → walks 100 units away while the partner's confirmation is in-flight → partner locks confirmed → swap completes. Not a dupe, but it defeats the trade-window proximity invariant. Lower severity than the container exploit.

**Fix:** add one final `partners_in_range` check inside `request_execute_trade` immediately before the snapshot, fall back to `cancel_session(..., Cancelled)` if it fails.

```rust
async fn request_execute_trade(...) {
    if !partners_in_range(entity_id, partner_entity_id, space_mgr) {
        cancel_session(entity_id, partner_entity_id, ETRADERESULTS_CANCELLED, tx, space_mgr).await;
        return;
    }
    let snapshot = { ... };
    // ...
}
```

---

### MEDIUM — Deadlock risk: trade and vendor stack take advisory locks on different (key, ns) tuples

**Handler:** `trade/execute.rs::atomic_swap` uses `pg_advisory_xact_lock(player_id, 0)`.
**Vendor stack** (e.g., `vendor/serializers.rs::reserve_free_inventory_slots`) uses `pg_advisory_xact_lock(player_id, container_id)` where `container_id = INV_MAIN = 1`.

Postgres treats `(player_id, 0)` and `(player_id, 1)` as independent locks. They do not serialize against each other at the advisory layer.

Lock acquisition order:
- Trade: advisory(p1, 0) → advisory(p2, 0) → naquadah FOR UPDATE (p1, then p2) → item rows FOR UPDATE.
- Vendor sell: item rows FOR UPDATE → naquadah FOR UPDATE.
- Vendor purchase: item rows (via `consume_design_quantity`) FOR UPDATE → naquadah FOR UPDATE.

If trade is holding p1's naquadah and waiting on p2's item row while vendor (running for the same player on another connection / alt scenario) is holding the item row and waiting on naquadah — classic ABBA. Postgres' deadlock detector resolves it, but trade gets a `DbError` → `Cancelled` to both clients. Failure shape is safe (no items lost, no cash moved), but it's a noisy user-visible failure and can become flaky under load.

**Severity is medium because:** the row-level FOR UPDATE on `sgw_inventory.item_id` is what actually saves correctness here — a vendor sell that commits first means the trade's lookup returns `None` → `ItemMissing` → rollback. No items lost, no cash spuriously moved. But the deadlock noise will surface as flaky aborted trades under contention.

**Fix:** either standardize on `pg_advisory_xact_lock(player_id, 1)` for trade too, or pick a third namespace (e.g., `2`) and document the namespace assignments in one place. Acquire all advisory locks before any FOR UPDATE on rows.

---

### MEDIUM — Stub InvItem in onTradeState carries is_bound=false and zero durability/charges/ammo

**Handler:** `cell/cell_methods/player/trade.rs::stub_inv_items_for`

The stub `InvItem` records emitted in `RemoteTradeProposal` lie to the partner's client about durability (always 100), bound flag (always false), ammo types (empty), and charges (0). The accompanying comment says "the client's trade panel renders by id lookup against the cached inventory it already has from onUpdateItem, so the stub-icon concern is mostly hypothetical." That assumption needs verification.

If the partner's client trusts these fields (because the partner has never had `onUpdateItem` for the offered item — and they will not, because the item belongs to player A not player B), the partner might agree to a trade believing they're getting a durable / charged / fully-ammoed item and then receive a broken / empty one. The atomic swap is correct; the **client UI may be deceived**.

This isn't a server-authority violation (server-side state is correct), but it's an **information-asymmetry exploit** the player can use socially. Pin a follow-up issue to either fetch full InvItem rows from base for `onTradeState`, or use sentinel values that the client knows mean "unknown — ask" rather than "100". Either way, do not ship 100/0/false placeholders to the wire long-term.

---

### LOW — request_execute_trade clears cell state BEFORE confirming send

**Handler:** `cell/cell_methods/player/trade.rs::request_execute_trade`

The order is: snapshot → `clear_trade_state` → `tx.send(ExecuteTrade)`. If the channel is closed (base task down), the cell-side state is gone, the players' UIs hang at `Locked&Confirmed`, no `onTradeResults` reaches them. The error log says exactly this. Items + cash are unchanged (base never touched the DB), so no loss. The only fix would be to keep the cell state until send succeeds, but reentering the LockedAndConfirmed → execute path is also a bug. Acceptable as-is given the log; advisory only.

---

### Advisory — duplicate-instance check is per-side, not cross-side

The de-dupe via `HashSet` catches `p1 offers item X twice`. If `p1 offers X and p2 also offers X`, base's two `lock_items` calls succeed only if X is owned by exactly one of them — and ownership is unique (one character_id per row). So this is naturally safe. No fix needed; flagging because it would be a TOCTOU shape on a different schema.

---

### Verdict: BLOCK on CRITICAL

The container-whitelist gap is exploitable today by anyone who can send a `tradeRequest` and a `tradeLockState`. The HIGH (distance check) and MEDIUMs can be follow-ups, but the container whitelist must land in this PR or in a fast-follow before #438 merges.

Trade.rs cell-side state machine, version checking, partner-mismatch guards, QA-client workaround, distance-recheck-on-update, and disconnect-cancellation wiring all look correct. The atomic swap inside `FOR UPDATE` is sound — I have no objection to the swap mechanics or the cancellation propagation. The exploit is purely the container-eligibility omission in `lock_items`.

Self-check: every client-supplied field traced; no client timestamp/quantity used as authoritative; no type_id-keyed mutation (UPDATE keys on item_id, correct); dedup-hash + auth-token check inherited from the framing layer (BaseToCellMsg::CellMethodCall arrives after framing-layer validation).

Cadacious:

- Test: `on_trade_results_result_field_is_int32_not_int8` — **VERIFIED**
  - Bug shape pinned: yes (INT8 vs INT32 wire-width)
  - Verification: reverted `result.to_le_bytes()` → `push(result as u8)` in `crates/entity/src/trade.rs:255`; ran the test; failed with `assertion left == right failed: left: 5, right: 8` — exact match to the agent's claim.
  - Risk: none.

- Test: `cancel_clears_state_on_both_sides_and_sends_completed_to_both` — **VERIFIED**
  - Bug shape pinned: yes (cancel result code Completed vs Cancelled)
  - Verification: flipped `ETRADERESULTS_COMPLETED` → `ETRADERESULTS_CANCELLED` in `handle_trade_request_cancel` at `crates/services/src/cell/cell_methods/player/trade.rs:155`; failed with `Got results: [2, 2]` — exact match.
  - Risk: none.

- Test: `trade_update_proposal_without_prior_request_begins_trading` — **VERIFIED**
  - Bug shape pinned: yes (QA-client workaround dropped → no implicit `beginTrading`)
  - Verification: reverted the `if !already_trading { begin_trading(...) }` block to a warn+return; failed with `left: None, right: Some(2)` — exact match.
  - Risk: none.

### Theatre findings

- Test: `cancel_serializes_result_as_completed_one_not_cancelled_two` (in `crates/entity/src/trade.rs:414`) — **THEATRE for TRAP #2**
  - This test asserts `ETRADERESULTS_COMPLETED == 1` and that the serializer emits those bytes when called with that constant. It does NOT exercise the cancel handler.
  - I verified by reverting the cell-side `cancel_session(... ETRADERESULTS_COMPLETED ...)` to `ETRADERESULTS_CANCELLED` — this test **PASSED** under that revert (as expected: it never calls the handler).
  - The agent's claim "fails with `Got results: [2, 2]`" appears to be cross-wired with the cell-level guard.
  - **Suggested fix:** keep this test (it's still a useful constants-stability guard against `enumerations.xml` renumbers), but stop citing it as the TRAP #2 regression guard. The real TRAP #2 guard is the cell-level `cancel_clears_state_on_both_sides_and_sends_completed_to_both` — that one fires.

### Live-DB rollback guards — static-read verdict

The 7 atomic-commit tests in `crates/services/src/base/world_entry/methods/trade/tests.rs` are well-structured:
- `commit_swaps_items_atomically` — exact post-state (player_a naquadah=950, player_b=550, owners flipped)
- `commit_rolls_back_on_insufficient_cash` — $1M demand from a 10-naquadah player → all state unchanged with exact values
- `commit_rolls_back_on_missing_item` — forged id 999_999_999 → real item stays with original owner
- `commit_rolls_back_on_bound_item` — bound flag → unchanged
- `commit_rolls_back_on_buyback_bag_item` — INV_BUYBACK container → unchanged
- `commit_rolls_back_on_duplicate_instance_in_proposal` — same id twice → unchanged
- `commit_rolls_back_on_negative_cash` — `-50` → no cash change

Each test asserts on **exact** values (specific naquadah, specific owner), not `is_ok()`. Sentinels use the `TEST_BASE = 0x7000_0F00` slot with per-test salts — distinct from the existing live-DB sentinel ranges. Cleanup is by exact player_id / account_id, not by range. Not run-verified (no local Postgres in this audit session) but the patterns are convincing.

### Coverage gaps

- No coverage of the **disconnect** path's cancellation result code. The code comment at `trade.rs:154` says "Cancelled (2) is reserved for the disconnect / distance-break / commit-failure paths" — there is a `distance_break_during_update_auto_cancels` test that pins Cancelled-on-distance-break, but no direct test for `cancel_trade_on_disconnect` (line 640). Adding a guard mirrored on the distance-break test would be a clean follow-up.

### Verdict

**PR #438: READY TO MERGE** — 3 of 3 main trap guards verified to fire on revert with exact failure messages matching the agent's claims. One auxiliary test (`cancel_serializes_result_as_completed_one_not_cancelled_two`) is a documentation guard, not a regression guard — recommend not citing it for TRAP #2 in future review correspondence, but no merge blocker.

Cadacious:
## Trust-boundary fixes — 4 commits on top of 389eda46

Addressing the four findings from the server-authority-enforcer adversarial review. Each fix has a regression guard verified by revert + run (in-process tests) or a documented gap (live-DB tests).

| # | Severity | Commit | Regression guard | Revert-verified |
|---|---|---|---|---|
| 1 | CRITICAL | `05ab4cf9` — whitelist INV_MAIN in `lock_items` | `lock_items_rejects_non_inv_main_containers` (live-DB) + `lock_items_accepts_inv_main` (live-DB) | static-read only — see below |
| 2 | HIGH | `fd11df82` — distance recheck at handoff | `execute_handoff_rechecks_distance_and_cancels_on_break` (cell unit) | YES — fails with "Saw ExecuteTrade in the channel — distance recheck missing" |
| 3 | MEDIUM | `5ea1538f` — advisory-lock namespace alignment | `advisory_lock_alignment::lock_sql_is_two_arg_form` (static-read) | YES — fails with exact SQL string mismatch |
| 4 | MEDIUM | `c67d8e5d` — sentinel InvItem fields | `on_trade_state_stub_invitem_carries_sentinel_not_lying_values` (byte-exact wire) | YES — fails with "stack_size MUST be -1 sentinel, NOT the pre-fix lie 1" |

### Fix 1 — container whitelist

Replaced the blacklist `if row.container_id == INV_BUYBACK { reject }` with a whitelist `if !TRADEABLE_CONTAINERS.contains(&row.container_id) { reject }` where `TRADEABLE_CONTAINERS = &[INV_MAIN]`. Matches Python's `canSell()` intent (Trade.py operates on main-bag rows only). Added `TradeAbort::IneligibleContainer` variant; removed `TradeAbort::BuybackOffered` (the whitelist subsumes it — INV_BUYBACK still gets rejected, just through the same code path as every other forbidden container).

Test design: parameterized sweep over `[INV_CHEST, INV_MISSION, INV_BANK, INV_BANDOLIER, INV_BUYBACK]` — each iteration inserts an item in the offending container, attempts the trade, asserts the item stays in its origin container with its original owner.

Revert-verify status: **not run in this session** — no live Postgres on `:5433`. The test design follows the existing live-DB pattern in this file (exact post-state assertions, distinct sentinel range `0x7000_0F00 + 1400 + container*10`, exact-id cleanup). The revert-verify will run when CI's `ci-live-db` profile picks the branch up, OR a reviewer with the bundled Postgres running can execute `DATABASE_URL=... cargo nextest run --profile=ci-live-db -p cimmeria-services --lib lock_items_rejects_non_inv_main_containers` against the pre-fix `lock_items` body to observe failure.

### Fix 2 — distance recheck

Added a `partners_in_range` check at the top of `request_execute_trade`, before the snapshot + handoff. On miss, `cancel_session(..., ETRADERESULTS_CANCELLED)` and return. The earlier check inside the `tradeLockState` handler is unchanged — this is defense-in-depth for any future refactor that introduces an `.await` between the handler's distance check and the handoff.

Test design: hand-rolls the "both sides at LockedAndConfirmed" precondition on the SpaceManager directly (bypassing dispatch, because the lock-state handler's own distance check would intercept any position mutation before `request_execute_trade` ran). Then mutates entity 2's position out of range and calls `request_execute_trade` directly. Asserts: zero `ExecuteTrade` in the channel, both sides receive `onTradeResults(Cancelled=2)`, `trade_partner_entity_id` cleared on both.

### Fix 3 — advisory-lock alignment

Trade was using `pg_advisory_xact_lock(player_id, 0)`. Vendor uses `pg_advisory_xact_lock(player_id, container_id)` with `container_id = INV_MAIN = 1`. Postgres treats `(player_id, 0)` and `(player_id, 1)` as completely independent lock keys — they didn't serialize against each other. Switched trade to the two-arg form binding `INV_MAIN`.

The SQL string is now a module constant `ADVISORY_LOCK_SQL`. The alignment test pins it against the canonical two-arg form, so a regression back to the single-arg form fails CI with the exact mismatch.

A full deadlock-style concurrency test (mirroring `move_and_concurrent_grant_serialize_on_container_lock` in `move_/concurrency_tests.rs`) would require a live DB and isn't included in this PR — the static-read guard is what's gated in CI without DB infrastructure.

### Fix 4 — sentinel InvItem fields

The wire format for `RemoteTradeProposal` requires a fixed-shape `InvItem` FIXED_DICT per item. Option (a) "drop the InvItem payload" is wire-incompatible. Option (b) "round-trip to base for canonical rows" is a multi-message-type refactor (new request/response, deferred onTradeState emission) — flagged as separate follow-up. I went with **option (c)**: emit `UNRESOLVED_INV_ITEM_FIELD = -1` (`0xFFFFFFFF` LE) for every i32 field the cell cannot prove.

Fields changed:
- `stack_size`: `1` → `-1`
- `durability`: `100` → `-1`
- `charges`: `0` → `-1`
- `cur_ammo_type`: `0` → `-1`

Fields kept at pre-fix values (with rationale in the code comment):
- `dbid = 0` — pre-existing sentinel
- `is_bound = false` — emitting `true` would be a worse lie (would imply soul-bound, contradicting the proposal)
- `ammo_types = []` — no in-band sentinel exists for an i32 array

The regression test parses the InvItem out of the serialized `onTradeState` body at the documented byte offsets and asserts each unresolved field encodes `0xFFFFFFFF`.

Caveat: if the partner's client implementation interprets `-1` as a literal value rather than detecting the sentinel, the visible result is bogus values (e.g. "durability: -1") rather than no-information. That's still strictly better than the pre-fix "durability: 100" plausible-lie. The full fix is option (b); ticket should be opened.

### Test totals

- Trade unit + wire tests pre-fix: 19
- Trade unit + wire tests post-fix: 25 (+6 new — 4 here + 2 from advisory-lock pair)
- Full `cimmeria-services` lib test suite: **1217 passing, 0 failing**
- `cargo fmt --check` clean
- `cargo clippy -p cimmeria-services -p cimmeria-entity --all-targets -- -D warnings` clean

### Coordination

Pushed 4 fast-forward commits onto `feat/trading-54` (no force, no rebase). The testing-validation-engineer agent's prior revert-verifies on this PR are unaffected — the existing test names + bodies it audited (`on_trade_results_result_field_is_int32_not_int8`, `cancel_clears_state_on_both_sides_and_sends_completed_to_both`, `trade_update_proposal_without_prior_request_begins_trading`) are untouched.

Cadacious:
Both remaining Copilot inline findings addressed in three commits (`951a58ee`, `8257d0ef`, `3992ebf4`).

## Finding 1 — `trade.rs` > 700-line hard cap

`refactor(trade): split trade.rs into module directory along natural seams` (`951a58ee`)

The cell-side `trade.rs` (1,895 lines after the trust-boundary fix-up) is now `trade/`:

```
crates/services/src/cell/cell_methods/player/trade/
├── mod.rs        (49)   docs + re-exports + MAX_INTERACT_DISTANCE
├── handlers.rs   (426)  4 inbound cell-method handlers + dispatch
├── state.rs      (259)  session lifecycle (begin/apply/cancel/in-range,
│                        cancel_trade_on_disconnect public entry)
├── wire.rs       (202)  onTradeState/onTradeResults senders + stub_inv_items_for
├── handoff.rs    (171)  request_execute_trade + ExecuteTradeSnapshot
└── tests/
    ├── mod.rs        (47)   shared helpers
    ├── handlers.rs   (373)  request/cancel/update/disconnect happy + reject
    ├── lock_state.rs (176)  tradeLockState progression + downgrade + reject
    ├── handoff.rs    (215)  both-confirm + distance-recheck regression guard
    └── wire.rs       (160)  stub_inv_items_for sentinel byte fixture
```

Every submodule is under the 500-line soft cap. Re-export discipline preserved: `crate::cell::cell_methods::player::trade::{dispatch, cancel_trade_on_disconnect}` (the two public entry points used by the dispatch router and base_messages handler) keep their existing paths via `pub use` in `trade/mod.rs`. No external callsite was touched.

## Finding 2 — recipient slot reservation didn't account for outgoing items

`fix(trade): account for outgoing items in recipient slot reservation` (`8257d0ef`)
`test(trade): regression guard for full-bag swap with simultaneous outgoing item` (`3992ebf4`)

The fix: extend `TradeItemRow` with `slot_id`, capture each side's outgoing-INV_MAIN slot IDs, and exclude them from the recipient's "occupied" set when picking destination slots via the new `reserve_main_slots_excluding` + pure `pick_free_main_slots_excluding` split. The exclusion is sound because the swap is one atomic sqlx transaction — every `UPDATE sgw_inventory` lands together or none do, so the vacating slots WILL be free by commit time.

Two-layer regression guard:

- **Unit-level** (`slot_exclusion_accounting` in `execute.rs`, 3 tests, no DB required) — exercises the pure `pick_free_main_slots_excluding` helper directly. Revert-verified locally: temporarily dropping the exclusion filter fails `full_bag_swap_succeeds_with_exclusion_fails_without` and `full_bag_two_for_two_swap_succeeds_with_exclusion` (the discriminating cases), while `partially_full_bag_doesnt_need_exclusion` continues to pass (it doesn't depend on the exclusion).
- **Live-DB** (`commit_succeeds_when_recipient_bag_full_but_trading_slot_away` in `tests.rs`) — full integration. P2's INV_MAIN is 40/40 full; one of those slots holds the item P2 is trading away. Asserts commit succeeds, items swap, outgoing-slot is the destination on the receiver side, untouched filler items stay put. Auto-skips via `require_db_or_skip!` locally; runs in CI's `ci-live-db` profile.

## Verification

- `cargo fmt --check` — clean
- `cargo clippy -p cimmeria-services -p cimmeria-entity --all-targets -- -D warnings` — clean
- `cargo test -p cimmeria-services --lib trade` — 29/29 pass (including the new unit tests; the live-DB test auto-skipped locally)
- Revert-verified locally: removing the exclusion filter in `pick_free_main_slots_excluding` causes 2 of 3 new unit tests to fail with the expected mismatch (`None != Some([39])`, `None != Some([5, 17])`).

Pushed via fast-forward.

## Out of scope / follow-up

`crates/services/src/base/world_entry/methods/trade/execute.rs` is now 932 lines (was 722 before this fix; already over the 700 hard cap before my changes). That's base-side trade, not the file Copilot flagged. Splitting it along seams (handle_execute_trade body / atomic_swap / lock_items + reserve_main_slots / send_results_to_both) would be a separate file-organization PR — happy to do it as a follow-up if you'd like.

Cadacious:
@coderabbitai review

Cadacious:
@coderabbitai review

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `state.rs` `apply_proposal` dedup silently drops duplicates | **The dedup silently drops duplicate instances instead of rejecting the proposal.** A malicious client could send `[item_1, item_1, item_1]` and the server would silently reduce it to `[item_1]`. The partner sees 3 items in the UI but only 1 transfers. Python does the same thing (Trade.py:53-58), so this is parity — but document it as intentional. |
| **G2** | `state.rs` `begin_trading` doesn't check if CALLER is a player | **The code checks `partner.is_player` but not `me.is_player`.** An NPC entity (if one ever gets a player_id) could initiate a trade. Unlikely in practice, but the validation is asymmetric. |
| **G3** | `state.rs` `partners_in_range` uses 3D distance | **`distance_squared_to` includes Y.** A player on a balcony 10m above another player would be "in range" if they're within 5m horizontally. Is this the intended behavior? The Python reference probably does the same thing, but it's worth verifying. |
| **G4** | `handlers.rs` `propagate_unlock_to_partner` logic | **The conditional is `new_lock == ETRADELOCKSTATE_NONE && prev_my_lock != ETRADELOCKSTATE_LOCKED`.** This means: unlock propagates if the actor was NOT locked. If the actor WAS locked and goes to None, propagate is FALSE. The comment says "if one of the players released the trade lock we'll release the lock on the partner as well" but the code does the opposite. Verify against Python Trade.py:207-210. |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G5** | `execute.rs` `read_cash` after commit | **The post-commit `read_cash` is a separate query outside the transaction.** If another transaction modifies naquadah between commit and read, the `onCashChanged` packet would show the wrong value. The window is tiny, but it exists. Consider reading the final balances inside the tx before commit. |
| **G6** | `handoff.rs` channel send failure path | **If `tx.send(ExecuteTrade)` fails, the cell state is already cleared but the trade never commits.** The code logs loudly (`tracing::error!`) but the players' UIs will linger. This is documented as acceptable degradation, but consider sending `onTradeResults(Cancelled)` to both clients on send failure. |
| **G7** | `execute.rs` `TradeAbort` doesn't distinguish which side failed | **The PR body says "Asymmetric `NoLocalCash`/`NoLocalSpace`/`NoRemoteCash`/`NoRemoteSpace` result codes" is out of scope.** But the `TradeAbort` enum already carries `which: &'static str` — the mapping to asymmetric wire codes is trivial. Why defer? |
| **G8** | `wire.rs` `stub_inv_items_for` sets `durability: 100` | **All stub items have 100 durability.** If the partner's client renders durability bars, every item looks pristine. Consider setting `durability: 0` or `-1` to signal "unknown." |
| **G9** | `handlers.rs` no rate limiting on `tradeUpdateProposal` | **A malicious client could spam `tradeUpdateProposal` with incrementing versions, forcing the server to broadcast `onTradeState` to both clients every tick.** The version check prevents replay, but not spam. Consider a per-session rate limit. |

MyPalClara:

| **G10** | No test for `apply_proposal` dedup behavior | The dedup silently drops duplicates, but there's no test pinning this behavior. |
| **G11** | No test for `propagate_unlock_to_partner` edge cases | The conditional logic (G4) isn't explicitly tested. |
| **G12** | No test for `begin_trading` with NPC caller | The asymmetric `is_player` check (G2) isn't tested. |

MyPalClara:

| **G13** | `state.rs` `apply_proposal` dedup comment doesn't explain the UI desync risk | The comment says "second occurrence dropped silently" but doesn't mention the partner sees a different item count. |
| **G14** | `handlers.rs` `propagate_unlock_to_partner` comment contradicts the code | The comment says "if one of the players released the trade lock we'll release the lock on the partner as well" but the conditional does the opposite. |

---

## Summary

| Category | Count |
|----------|-------|
| Correctness issues | 4 |
| Potential issues | 5 |
| Missing test coverage | 3 |
| Documentation | 2 |

**Total: 14 gaps**

Cadacious:
@coderabbitai review

PR #486, opened 2026-06-03
fix(abilities+effects+inventory): four PR #420 follow-ups (A/B/C/E)

Cadacious:
## Summary

Four PR #420 follow-up fixes ship together because they collectively make combat actually function after world entry. Live debug session on 2026-06-02 surfaced all four — peer `72.206.34.241` couldn't attack the Cellblock guard despite holding a properly-granted, properly-equipped pistol. Trace below.

| Fix | Severity | What broke |
|---|---|---|
| **A** | Critical | `useAbility` gate rejected every weapon fire — `entity.abilities.has_ability` returned false because PR #420's per-weapon ability resolver was wired into the right-click path but the resolved IDs were never injected into the player's known set, and the `use_ability` gate didn't consult the resolver as a fallback |
| **B** | Critical | `load_effect_defs` failed at startup with a PG ENUM decode error → `effect_defs` map empty for the whole process lifetime → every combat ability that resolves through an effect silently no-ops |
| **C** | Medium (observability) | `useAbility: entity does not have ability` was DEBUG, hiding A from operators and SigNoz |
| **E** | Medium | `character_create` dropped starter items whose first valid bag was full instead of overflowing to a later valid bag |

## Live trace that surfaced these

```
23:40:58 (server startup)   ⚠ Failed to load effect defs: PG ENUM decode error  ← B
                            → effect_defs = {} for entire process

23:41:24 character_create   ⚠ Bag full, skipping starter item bag=5 item=4343    ← E

23:42:01 cellblock chain    grant item 55 (pistol) + 3730 (ammo)
23:42:17 SyncBandolierItems drew_weapon=true (pistol equipped)
         ⚠ ZERO add_ability calls for entity 2 — ability 579 never enters
           entity.abilities

23:42:31 → 23:42:54         25× useAbility(ability=579, entity=2)
                            EVERY ONE REJECTED at use_ability:125:
                            "entity does not have ability" (DEBUG)         ← A + C

23:43:55 inactivity_timeout client gave up, cleaned up after 60s idle
```

Detail in commit message.

## Test plan

- [x] Regression test `weapon_granted_ability_commits_even_when_not_in_known_set` — pins A; reverting the `is_ability_granted_by_active_weapon` fallback fails the test
- [x] Negative companion `ungranted_ability_still_rejected_when_active_weapon_does_not_grant_it` — pins the fallback isn't an accept-anything hole
- [x] Live-DB regression `load_effect_defs_succeeds_against_seeded_db` — asserts non-empty map AND non-default TCM column (proves the column actually decodes, not just falls back to TCM_Single)
- [x] 5 unit tests on `pick_first_open_bag` — empty / overflow / all-full / no-valid-container / multi-item walk

## PR #420 audit — other ticking time bombs

I walked PR #420 (`da4e3451`) systematically. The four fixes in this PR are the immediate-impact ones. The audit below is for follow-up issues, NOT this PR.

### Effect-script registry coverage gap

The Rust effect-script registry [`cell/effects/registry.rs`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/effects/registry.rs) ships **6 scripts**: `HealHealth`, `HealFocus`, `MeleeDamage`, `AbsorbShield`, `Stun`, `Suppression`.

The canonical effects seed [`db/resources/Effects/Seed/effects.sql`](https://github.com/SandboxServers/Cimmeria/blob/main/db/resources/Effects/Seed/effects.sql) has 3,216 rows. Of those:

- **3,203 rows** have `script_name = NULL` — work via the NVP-only path (HealthDamage / FocusDamage NVPs); no script needed. Fine.
- **13 rows** carry a non-NULL `script_name`. Of these, only 2 distinct values match the registry (`Stun` ×1, `HealFocus` ×1).
- **11 rows** reference scripts not in the registry — silent no-op via `effects/mod.rs:115` `registry::lookup(name)` returning `None`:
  - **`RangedPhysicalDamage` × 7** — probably significant (ranged damage type modifier?)
  - `RangedEnergyDamage` × 1
  - `Reload` × 1
  - `"Medium Cone\nSuppression: 15 Seconds"` × 1 — looks like a multi-line value the seed parser swallowed; either a data bug or a placeholder
  - empty string × 1

The migration [`db/scripts/seed_effect_script_names.sql`](https://github.com/SandboxServers/Cimmeria/blob/main/db/scripts/seed_effect_script_names.sql) sets `script_name` on exactly 3 more effects (659 → `HealFocus`, 1422 → `Stun`, 1475 → `Suppression`). If the migration doesn't run, effect 659 stays `TestEffect` which is also unregistered.

**Follow-up issue**: implement (or no-op-with-warn-once) the 4 missing scripts (`RangedPhysicalDamage`, `RangedEnergyDamage`, `Reload`, fix the parsed-wrong row).

### Mental resist roll is a flag, not a roll

PR body acknowledges this. [`cone_aoe.rs:198`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/abilities/cone_aoe.rs#L198) pushes `mental_resist_roll` into a `categories` log vec — purely observability, no actual resist roll. Crowd-control abilities that should let the target resist (stuns, fears, etc.) currently always land.

**Follow-up issue**: design + wire the mental resist roll formula. Needs design pass on (a) formula, (b) wire surface for "resisted" results (probably an `onEffectResults` flag variant).

### Effect channelled-cancel "TODO" lingering on EffectDef doc

[`entity/src/abilities.rs:141`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/entity/src/abilities.rs#L141) (PR-diff line 882): comment says *"v1 treats interrupt and channel-cancel as TODO."* The implementation in commits 12-13 actually DID wire channel-cancel (`cancel_channels_from_attacker`, `channel_interrupt_on_movement_tick`). The doc comment is stale and will mislead future readers.

**Follow-up**: trivial doc fix.

### Trainer NPC content coverage

PR body acknowledges: today only `template_id=25` ("Interaction Debug NPC") offers abilities, all `ARCHETYPE_Commando`. Every other archetype + every other trainer NPC has empty offerings. Per-class trainer rollouts are content work, not engine.

**Follow-up issue**: content backlog (~560 rows × 7 archetypes).

### Respec is a stub

[`crates/services/src/cell/cell_methods/player/crafting.rs:84`](https://github.com/SandboxServers/Cimmeria/blob/main/crates/services/src/cell/cell_methods/player/crafting.rs#L84) — `respecCrafting` logs `UNIMPLEMENTED`. PR body acknowledges. Default cost = 1000 Naq.

**Follow-up issue**: tracked in PR body, no new finding.

### Bandolier-equip ability sync (deeper than Fix A)

This PR makes the use_ability *gate* consult weapon-granted abilities via the resolver fallback. Cleaner long-term: inject weapon-granted abilities into `entity.abilities` on equip and remove them on unequip. That centralizes the source of truth and also fixes secondary consumers (`onKnownAbilitiesUpdate` to populate the hotbar, ability cooldown display, etc.) that currently might not see weapon abilities.

**Follow-up issue**: refactor equip path to sync weapon abilities into `entity.abilities`. Decide whether to keep the gate-fallback as defense-in-depth.

### `Bag full` warning continues to fire today even after Fix E

Fix E unblocks overflow but won't help if `BAG_FILL_ORDER` itself is incomplete or `container_sets` references container ids outside `BAG_FILL_ORDER`. The new "All valid containers full" log is the operator-actionable signal — if it fires in the wild, that's a content-side fix (more bag slots) or a `BAG_FILL_ORDER` extension.

**Follow-up**: monitor SigNoz for the new warning text; act if seen.

### Effect script registry has no "unknown script" warn-once

`registry::lookup(name).is_none()` is silent today — every effect-apply that references an unknown script just no-ops. Should log a `warn!` once per unknown script name (rate-limited so a popular unknown doesn't spam the index).

**Follow-up issue**: add a `OnceCell<HashSet<String>>`-style first-sighting log in `effects/mod.rs`.

## Risks

- Fix A's fallback is correct for the SGW model but if anything else mutates `entity.abilities` based on equip events, behavior may double-count. Verified no such mutation exists today (`grep add_ability` returned only test code + train_ability paths).
- Fix B's `::TEXT` cast preserves the existing `Option<String>` shape — no other call sites depend on the column type.
- Fix C's WARN promotion may surface noise if any path legitimately calls useAbility for abilities the player doesn't have (e.g., speculative fires from the auto-cycle loop). Auto-cycle code-path was checked — it stashes the resolved ability_id at arm-time, so it always invokes with an ability the entity has known.
- Fix E's helper extraction is a refactor — the bag-pick + INSERT loop stays identical; only the predicate moved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Starter item placement during character creation now reliably fills available containers and gives clearer failure warnings.
  * Effect definitions now load correctly from the database, restoring expected in-game effects.

* **New Features**
  * Players can use abilities granted by their currently equipped weapon even if not listed in their known ability set.

* **Tests**
  * Added regression tests covering weapon-granted abilities, starter-item placement behavior, and effect-definition loading.

### DIFF
--- a/crates/services/src/base/character_create.rs
+++ b/crates/services/src/base/character_create.rs
@@ -12,7 +12,7 @@ use crate::mercury::read_wstring;
 use super::character::{query_character_list, send_char_create_failed};
 use super::chardef::chardef_lookup;
 use super::helpers::{drain_acks_and_seq, get_account_entity_id};
-use super::resources::{bag_max_slots, bag_min_slot, pick_first_open_bag, BAG_FILL_ORDER};
+use super::resources::{bag_min_slot, pick_first_open_bag, BAG_FILL_ORDER};
 use super::ConnectedClientState;
 
 /// Handle `createCharacter` (0xC4) -- parse args and INSERT into sgw_player.
@@ -456,20 +456,6 @@ pub(crate) async fn handle_create_character(
                     .or_insert_with(|| bag_min_slot(bag_id));
                 let current_slot = *entry;
                 *entry += 1;
-
-                // Defensive: the `.find()` predicate above already gates on
-                // `next_slot < bag_max_slots(bag)`, so this branch is
-                // unreachable through normal flow. Keep it as a guard
-                // against future refactors that might separate selection
-                // from increment.
-                if current_slot >= bag_max_slots(bag_id) {
-                    tracing::error!(
-                        %addr, bag_id, item_id = item.item_id,
-                        "Bag-full guard tripped after fill-order selection — \
-                         selection logic and slot accounting are out of sync"
-                    );
-                    continue;
-                }
 
                 if let Err(e) = sqlx::query(
                     "INSERT INTO sgw_inventory \
--- a/crates/services/src/base/character_create.rs
+++ b/crates/services/src/base/character_create.rs
@@ -12,7 +12,7 @@ use crate::mercury::read_wstring;
 use super::character::{query_character_list, send_char_create_failed};
 use super::chardef::chardef_lookup;
 use super::helpers::{drain_acks_and_seq, get_account_entity_id};
-use super::resources::{bag_max_slots, bag_min_slot, BAG_FILL_ORDER};
+use super::resources::{bag_max_slots, bag_min_slot, pick_first_open_bag, BAG_FILL_ORDER};
 use super::ConnectedClientState;
 
 /// Handle `createCharacter` (0xC4) -- parse args and INSERT into sgw_player.
@@ -417,14 +417,36 @@ pub(crate) async fn handle_create_character(
                 .flatten()
                 .unwrap_or_default();
 
-                // Find the best container from BagFillOrder
-                let bag_id = match BAG_FILL_ORDER
-                    .iter()
-                    .find(|&&bag| container_sets.contains(&bag))
-                {
-                    Some(&bag) => bag,
+                // Pick the first bag that's both valid for this item AND
+                // still has room. Pre-fix this picked the first valid bag
+                // unconditionally and `continue`d if it was full — so an
+                // item that could overflow to a later bag was silently
+                // dropped (live observation 2026-06-02: item 4343 lost
+                // at character create because its primary bag filled up
+                // first while a later valid bag still had room).
+                let bag_id = match pick_first_open_bag(&container_sets, &slot_indices) {
+                    Some(bag) => bag,
                     None => {
-                        tracing::warn!(%addr, item_id = item.item_id, "No valid container for starter item");
+                        // Either the item has no valid container (content
+                        // gap) or every valid container is genuinely full.
+                        // Both are operator-actionable — surface the
+                        // distinction so a real content gap doesn't get
+                        // confused with "too many starter items."
+                        let any_valid_container =
+                            BAG_FILL_ORDER.iter().any(|b| container_sets.contains(b));
+                        if any_valid_container {
+                            tracing::warn!(
+                                %addr,
+                                item_id = item.item_id,
+                                "All valid containers full — starter item dropped"
+                            );
+                        } else {
+                            tracing::warn!(
+                                %addr,
+                                item_id = item.item_id,
+                                "No valid container for starter item"
+                            );
+                        }
                         continue;
                     }
                 };
@@ -435,8 +457,17 @@ pub(crate) async fn handle_create_character(
                 let current_slot = *entry;
                 *entry += 1;
 
-                if *entry > bag_max_slots(bag_id) {
-                    tracing::warn!(%addr, bag_id, item_id = item.item_id, "Bag full, skipping starter item");
+                // Defensive: the `.find()` predicate above already gates on
+                // `next_slot < bag_max_slots(bag)`, so this branch is
+                // unreachable through normal flow. Keep it as a guard
+                // against future refactors that might separate selection
+                // from increment.
+                if current_slot >= bag_max_slots(bag_id) {
+                    tracing::error!(
+                        %addr, bag_id, item_id = item.item_id,
+                        "Bag-full guard tripped after fill-order selection — \
+                         selection logic and slot accounting are out of sync"
+                    );
                     continue;
                 }
 
--- a/crates/services/src/base/resources/mod.rs
+++ b/crates/services/src/base/resources/mod.rs
@@ -49,6 +49,44 @@ pub(crate) fn bag_min_slot(_container_id: i32) -> i32 {
     0
 }
 
+/// Pick the first bag in [`BAG_FILL_ORDER`] that (a) is in the item's
+/// `container_sets` list AND (b) still has room based on the per-bag
+/// next-free-slot map.
+///
+/// Used by `character_create` to place starter items. Pre-fix this was
+/// inline in `handle_create_character` and the selection ignored
+/// fullness — picking the first valid bag unconditionally then
+/// `continue`-ing if it was full. That dropped items that could
+/// otherwise overflow to a later bag in the fill order (live observation
+/// 2026-06-02: item 4343 dropped at character create because its
+/// primary bag filled up first while a later valid bag still had room).
+///
+/// `slot_indices` carries the next free slot for each bag that's been
+/// touched so far. A bag with no entry yet starts at
+/// `bag_min_slot(bag)`; an entry that has reached `bag_max_slots(bag)`
+/// is full.
+///
+/// Returns `Some(bag_id)` for the first bag that satisfies both gates,
+/// `None` if no valid + non-full bag exists.
+pub(crate) fn pick_first_open_bag(
+    container_sets: &[i32],
+    slot_indices: &HashMap<i32, i32>,
+) -> Option<i32> {
+    BAG_FILL_ORDER
+        .iter()
+        .find(|&&bag| {
+            if !container_sets.contains(&bag) {
+                return false;
+            }
+            let next_slot = slot_indices
+                .get(&bag)
+                .copied()
+                .unwrap_or_else(|| bag_min_slot(bag));
+            next_slot < bag_max_slots(bag)
+        })
+        .copied()
+}
+
 // ── Resource cache ───────────────────────────────────────────────────────────
 
 /// Per-category cooked data loaded from a PAK file.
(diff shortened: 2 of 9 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `resolve.rs` `is_ability_granted_by_active_weapon` iterates 3 event IDs | **The function probes `EVENT_ITEM_RANGED`, `EVENT_ITEM_MELEE`, `EVENT_ITEM_USE_ABILITY` in sequence.** If a weapon binds the SAME ability_id to multiple event_ids (unlikely but possible), the function returns true on the first match. **This is correct behavior, but the doc comment doesn't mention the iteration order or early-return semantics.** |
| **G2** | `use_ability/mod.rs` fallback drops immutable borrow | **The comment says "We need to drop the immutable \`entity\` borrow before the call into the helper."** But the code structure already does this — the `entity` binding goes out of scope at the end of the `if !entity.abilities.has_ability(ability_id)` block. **The comment is slightly misleading about WHY the borrow is dropped (it's scope, not explicit drop).** |
| **G3** | `abilities.rs` `::TEXT` cast is a workaround | **The comment says "Long-term fix is a \`#[sqlx(type_name = ...)]\` Rust enum mirror."** But there's no follow-up issue filed for this. **The cast works, but the tech debt is undocumented outside this PR.** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `use_ability/mod.rs` WARN promotion rate | **The comment says "The 25-per-second rate observed during the regression... is exactly the volume that disappears at DEBUG and burns at WARN."** But if a legitimate edge case causes repeated rejections (e.g., a bugged ability binding), the WARN would spam logs. **Consider rate-limiting the WARN per (entity_id, ability_id) pair.** |
| **G5** | `resources/mod.rs` `pick_first_open_bag` doesn't validate `container_sets` | **The function trusts that `container_sets` contains valid container IDs.** If a content error puts an invalid ID in `container_sets`, the function would skip it (correct) but without logging. **Consider a debug-level log for "container_sets contains ID not in BAG_FILL_ORDER."** |
| **G6** | `character_create.rs` defensive guard is `error!` | **The "Bag-full guard tripped after fill-order selection" log is `error!` level.** But the comment says it's "unreachable through normal flow" — if it fires, it's a logic bug, not an operational error. **Consider `debug_assert!` or `unreachable!` instead of `error!` + `continue`.** |

MyPalClara:

| **G7** | No test for `is_ability_granted_by_active_weapon` with multiple event_ids binding same ability | The function iterates 3 event_ids, but the test only covers single-event bindings. |
| **G8** | No test for `is_ability_granted_by_active_weapon` with NPC entity | The function checks `bandolier_items` which NPCs don't have. A test would pin the "returns false for NPCs" behavior. |
| **G9** | No test for `load_effect_defs` with NULL `target_collection_method` | The code falls back to `TCM_SINGLE` for NULL, but the live-DB test only checks for non-NULL values. |

MyPalClara:

| **G10** | `resolve.rs` `is_ability_granted_by_active_weapon` doesn't document NPC behavior | The function returns false for entities without bandolier items, but the doc comment doesn't mention this. |
| **G11** | PR body audit section is excellent but not linked to issues | The audit identifies 8 follow-up items but doesn't create GitHub issues for them. The PR body will be harder to find than issue links. |

---

## Summary

| Category | Count |
|----------|-------|
| Correctness issues | 3 |
| Potential issues | 3 |
| Missing test coverage | 3 |
| Documentation | 2 |

**Total: 11 gaps, all low severity**

Cadacious:
Addressed in `7de4b7c5`. Walkthrough by reviewer:

### Substantive fixes

- **Copilot — `use_ability/mod.rs:158` (WARN spam on client-controlled path)**: split severity on `ability_def.is_some()`. WARN only when the server knows the ability (real wiring issue — player tried to fire a valid ability the active weapon doesn't bind). DEBUG when the ability_id has no server def at all (forged/stale client packet). Preserves Fix C's original "real bugs visible at WARN" intent while closing the spam vector.
- **CodeRabbit `character_create.rs:472` + Copilot `:465` + Clara G6**: deleted the defensive bag-full guard and its `bag_max_slots` import. All three reviewers agreed; repo guideline disallows defensive code for impossible scenarios. Also resolves codecov's 1-line missing-coverage hit (the unreachable `tracing::error!`).
- **Clara G7 / G8** (`resolve.rs` test gaps): added three tests on `is_ability_granted_by_active_weapon` — `accepts_melee_binding_too` (iteration covers all event ids, not just RANGED), `same_ability_multi_event_returns_true` (any-match-wins), `returns_false_for_entity_without_bandolier` (NPC behavior).
- **Clara G9** (`load_effect_defs` NULL fallback): added a companion assertion to the existing live-DB guard — at least one row in the loaded map must carry `TCM_SINGLE`, exercising the `unwrap_or_else(TCM_SINGLE)` path for NULL rows.
- **Clara G1 / G10** (`resolve.rs` doc): documented iteration order + early-return semantics + NPC-returns-false behavior on `is_ability_granted_by_active_weapon`.
- **Clara G2** (`use_ability/mod.rs` misleading "drop borrow" comment): rewritten — the entity binding is scope-dropped, not explicitly dropped. Comment now describes what's happening instead of asserting a borrow movement.

### Style sweep (CodeRabbit + Copilot)

Stripped PR/issue numbers and live-session player IPs from source comments across **six files**: `character_create.rs`, `resolve.rs`, `use_ability/mod.rs`, `use_ability/tests.rs`, `spawner/abilities.rs`, `spawner/tests.rs`. Per repo guideline. Behavioral explanations kept where load-bearing; identifying details removed.

### Deferred with reasons (in commit message + reproduced here)

- **Clara G3** (`#[sqlx(type_name = ...)]` enum mirror for the long-term effect-def loader fix): valid tech debt. `::TEXT` cast is the safe immediate patch. Will file as a separate follow-up issue alongside the other PR #420 audit items.
- **Clara G4** (rate-limit WARN per `(entity_id, ability_id)`): the Copilot fix above bounds WARN volume to "real wiring issues only" — small distinct (entity, ability) cardinality in that category. Per-pair rate-limit adds a `Mutex<HashMap>` with timestamps for marginal value; skipping.
- **Clara G5** (debug log when `container_sets` contains IDs not in `BAG_FILL_ORDER`): would noise on every starter item with any non-fill-order container — most of which are invalid by design. Skipping.
- **Clara G11** (file the audit items in the PR body as issues): the audit section enumerates the follow-ups; will file separately on operator request rather than auto-creating issues from this PR.

`cargo check -p cimmeria-services` clean; `cargo fmt --all` ran.

PR #488, opened 2026-06-04
feat(effects+dialog): RangedPhysical/Energy scripts + monologue dialog path

Cadacious:
## Summary

Two independent gaps surfaced in the latest play-session SigNoz trace, bundled into one PR per request.

### `RangedPhysicalDamage` + `RangedEnergyDamage` Rust scripts

The legacy effect-script dispatcher has 6 names registered; live combat WARNed ~2 Hz on every shot:

```
effect_script_unknown script=RangedPhysicalDamage effect_id=641
```

Effect 641 backs ability 579 (Pistol Auto Attack) — the canonical player auto-attack. With the script missing, every shot fell back to the NVP-only path: `HealthDamage=10` + `FocusDamage=100` applied independently. **The player took HEALTH damage from shot one even at full Focus**, and the NPC's return fire bypassed the player's Focus shield the same way. ~150 WARNs per fight.

Both scripts ported from `deprecated/python/cell/effects/`:

- **`RangedPhysicalDamage`** — focus-gated. Apply FocusDamage to FOCUS pool; if overflow exists, `spillover = overflow / 3` (the legacy `remaining_pct × FocusDamage / 300` collapses algebraically); apply `spillover + HealthDamage` to HEALTH. **If Focus absorbs everything, no HEALTH damage at all** — the load-bearing "shields held" gate that distinguishes this from the NVP fallback. Reverting the gate fails the new `ranged_physical_full_focus_absorbs_no_health_damage` test.
- **`RangedEnergyDamage`** — parallel apply to HEALTH and FOCUS, no gating. Energy weapons "ignore shields" by design.

### `DisplayDialog` monologue path

Same SigNoz trace, every session start:

```
WARN DisplayDialog: no NPC entity id in chain params or
     last_interaction_target chain_id=1001 dialog_id=2982
```

Dialog 2982 is the Castle Cellblock wake-up monologue. Two screens, both `speaker_id = 0` (player inner-thought):

> "The last thing you remember is going into stasis after another session with that sadist, Romney..."
>
> "But now you are free — and awake. ... you need to find a weapon. Quickly. Perhaps there's something on those corpses nearby..."

The existing `DisplayDialog` executor required an NPC entity id and bailed when none was resolved — correct for NPC dialogs (binding the player would blank the NPC portrait and substitute the player's name for every screen), wrong for monologue dialogs (where binding the player is exactly the right render — the client's per-screen lookup of `speaker_id=0` falls back to the player's name naturally, per [the RE finding](docs/reverse-engineering/findings/dialog-portrait-lookup.md)).

**5,713 of 13,466 dialog screens in the canonical seed carry `speaker_id = 0`** — about 42% of the entire authored dialog catalog was silently never displaying for chains that didn't resolve an NPC context.

Fix shape:

- New `load_monologue_dialog_ids(pool)` loader in [`spawner/dialogs.rs`](crates/services/src/cell/spawner/dialogs.rs). PG query groups by dialog_id and filters to those where every screen has `speaker_id = 0` (`COUNT(*) FILTER (WHERE speaker_id <> 0) = 0`).
- `SpaceManager.monologue_dialog_ids: HashSet<i32>` cache populated at cell startup.
- `DisplayDialog` executor: when no NPC resolves AND the dialog is in the monologue set, bind the player's own entity id as the wire `EntityId`. Otherwise, preserve the existing warn-and-bail path — the cache acts as the gate.

## Test plan

- [x] 8 unit tests on the two new scripts cover: full absorb, partial absorb with spillover, zero-focus full-overflow, zero-FocusDamage edge, missing-target no-op, parallel energy apply, energy zero-NVP, energy pool-clamp-at-zero
- [x] 2 executor tests on the monologue path: `display_binds_player_when_no_npc_and_dialog_is_monologue` (the cellblock wake-up case) and `display_still_aborts_for_npc_dialog_not_in_monologue_cache` (pins the fallback isn't an accept-anything hole)
- [x] 1 live-DB test: `load_monologue_dialog_ids_includes_cellblock_wakeup` asserts dialog 2982 makes it into the cache AND that at least one mixed-speaker dialog from the seed does NOT

## What this fixes for the player

- Pistol-vs-Pistol exchange now feels like the design intended: Focus traded first ("shields hold"), then Health bleeds when shields are down. Previously every shot chipped Health from full Focus.
- ~150 fewer WARNs per fight in SigNoz (no more `effect_script_unknown RangedPhysicalDamage` flood)
- Cellblock wake-up narration finally shows on session start — the player gets the "find a weapon, check the corpses" prompt that explains what to do, instead of waking up in silence

## Risks

- Both scripts currently use raw stat mutation (matching `MeleeDamage`'s pattern) — not the full QR + armor + absorption pipeline of `combat::calculate_damage`. The pipeline migration is a separate effort; doing it here would couple two refactors. Behavior is consistent with the NVP fallback (which also bypasses absorption for FOCUS damage) — net behavior change vs. today is the new spillover gate + correct shield-first cadence.
- The monologue cache is computed at startup and not re-evaluated. If a content patch adds new monologue dialogs at runtime (it doesn't today), they'd silently fall through to the bail path until next restart. Acceptable for current content authoring cadence.
- The monologue branch only triggers when *no* NPC resolves. NPC dialogs whose chain wired `target_entity_id` correctly are unaffected.

## Out of scope

- Migration of `MeleeDamage` (and the new scripts) onto the unified `calculate_damage` pipeline.
- Other PR #420 audit items from #486 — separate follow-ups still queued.

## Summary by CodeRabbit

* **New Features**
  * Monologue dialog support: dialogues without an NPC now display as player-led monologues.
  * Two new ranged damage effects: RangedPhysicalDamage (focus-absorption with spillover) and RangedEnergyDamage (independent focus and health damage).

* **Improvements**
  * Startup now loads and caches monologue dialog identifiers for faster resolution.

* **Tests**
  * Added tests covering monologue display and the new ranged damage behaviors.

### DIFF
--- a/crates/services/src/cell/content/executor/dialog.rs
+++ b/crates/services/src/cell/content/executor/dialog.rs
@@ -486,21 +486,10 @@ mod tests {
         );
     }
 
-    /// **Monologue fallback: when no NPC resolves AND the dialog is a
-    /// player-monologue (all screens speaker_id=0), bind the player as
-    /// the wire EntityId instead of bailing.** This is the cellblock
-    /// wake-up case — dialog 2982 fires from chain 1001 with no NPC
-    /// context, both screens carry `speaker_id = 0`, and the intended
-    /// render is the player's own portrait + name (inner thought). The
-    /// client's per-screen lookup of speaker_id=0 falls back to the
-    /// player's name naturally (see RE finding
-    /// `docs/reverse-engineering/findings/dialog-portrait-lookup.md`).
-    ///
-    /// Without this branch, ~42% of authored dialog screens (the
-    /// speaker_id=0 ones) silently never displayed. Live observation
-    /// 2026-06-02: dialog 2982 chain 1001 fired this exact bail every
-    /// session start. Reverting the monologue branch must fail this
-    /// test.
+    /// When no NPC resolves and the dialog is in the monologue cache,
+    /// the wire EntityId must be the player's own id (not bail). The
+    /// per-screen `speaker_id = 0` lookup on the client falls back to
+    /// the player's name, rendering as inner thought.
     #[tokio::test]
     async fn display_binds_player_when_no_npc_and_dialog_is_monologue() {
         let mut mgr = make_space_manager();
--- a/crates/services/src/cell/content/executor/dialog.rs
+++ b/crates/services/src/cell/content/executor/dialog.rs
@@ -10,10 +10,9 @@ use crate::cell::space_manager::SpaceManager;
 /// to the player. They take different field names but the same id semantics,
 /// merged into one handler.
 ///
-/// The wire `EntityId` field of `onDialogDisplay` MUST be the NPC the
-/// player is talking to — the client uses it as the key into
-/// `LookupEntityListenerEntry` to bind the dialog portrait actor and the
-/// per-screen speaker entity (see
+/// The wire `EntityId` field of `onDialogDisplay` is the key the client
+/// uses for `LookupEntityListenerEntry` to bind the dialog portrait actor
+/// and the per-screen speaker entity (see
 /// `docs/reverse-engineering/findings/dialog-portrait-lookup.md`).
 /// Resolution order, from most to least direct:
 ///
@@ -24,15 +23,25 @@ use crate::cell::space_manager::SpaceManager;
 ///    `handle_interact`. Covers follow-up chains (e.g. an
 ///    `OnDialogChoice` trigger that fires another `display_dialog` on
 ///    the same NPC) where the trigger event itself carries no NPC.
-/// 3. Abort with a `warn` — falling back to the player's own entity ID
-///    here is the bug this commit fixed: the client would bind the
-///    player as the dialog speaker, blanking the portrait and
-///    rendering the player's name for every screen.
+/// 3. **Monologue fallback** — if the dialog is in
+///    `space_mgr.monologue_dialog_ids` (every screen has
+///    `speaker_id = 0`, i.e. player-narration / inner-thought), bind
+///    the player as the wire EntityId. The client's per-screen lookup
+///    of `speaker_id = 0` naturally falls back to the player's name and
+///    the portrait shows the player — exactly the intended render for
+///    "the player is talking to themselves." Without this branch,
+///    monologue dialogs (~42% of authored screens) silently never
+///    display.
+/// 4. Abort with a `warn` — the dialog is NPC-shaped but no NPC could
+///    be resolved. Binding the player here would blank the NPC portrait
+///    and substitute the player's name for every NPC line — the bug
+///    that motivated the original gate. Bail loud so the chain author
+///    can fix the missing target_entity_id wire-up.
 #[tracing::instrument(
     name = "dialog.display",
     level = "info",
     skip_all,
-    fields(entity_id, dialog_id, chain_id, npc_entity_id = tracing::field::Empty)
+    fields(entity_id, dialog_id, chain_id, npc_entity_id = tracing::field::Empty, monologue = tracing::field::Empty)
 )]
 pub(super) async fn display(
     dialog_id: i32,
@@ -56,14 +65,27 @@ pub(super) async fn display(
     let npc_entity_id = match npc_entity_id {
         Some(id) => id,
         None => {
-            tracing::warn!(
-                entity_id,
-                dialog_id,
-                chain_id,
-                "DisplayDialog: no NPC entity id in chain params or last_interaction_target -- \
-                 cannot send onDialogDisplay (would bind player as speaker and blank portrait)"
-            );
-            return;
+            // Monologue fallback — see doc on this fn for rationale.
+            if space_mgr.monologue_dialog_ids.contains(&dialog_id) {
+                tracing::Span::current().record("monologue", true);
+                tracing::info!(
+                    entity_id,
+                    dialog_id,
+                    chain_id,
+                    "DisplayDialog: no NPC resolved, dialog is player-monologue \
+                     (all screens speaker_id=0) — binding player as context entity"
+                );
+                entity_id as i32
+            } else {
+                tracing::warn!(
+                    entity_id,
+                    dialog_id,
+                    chain_id,
+                    "DisplayDialog: no NPC entity id in chain params or last_interaction_target -- \
+                     cannot send onDialogDisplay (would bind player as speaker and blank portrait)"
+                );
+                return;
+            }
         }
     };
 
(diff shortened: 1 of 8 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `scripts.rs` `RangedPhysicalDamage` integer division truncation | **`spillover = focus_overflow / 3` uses integer division.** An overflow of 1 or 2 produces spillover = 0, meaning tiny overflows add no health damage beyond the base `HealthDamage` NVP. **This matches the legacy Python behavior (integer division there too), but document it as intentional — a future contributor might "fix" it to floating-point and change combat feel.** |
| **G2** | `scripts.rs` `RangedPhysicalDamage` zero-FocusDamage gate | **When `FocusDamage = 0`, the script returns without applying `HealthDamage` at all.** The test `ranged_physical_zero_focus_damage_skips_health_too` pins this, but **is this the intended design?** A physical weapon with only `HealthDamage` configured would do nothing. The comment says "matches the legacy script's conditional branch off the QR result" — verify this is correct parity. |
| **G3** | `dialogs.rs` `load_monologue_dialog_ids` query assumes `speaker_id` is never NULL | **The query uses `speaker_id <> 0` which would exclude NULL values from the count.** If any `dialog_screens` row has `speaker_id = NULL`, it would be treated as "not a monologue screen" and the dialog would be excluded from the cache. **Is NULL a valid value in the schema? If so, should it be treated as 0 (player) or as "unknown speaker"?** |

MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `startup.rs` monologue cache failure is WARN, not ERROR | **The startup logs `warn!` on cache load failure and continues with an empty set.** This is documented as "graceful degradation" but **42% of dialog screens silently won't display if the cache fails to load.** Consider promoting to `error!` so operators notice, or adding a startup health check that surfaces the gap. |
| **G5** | `space_manager/mod.rs` `monologue_dialog_ids` is `HashSet<i32>` not `HashSet<u32>` | **Dialog IDs are `i32` throughout, but the PR body mentions "5,713 of 13,466 dialog screens."** If dialog_id can be negative (unlikely but schema-dependent), the HashSet would handle it correctly. **Verify the schema constraint — if dialog_id is always positive, consider documenting that assumption.** |
| **G6** | `scripts.rs` `RangedEnergyDamage` doesn't log when target is missing | **`RangedPhysicalDamage` has a debug log for "Focus absorbed all damage" but `RangedEnergyDamage` has no equivalent for the missing-target early return.** Inconsistent observability between the two scripts. |
| **G7** | `scripts.rs` both scripts use `let _: i8 = DT_PHYSICAL/DT_ENERGY` | **The damage_type constants are assigned to `_` with a comment about future pipeline migration.** This is a clever way to document intent, but **a future contributor might delete the "unused" line.** Consider a `#[allow(unused)]` annotation or a more explicit comment. |

MyPalClara:

| **G8** | No test for `RangedPhysicalDamage` with negative NVP values | **The code does `.max(0)` on both NVPs, but there's no test pinning that negative values are clamped to 0.** |
| **G9** | No test for `RangedEnergyDamage` with missing target | **`RangedPhysicalDamage` has `ranged_physical_missing_target_is_noop` but `RangedEnergyDamage` doesn't have an equivalent.** |
| **G10** | No test for monologue cache with zero-screen dialogs | **The query has `HAVING COUNT(*) > 0` to filter out zero-screen dialogs, but there's no test verifying this edge case.** |
| **G11** | No test for `display()` when `monologue_dialog_ids` cache is empty but dialog IS a monologue | **The live-DB test checks that dialog 2982 is in the cache, but doesn't test the executor behavior when the cache failed to load (empty set).** |

MyPalClara:

| **G12** | `scripts.rs` `RangedPhysicalDamage` doc doesn't mention the integer division behavior | **The doc explains the `/3` algebraic simplification but doesn't note that small overflows (1-2) produce zero spillover.** |
| **G13** | `dialog.rs` monologue branch doesn't cite the 42% statistic | **The PR body mentions "5,713 of 13,466 dialog screens" but the code comment just says "~42%."** Consider adding the exact count to the code comment for future reference. |
| **G14** | No issue filed for the `calculate_damage` pipeline migration | **Both scripts have comments about "follow-up that unifies every script's stat-mutation path" but there's no linked issue.** The PR body's "Out of scope" section mentions it but doesn't create a tracking issue. |

---

## Summary

| Correctness issues | 3 | **Low-Medium** — G1-G3 are parity questions |
| Potential issues | 4 | **Low** — G4-G7 are hardening |
| Missing test coverage | 4 | **Low** — G8-G11 are edge cases |
| Documentation | 3 | **Low** — G12-G14 are polish |

**Total: 14 gaps, all low-medium severity.**

---

## What's Excellent

1. **The spillover algebra is correct.** The `/300` → `/3` simplification is mathematically sound and the test `ranged_physical_partial_absorb_spills_to_health` pins the exact arithmetic.

2. **The monologue cache query is elegant.** `COUNT(*) FILTER (WHERE speaker_id <> 0) = 0` is a clean way to express "all screens have speaker_id = 0" without a subquery.

3. **The tests pin the exact bug shapes.** `ranged_physical_full_focus_absorbs_no_health_damage` would fail if the shield-first gate were removed. `display_binds_player_when_no_npc_and_dialog_is_monologue` would fail if the monologue branch were reverted.

4. **The live-DB test is load-bearing.** `load_monologue_dialog_ids_includes_cellblock_wakeup` asserts dialog 2982 is in the cache AND that mixed-speaker dialogs are excluded — both halves of the predicate are verified.

5. **The PR body is exceptional.** Live SigNoz trace, exact WARN counts, algebraic derivation, RE finding citations, explicit out-of-scope documentation.

---

## Verdict

**This PR is ready to ship.** The two fixes are well-scoped, well-tested, and well-documented. The 14 gaps are all low-medium severity polish items.

- `RangedPhysicalDamage` fixes the "shields held but I still took damage" feel
- `RangedEnergyDamage` provides the "ignore shields" counterpart
- Monologue fallback unblocks 42% of authored dialog screens

**Ship it.** 🔬🎯💬

Cadacious:
Addressed in `c629c178`.

### Substantive

- **Copilot + CodeRabbit + Clara G3** (NULL `speaker_id`): the predicate `COUNT(*) FILTER (WHERE speaker_id <> 0) = 0` treated NULL as not-non-zero, so all-NULL-speaker dialogs were wrongly admitted to the monologue cache. Rewrote to require explicit `speaker_id = 0` on every row. Added a live-DB assertion that a NULL-speaker dialog from the seed is NOT in the cache.
- **Copilot G4 + Clara G1** (spillover parity): restored the legacy two-step integer truncation `(overflow*100/total) * total / 300` and reverted the algebraic `overflow / 3` shortcut. With FocusDamage=80, overflow=3, the algebraic shortcut gave 1 spillover while the legacy gives 0. Added regression test `ranged_physical_small_overflow_truncates_to_zero_spillover` that fails on the algebraic shortcut.
- **CodeRabbit + Clara G7** (dead `let _: i8 = DT_*` bindings): removed.

### Polish

- **CodeRabbit** (dialog test comment): trimmed from 14 lines of provenance to 4 lines of invariant.
- **Clara G4** (startup cache failure): WARN → ERROR. An empty cache silently strips ~42% of authored dialog screens.
- **Clara G6** (missing-target log): RangedEnergyDamage now logs at DEBUG, matching the physical script.
- **Clara G8/G9** (test gaps): added `ranged_physical_negative_nvps_clamp_to_zero` and `ranged_energy_missing_target_is_noop`.
- **Clara G10** (NULL-speaker dialog test): folded into the existing live-DB test.

### Deferred with reasons (in commit body)

- Clara G2 (zero-FocusDamage gate): current behavior matches legacy + avoids div-by-zero. Skipping.
- Clara G5 (i32 vs u32): schema uses signed `integer`. Match.
- Clara G11 (executor test with empty cache for monologue dialog): the existing `display_still_aborts_for_npc_dialog_not_in_monologue_cache` covers the same failure shape.
- Clara G12/G13 (re-add 42% / exact counts to source comments): CodeRabbit explicitly flagged exactly this provenance as rot-prone. Keeping comments terse.
- Clara G14 (file follow-up issue for `calculate_damage` migration): out of scope for this commit.

`cargo check -p cimmeria-services --tests` clean; `cargo fmt --all` ran.

PR #490, opened 2026-06-04
fix(bandolier): clear auto-cycle stash on weapon swap

Cadacious:
## Summary

`handle_request_active_slot_change` cleared `pending_*` reload/attack state on a weapon swap but left `auto_cycle_ability_id` and `last_fired_ability_id` intact. Both are stashed at fire time, resolved against the THEN-active weapon's `items_event_sets`. Swapping weapons mid-session leaves a stale id pointed at the *outgoing* weapon — the next `setAutoCycle` press fires that stale id, the `use_ability` gate rejects with *"not in known set and not granted by active weapon"*, no shots fire, the cycling icon stays lit on the client.

## Live observation

SigNoz session 2026-06-04 10:04 UTC, peer `176.3.8.17:3650` (lomiada1):

```
10:04:39.043566  setAutoCycle  enabled=true  entity_id=2
10:04:39.043580  setAutoCycle: immediate fire on enable
                 ability_id=559  target_id=100130
10:04:39.043611  useAbility: ability not in known set
                 and not granted by active weapon
                 ability_id=559  entity_id=2
```

Ability 559 = SMG/P90 Auto Attack. lomiada1's active weapon at the time wasn't an SMG, so the resolver fallback failed and the shot dropped.

## Fix

On a real slot change (`slot != prev_slot`), clear:

- `entity.abilities.auto_cycle`
- `entity.abilities.auto_cycle_ability_id`
- `entity.abilities.last_fired_ability_id`
- `BSF_AUTO_CYCLING` state-field bit

Broadcast `onStateFieldUpdate` when the bit transitioned so the cycling icon clears on the client. Same-slot calls (replayed packets / no-op swaps) preserve the stash — the loop continues as intended.

## Test plan

- [x] `cargo fmt --all` ran
- [x] `slot_change_clears_auto_cycle_stash_and_broadcasts` — fixture matches lomiada1's state (auto_cycle on with ability 559, swap to slot 1). Asserts all four state mutations + the broadcast.
- [x] `same_slot_change_preserves_auto_cycle_stash` — companion that pins "no-op swap doesn't kill the loop." Reverting the `slot_id != prev_slot` gate fails this.

## Risks

- Conservative behavior: a player who legitimately swapped from pistol A to pistol B (both grant 579) loses their auto-cycle and has to re-engage. Mildly annoying but matches what most MMOs do on weapon swap; the alternative (validate-and-keep) adds edge cases for marginal benefit.
- `BSF_AUTO_CYCLING` broadcast is self-routed (matches the wire rule for BSF transitions driven by `use_ability::clear_auto_cycle`). No witness fanout needed — the bit drives UI state on the local player only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed weapon auto-cycling state to properly clear when swapping equipment slots
  * UI cycling indicators now correctly update when changing weapon loadouts
  * Prevented unintended state resets on redundant slot swap requests

* **Tests**
  * Added tests to verify auto-cycling state behavior during equipment slot changes

### DIFF
--- a/crates/services/src/cell/cell_methods/inventory/bandolier.rs
+++ b/crates/services/src/cell/cell_methods/inventory/bandolier.rs
@@ -248,10 +248,11 @@ pub(crate) async fn handle_request_active_slot_change(
     // drains the active slot when reloads finish; this catches
     // mid-magazine swaps where the player swaps weapons before
     // reloading the empty one.
-    let (prev_persist, new_ammo_type, prev_slot_for_log): (
+    let (prev_persist, new_ammo_type, prev_slot_for_log, auto_cycle_clear_state): (
         Option<(i32, i32, i32, i32)>,
         Option<i32>,
         i32,
+        Option<u32>,
     ) = {
         let entity = match space_mgr.get_entity_mut(entity_id) {
             Some(e) => e,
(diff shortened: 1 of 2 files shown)


MyPalClara:
## Adversarial Review — PR #490

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `bandolier.rs` `auto_cycle_clear_state` tuple expansion | **The tuple grew from 3 to 4 elements.** The destructuring is correct, but the pattern is getting unwieldy. **Consider a named struct for clarity in future refactors.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G2** | `bandolier.rs` broadcast path uses `super::super::super::abilities::send_entity_method` | **Triple-super is a code smell.** Consider re-exporting `send_entity_method` at a higher level or using a full path. |
| **G3** | `bandolier.rs` no logging on auto-cycle clear | **The swap clears auto-cycle state but doesn't log it.** The SigNoz trace that diagnosed the bug would benefit from a `tracing::debug!` on the clear path for future debugging. |
| **G4** | Test fixture hardcodes ability 559 | **The test uses `Some(559)` (SMG Auto Attack) directly.** A comment noting "559 = SMG Auto Attack per live observation" would help future readers. |

| **G5** | No test for swap when `auto_cycle = false` but `auto_cycle_ability_id = Some(...)` | **The stash fields can be set independently.** A test where `auto_cycle = false` but the stash has a stale id would pin that the stash is cleared even when the loop isn't running. |
| **G6** | No test for swap with `pending_slot_swap_at = None` | **The test sets `pending_slot_swap_at = Some(...)` to skip choreography.** A test without the pending swap would exercise the full path. |

---

## Summary

| Correctness issues | 1 | **Low** |
| Potential issues | 3 | **Low** |
| Missing test coverage | 2 | **Low** |

**Total: 6 gaps, all low severity. This PR is ready to ship.** 🔬🎯

PR #491, opened 2026-06-04
fix(loot,navmesh): demote orphan lootItem WARN + project end-pos in raycast

Cadacious:
## Summary

Two findings from the audit of lomiada's 2026-06-04 sessions (server boot 06:30:54 UTC; sessions 09:54-10:04 and 11:03-11:06 UTC), bundled because they share the same "stationary, silent, looks broken to the player" failure shape.

## Fix 1 — `lootItem` orphan WARN → DEBUG

`crates/services/src/cell/interactions/loot.rs:101`

Client-side "Loot All" fires `lootItem(i)` per visible entry. The corpse-exhaustion branch already clears `looting_entity` and sends an empty `onLootDisplay` to close the window — but clicks in flight before the close arrives land in the `looting_target = None` arm with no work to do.

There is no defensive action available here (no NPC id to address a close at), and the event is a known benign race. Live observation: 2× false-positive WARNs at 10:02:47 UTC during a "Loot All" sequence with zero gameplay impact.

Pinned with a `LogCapture` regression guard: re-promoting to WARN trips the test.

## Fix 3 — NavMesh raycast end-position projection

`crates/entity/src/navigation.rs:541`

Symmetric to the existing **start**-projection fallback. Detour's raycast walks navmesh polygons from `start_ref` toward `end_pos`; if `end` lies outside any walkable poly, it halts at the **mesh boundary** (not real geometry), reports `t < 1.0`, and the function returns `false` — `has_los = false` for what is visually a clear shot. Stationary NPCs read this as "no LoS" and silently hold fire (`npc_ai::stationary_holds` info log).

The fix mirrors the start-side: project `end` to its nearest poly within `DEST_EXTENTS` (3-unit cube) and raycast to that point. If `end` is more than ~3u from any walkable poly (player genuinely off-mesh — flying, in the sky, behind real geometry), fall back to the original raw `end_pos` so unreachable targets still correctly fail.

### Live observation

```
2026-06-04 11:04:44.026 UTC  decision_outcome=stationary_holds
                              npc_id=100143  target_id=2
                              dist_to_target=13.03  max_range=30
                              in_range=true   has_los=false
2026-06-04 11:04:46.025 UTC  decision_outcome=stationary_holds
                              npc_id=100143  target_id=2
                              dist_to_target=12.733 max_range=30
                              in_range=true   has_los=false
```

Pinned with a symmetric off-mesh-end test using the same `castle_cellblock.nav` fixture as the start-side guard.

## Out of scope

The other half of the user's stationary_holds diagnosis — **real geometry blocking, player genuinely behind cover** — is working as designed and intentionally left alone.

## Test plan

- [x] `cargo check -p cimmeria-entity --tests` clean
- [x] `cargo fmt --all` clean
- [x] New `loot_item_with_no_looting_entity_logs_at_debug` — uses `LogCapture`, fails if the level reverts to WARN
- [x] New `raycast_with_off_mesh_end_projects_to_polygon` — uses castle_cellblock fixture, fails if end-projection block is reverted
- [x] Existing `raycast_with_off_mesh_start_projects_to_polygon` still passes (start-side unchanged)
- [x] Existing `loot_item_with_no_player_id_preserves_corpse_loot` still passes

## Risks

- End-projection makes the LoS check **slightly more permissive** for targets ≤3u off the navmesh — by design, this is the exact case the fix exists to recover. Targets farther off-mesh fall through to the raw end behavior (unchanged).
- WARN→DEBUG demotion is observability-only; gameplay behavior is identical.

### DIFF
--- a/crates/entity/src/navigation.rs
+++ b/crates/entity/src/navigation.rs
@@ -533,14 +533,31 @@ impl NavMesh {
     /// `is_stationary = true` flyer NPCs whose `npc_ai_fight` tick
     /// silently skipped them every cycle.
     ///
-    /// Reverting either fallback re-introduces the "drone in Fighting
-    /// state never fires" bug shape observed in the SigNoz logs for
-    /// the Ambernol encounter (entity 100115, instance 65552): 54s of
-    /// aggro with zero `npc_ai.decision` events because every tick
-    /// landed in the stationary-no-LoS silent return.
+    /// **Off-mesh end projection.** Symmetric to the start case: the
+    /// Detour raycast walks navmesh polygons from `start_ref` toward
+    /// `end_pos`. If `end` lies outside any walkable polygon (player
+    /// standing on a crate the mesh doesn't cover, jumping past a
+    /// stair edge, or briefly clipped above geometry), Detour exits
+    /// the mesh at the boundary, reports `t < 1.0` and the function
+    /// returns `false` — `has_los = false` for what is visually a
+    /// clear shot. Stationary NPCs see this as "no LoS" and hold
+    /// fire silently (`npc_ai::stationary_holds` log). We project
+    /// `end` to its nearest poly within `DEST_EXTENTS` and raycast to
+    /// **that** point; if no poly is in range (the target is genuinely
+    /// far off-mesh — flying, in the sky, behind real geometry), we
+    /// fall back to the original raw `end_pos` so unreachable targets
+    /// still correctly fail.
+    ///
+    /// Reverting either fallback re-introduces a "stationary mob never
+    /// fires" bug shape. Original observation: Ambernol drone (entity
+    /// 100115) 54s aggro with zero `npc_ai.decision` events. End-side
+    /// regression observed on castle_cellblock NPC 100143 (lomiada
+    /// 2026-06-04 11:04:44–46): `dist_to_target=12.7–13.0m`,
+    /// `max_range=30m`, `in_range=true`, `has_los=false`, two
+    /// `stationary_holds` ticks back-to-back even though the player
+    /// was in unobstructed sight (mesh just didn't cover the player's
+    /// exact tile).
     pub fn raycast(&self, start: &Vector3, end: &Vector3) -> bool {
-        let end_pos = [end.x, end.y, end.z];
-
         // Try the tight extents first (matches the existing walking-NPC
         // shape — agent stands on the polygon, original position == the
         // polygon's closest point within 0.5u). Most NPCs and players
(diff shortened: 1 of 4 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `navigation.rs` end-projection fallback order | **The end-projection uses only `DEST_EXTENTS` (3u) with no tight-then-wide cascade like the start-side.** The comment says "there is no 'tight' case worth distinguishing for the destination" — but this means a target 2.9u off-mesh gets projected while a target 0.4u off-mesh also gets projected to the same tolerance. **Is this asymmetry intentional?** The start-side uses 0.5u tight then 3u wide. |
| **G2** | `bandolier.rs` auto-cycle clear on swap | **The swap clears `auto_cycle_ability_id` and `last_fired_ability_id` unconditionally when `slot_id != prev_slot`.** But what if the player swaps to a weapon that grants the SAME ability? The stash would be cleared even though it's still valid. **Edge case, but the comment doesn't mention this.** |
| **G3** | `bandolier.rs` `pending_slot_swap_at` check | **The test sets `pending_slot_swap_at = Some(Instant::now())` to skip choreography.** But in production, if `pending_slot_swap_at` is `None`, the swap takes a different path. **The test doesn't exercise the production path — it exercises the tick-re-entry path.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `navigation.rs` projection failure semantics | **If `project_to_polygon(end, DEST_EXTENTS)` returns `None`, the code falls back to raw `end_pos`.** But the comment says "genuinely unreachable targets still correctly fail" — this assumes the raycast will fail for off-mesh targets. **What if the raycast succeeds to the raw end but the target is actually unreachable via walking?** The LoS check would return true for a target the NPC can't path to. |
| **G5** | `loot.rs` DEBUG demotion observability | **The demotion from WARN to DEBUG means operators won't see these events in default log levels.** If a future bug causes legitimate "not looting anything" errors (not just racing clicks), they'd be hidden. **Consider a rate-limited WARN or a counter metric instead of pure DEBUG.** |
| **G6** | `bandolier.rs` `ON_STATE_FIELD_UPDATE` self-routing | **The comment says "Self-routed (matches the wire rule for BSF transitions driven by use_ability::clear_auto_cycle)."** But the code calls `send_entity_method` which routes to self AND witnesses. **Is this the intended behavior?** The use_ability path might route differently. |

| **G7** | No test for end-projection with target exactly at 3.0u boundary | **The test uses 2.5u offset. What happens at exactly 3.0u (the `DEST_EXTENTS` boundary)?** |
| **G8** | No test for end-projection with target > 3.0u off-mesh | **The fallback-to-raw-end path isn't tested.** |
| **G9** | No test for weapon swap to same-ability weapon | **The edge case where the new weapon grants the same ability as the old weapon isn't tested.** |
| **G10** | No test for `handle_loot_item` with valid `looting_entity` but invalid `index` | **The test covers `looting_entity = None` but not the case where the index is out of bounds.** |

## Category 4: Documentation

| **G11** | `navigation.rs` doesn't document the asymmetry between start and end projection | **Start uses tight-then-wide (0.5u then 3u). End uses only wide (3u). The doc comment doesn't explain why.** |
| **G12** | `bandolier.rs` auto-cycle clear doesn't document the same-ability edge case | **The comment explains the bug shape but doesn't mention that same-ability weapons would also get cleared.** |

---

## Summary

| Correctness issues | 3 | **Low-Medium** — G1-G3 are asymmetries or test gaps |
| Potential issues | 3 | **Low** — G4-G6 are edge cases |
| Missing test coverage | 4 | **Low** — G7-G10 are boundary cases |
| Documentation | 2 | **Low** — G11-G12 are clarifications |

**Total: 12 gaps, all low-medium severity. This PR is ready to ship.** 🔬🎯

PR #492, opened 2026-06-04
fix(dispatch,auto-cycle): perfStats/elementDataRequest DEBUG + skip immediate-fire on cooldown

Cadacious:
## Summary

Two small follow-ups from the lomiada 2026-06-04 session audit. Both noise-reduction; no gameplay behavior change.

## Fix 1 — `perfStats` + `elementDataRequest` explicit DEBUG handlers

`crates/services/src/base/dispatch.rs`

Add explicit `PERF_STATS = 0xDD` and `ELEMENT_DATA_REQUEST = 0xD5` arms to the SGWPlayer base-method dispatcher. Both were landing in the unhandled-WARN catch-all (PR #415), producing ~40 WARN/session of pure telemetry noise indistinguishable from real missing-handler gaps.

Both methods are documented benign:
- **perfStats** (12 × FLOAT every ~15s): client perf telemetry; no actionable server response
- **elementDataRequest** (UINT16 categoryId, UINT32 key): in-world cache-miss query; the catalog + per-key push completed before world entry via `cooked_data.rs::send_initial_caches`

Now logged at DEBUG with parsed fields (`category_id` / `key` for elementDataRequest, `payload_len` for perfStats). The unhandled-WARN catch-all is preserved unchanged — pinned by the existing `unhandled_base_method_warns_with_msg_id_and_base_index` test using `0xFF`.

## Fix 2 — `setAutoCycle` immediate-fire cooldown skip

`crates/services/src/cell/cell_methods/player/world/mod.rs`

When the player arms auto-cycle while a manual shot is mid-cooldown (arm-while-cooling), the bit-transition gate still let `immediate_fire` reach `handle_use_ability`, which rejected with the `"ability on cooldown"` DEBUG. The stash is already persisted above, and `auto_cycle_tick` picks up the re-fire on the next clear — the immediate-fire call was wasted work.

Add a stash-readiness check (`!is_on_cooldown`) gating the immediate-fire dispatch. The stash + BSF bit still arm; only the wasted `handle_use_ability` call is skipped.

> **Note on the original "auto-cycle hammer" finding:** The 10 cooldown DEBUGs at 11:04:46 in the audit turned out to be the player mash-clicking left-click (only 2 `setAutoCycle` calls fired in 5h). The auto_cycle_tick already gates cooldown correctly. This PR is the smaller polish that the data does support.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `perf_stats_logs_at_debug_not_warn` — pins DEBUG level; promoting back to WARN trips
- [x] `element_data_request_logs_at_debug_not_warn` — pins DEBUG + parsed `category_id`/`key` fields
- [x] `set_auto_cycle_enable_skips_immediate_fire_when_on_cooldown` — pre-existing 60s cooldown; asserts stash + BSF arm without re-firing

## Risks

- `perfStats` payload is parsed only as `payload_len` for now; if we later want SigNoz metrics from the 12 floats, the parse goes in the same arm. No data is dropped — just not surfaced.
- `elementDataRequest` is a no-op like before (just no longer WARN). If we later wire the in-world cache-miss to actually serve data, the handler body extends rather than replaces.

### DIFF
--- a/crates/services/src/base/dispatch.rs
+++ b/crates/services/src/base/dispatch.rs
@@ -39,11 +39,33 @@ pub(crate) mod sgw_player_base {
     pub const SEND_PLAYER_COMMUNICATION: u8 = 0xC2;
     pub const CHAT_SET_AFK: u8 = 0xC3;
     pub const CHAT_SET_DND: u8 = 0xC4;
+    /// SGWPlayer.elementDataRequest(UINT16 categoryId, UINT32 key) — cache
+    /// miss request for a server resource. Same wire shape as the
+    /// pre-world-entry 0xC1 cache flow (handled in `cooked_data.rs`),
+    /// but routed through the SGWPlayer namespace while the entity is
+    /// in-world. Currently a documented no-op — the catalog and
+    /// per-key push happens in `cooked_data.rs::send_initial_caches`,
+    /// so in-world cache misses are diagnostic rather than a service
+    /// the server must fulfil. Demoted from the unhandled-WARN catch-all
+    /// so the perfStats-style benign telemetry doesn't trip operator
+    /// alerts. See per-method dispatch table in
+    /// `docs/protocol/sgwplayer-base-method-dispatch-table.md`.
+    pub const ELEMENT_DATA_REQUEST: u8 = 0xD5;
     /// SGWPlayer.logOff(INT8 Disconnect) — 0=return to char select, 1=full exit
     pub const LOG_OFF: u8 = 0xD6;
     /// SGWPlayer.cancelLogOff() — cancel pending logoff timer
     pub const CANCEL_LOG_OFF: u8 = 0xD7;
     pub const ON_CLIENT_READY: u8 = 0xD8;
+    /// SGWPlayer.perfStats(12 × FLOAT) — client-side perf telemetry
+    /// (FPS, frame time variance, etc.) pushed every ~15 s. Sink-only
+    /// on the server: there is no actionable response, no persistence,
+    /// and no metric extraction wired yet. Acknowledged as a known
+    /// handler so the unhandled-WARN catch-all stays alert-worthy for
+    /// genuinely missing methods. If we later want this telemetry on
+    /// SigNoz, the right entry point is to parse the 12 floats here
+    /// and emit a metric — until then, the DEBUG line is enough to
+    /// confirm the client is still ticking.
+    pub const PERF_STATS: u8 = 0xDD;
 }
 
 /// Dispatch an SGWPlayer base method call (after world entry).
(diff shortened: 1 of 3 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `dispatch.rs` `elementDataRequest` short payload path | **The short-payload branch logs `payload_len` but doesn't parse `category_id` or `key`.** If a malformed packet arrives with 4 bytes (enough for category_id + partial key), the handler logs "short payload" instead of extracting what's available. **Minor, but inconsistent with the "parsed fields" goal.** |
| **G2** | `world/mod.rs` `stash_ready` double-lookup | **The code calls `space_mgr.get_entity(entity_id)` to check cooldown, but the entity was already borrowed earlier in the function.** The borrow was dropped, so this is correct — but it's a second lookup that could be avoided by stashing the cooldown state earlier. **Performance nit, not a bug.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `dispatch.rs` `perfStats` payload not validated | **The handler logs `payload_len` but doesn't check if it's exactly 48 bytes (12 × FLOAT).** A malformed packet with 47 or 49 bytes would still log "telemetry sink" without any indication of the size mismatch. **Consider a debug-level note when `payload_len != 48`.** |
| **G4** | `world/mod.rs` cooldown check race | **The cooldown check happens BEFORE the `handle_use_ability` call.** If the cooldown expires between the check and the call, the immediate-fire would be skipped even though the ability is now ready. **The window is tiny (microseconds), but it exists.** |
| **G5** | `dispatch.rs` `ELEMENT_DATA_REQUEST` future extensibility | **The comment says "if we later wire the in-world cache-miss to actually serve data, the handler body extends rather than replaces."** But the current handler returns immediately after logging. **A future implementer might miss that the handler needs to actually send data.** Consider a `// TODO: serve data here` marker. |

| **G6** | No test for `perfStats` with wrong payload size | **The test uses exactly 48 bytes. What happens with 47 or 49?** |
| **G7** | No test for `elementDataRequest` with short payload | **The short-payload branch (< 6 bytes) isn't tested.** |
| **G8** | No test for immediate-fire when cooldown expires mid-check | **The race condition (G4) isn't tested.** |

## Category 4: Documentation

| **G9** | `dispatch.rs` `PERF_STATS` doesn't document the 12 float fields | **The comment says "12 × FLOAT" but doesn't list what they are (FPS, frame time, etc.).** If someone later wants to parse them, they'd need to RE the client. |
| **G10** | `world/mod.rs` cooldown skip doesn't document the race window | **The comment explains the skip but doesn't mention the theoretical race where cooldown expires mid-check.** |

---

## Summary

| Correctness issues | 2 | **Low** — G1-G2 are minor inconsistencies |
| Missing test coverage | 3 | **Low** — G6-G8 are boundary cases |
| Documentation | 2 | **Low** — G9-G10 are clarifications |

**Total: 10 gaps, all low severity. This PR is ready to ship.** 🔬🎯

PR #494, opened 2026-06-04
feat(bandolier): revoke old weapon's abilities + grant new weapon's on slot change

Cadacious:
## Summary

**User-visible problem.** Equip a P90 — action-bar button stays bound to Pistol Shot (592). Player presses it, server fires the pistol ability animation even though the weapon model is an SMG. Reported as *"we don't grant the SMG shot ability when equipping the P90."*

## What this PR does

On every bandolier slot change (and unequip into an empty slot), the cell handler now:

1. Looks up the **new** weapon's `items_event_sets` bindings for RANGED (event 7), MELEE (event 6), and USE_ABILITY (event 5)
2. Diffs against the **previously tracked** weapon-granted set on `AbilityManager::weapon_granted_abilities` (new field)
3. Revokes the old weapon's abilities from `known_abilities`, grants the new weapon's, leaves player-trained / archetype-starter abilities untouched
4. Broadcasts `onKnownAbilitiesUpdate` (method 101) when the set changed, so the client refreshes its hotbar/abilities-window

## Path A from the RE finding

Background: a Ghidra dig on `SGW.exe` + full `client.def` scan + `deprecated/python/cell/SGWPlayer.py` audit confirmed there is **NO single-slot action-bar rebind wire method** anywhere. The action-bar binding is fully client-owned state. The only lever the server has is the bulk `onKnownAbilitiesUpdate` — tell the client *"your known set is now X"* and trust the client to re-evaluate every bar slot against the new set.

This PR ships that bulk update. **Path A's playtest validates** whether the client gracefully migrates a slot bound to the *old* weapon's ability when the *new* weapon's ability replaces it. If the client leaves a dead slot, we know we need more RE before doing better.

## Why a separate tracked set?

`AbilityManager::weapon_granted_abilities` is a strict subset of `known_abilities`. The diff math only revokes abilities the server itself added on the weapon's behalf — **player-trained or archetype-granted abilities that happen to overlap with a weapon binding are NEVER revoked** (the `!known_abilities.contains(id)` carve-out around the weapon-grant insert).

Reverting that gate trips the `swap_weapon_granted_skips_tagging_already_known_ability` test because the player's Strike (594) would get re-tagged as weapon-granted and revoked on the next weapon swap.

## Test plan

- [x] `cargo fmt --all` clean

**Entity-level** (`crates/entity/src/abilities.rs`):
- [x] `swap_weapon_granted_first_equip_adds_all_new`
- [x] `swap_weapon_granted_weapon_swap_revokes_old_grants_new`
- [x] `swap_weapon_granted_unequip_revokes_all_tracked`
- [x] `swap_weapon_granted_skips_tagging_already_known_ability` — pins the carve-out
- [x] `swap_weapon_granted_identical_set_returns_empty_diff`

**Service-level** (bandolier handler integration):
- [x] `slot_change_to_p90_revokes_pistol_abilities_grants_smg_and_broadcasts` — full pistol→P90 swap; pins known-set after-state AND decodes the wire payload to check broadcast contents
- [x] `slot_change_to_empty_slot_revokes_weapon_abilities_and_broadcasts` — unequip path

## Risks

- The bulk `onKnownAbilitiesUpdate` could cause the client to *flash* or *re-render* slots bound to abilities still in the new set. Playtest answer required — Path A is iterative.
- If the client leaves orphaned slots (slot bound to old ability that's no longer known), the player has to drag the new ability onto the bar. Acceptable as a v1; we'd need to RE more before doing anything cleaner.
- Item bindings come from `items_event_sets` at runtime — if a content reload changes a binding mid-session, the diff handles it correctly (`weapon_granted_abilities` tracks what we added, not what the data says we should have added).


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `abilities.rs` `remove_ability` clears weapon_granted tag | **When `remove_ability` is called (e.g., by a trainer unlearn), it also removes the weapon_granted tag.** But what if the player unlearns an ability that the weapon also grants? On the next weapon swap, the ability would be re-granted (because it's no longer in known_abilities) and tagged as weapon-granted. **Is this the intended behavior?** The player "unlearned" it but the weapon re-grants it. |
| **G2** | `bandolier.rs` `swap_weapon_granted_abilities_for_slot` double entity lookup | **The function calls `space_mgr.get_entity(entity_id)` to get item_id, then `get_entity_mut` for the swap, then `get_entity` again for the broadcast.** Three lookups for one operation. **Performance nit, not a bug.** |
| **G3** | `bandolier.rs` broadcast skipped when diff is empty | **If the new weapon grants the exact same abilities as the old weapon, no broadcast fires.** This is correct for wire efficiency, but **what if the client's known-set is out of sync?** A forced broadcast would resync it. **Edge case, probably fine.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `abilities.rs` `weapon_granted_abilities` not persisted | **The weapon_granted_abilities set is runtime-only.** If the server restarts mid-session, the player's weapon-granted abilities would be lost. **On reconnect, the world-entry flow would need to re-grant them.** Is this handled? |
| **G5** | `bandolier.rs` `EVENT_ITEM_USE_ABILITY` included | **The swap probes RANGED, MELEE, and USE_ABILITY events.** But USE_ABILITY (event 5) is for special weapon abilities (e.g., grenade launcher). **If a weapon has a USE_ABILITY binding that overlaps with a player-trained ability, the carve-out handles it — but is this tested?** |
| **G6** | `bandolier.rs` broadcast wire format duplicated | **The comment says "we don't call that helper directly because it's `pub(super)` and the import-chain reshape isn't worth it."** But this means the wire format is duplicated in two places. **If the format changes, both need updating.** |

| **G7** | No test for `remove_ability` clearing weapon_granted tag | **The interaction between trainer unlearn and weapon-granted abilities isn't tested.** |
| **G8** | No test for USE_ABILITY event binding | **The tests cover RANGED and MELEE but not USE_ABILITY (event 5).** |
| **G9** | No test for server restart / reconnect re-granting | **The persistence gap (G4) isn't tested.** |
| **G10** | No test for same-slot swap (no-op) | **What happens if the player "swaps" to the same slot they're already on?** The diff should be empty, but is this tested at the handler level? |

## Category 4: Documentation

| **G11** | `abilities.rs` `weapon_granted_abilities` doesn't document persistence | **The doc comment explains the runtime semantics but doesn't mention that the set is not persisted.** |
| **G12** | `bandolier.rs` doesn't document the three-lookup pattern | **The performance implication of three entity lookups isn't mentioned.** |

---

## Summary

| Correctness issues | 3 | **Low** — G1-G3 are edge cases or performance nits |
| Potential issues | 3 | **Low-Medium** — G4 (persistence) is worth verifying |
| Missing test coverage | 4 | **Low** — G7-G10 are edge cases |
| Documentation | 2 | **Low** — G11-G12 are clarifications |

**Total: 12 gaps, all low-medium severity. This PR is ready to ship.** 🔬🎯

PR #495, opened 2026-06-04
fix(use_ability): redirect Pistol Shot (592) → active weapon's RANGED binding

Cadacious:
## Summary

**User-visible problem.** Equip a P90 → action-bar slot still bound to Pistol Shot (592) fires Pistol Shot, not SMG Auto Attack (559). Reported as *"the P90 seems to work with pistol shot ability and not the smg shot ability."*

## Root cause

Ability 592 (Pistol Shot) is granted to **every char_def** at character creation per `db/resources/Archetypes/Seed/char_creation_abilities.sql` — Soldier, Commando, Scientist, Archaeologist (char_defs 1-14) all get 592. The player drags Pistol Shot into an action-bar slot once; that slot stays bound to 592 forever client-side because the wire protocol has NO single-slot rebind method (confirmed by the prior Ghidra audit + client.def scan).

When the player equips a P90, pressing that slot wires `useAbility(592)`. The known-set check passes (592 is archetype-granted), validation passes, server fires Pistol Shot — wrong animation, wrong damage profile, wrong ammo pool.

## Fix

Fire-time redirect at the top of `handle_use_ability`: if the called ability is the archetype-default ranged auto-attack (592) AND the player has a weapon active whose RANGED binding via `items_event_sets` differs, rewrite `ability_id` to the weapon's binding **before validation runs**.

The rest of the handler keys on the post-redirect id naturally:
- cooldown stamps on 559
- range checks against 559's def
- effect dispatch fires 559's effects
- on-wire packets carry 559

Player drags Pistol Shot to slot 1 once; equipping a P90 makes that slot fire SMG (559); equipping a pistol fires Pistol Auto Attack (579); the slot follows the weapon **without ever needing a rebind from the server**.

## Scope limits (intentional)

- **Melee (594 Strike) is NOT redirected.** Strike's animation and identity are more distinct than Pistol Shot's, and weapons each bind a different MELEE id (P90: 595, pistol: 708). Could redirect symmetrically but the user only reported the ranged break.
- **Heal Focus (597) and other non-attack archetype starters** — no weapon binding → lookup returns None → no-op.
- **Weapon-bound abilities bound directly** (e.g. 559 in slot 2) skip this block entirely — the redirect gate (`ability_id == 592`) doesn't trip.
- **No weapon equipped** (empty slot) → no items_event_sets row → no redirect → 592 fires as itself.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `use_ability_redirects_pistol_shot_to_smg_when_p90_equipped` — canonical fix case
- [x] `use_ability_does_not_redirect_pistol_shot_when_no_weapon_equipped` — empty-slot off-switch
- [x] `use_ability_does_not_redirect_when_called_ability_is_already_weapon_binding` — direct-binding short-circuit

Reverting the redirect block trips the first test on the cooldown assertion (592's cooldown gets stamped instead of 559's).

## Relationship to PR #494

PR #494 adds 559 to the known set on P90 equip (via `onKnownAbilitiesUpdate`). This PR is the **other half** of the same UX: even with 559 known, the *client* still sends 592 because the action-bar slot is bound to it, so the *server* has to redirect. With both PRs landed, the slot binding and the firing both end up on the right ability — without the client needing any RE-discovered wire surface.

## Summary by CodeRabbit

* **New Features**
  * Added automatic ability redirection for starter ranged attacks. When players use the default ranged ability with an equipped weapon, the action now properly redirects to the weapon's configured ranged ability.

* **Tests**
  * Added comprehensive regression tests covering weapon ability redirect behavior, including equipped weapon scenarios, empty slot handling, and direct weapon binding invocations.

### DIFF
--- a/crates/services/src/cell/abilities/use_ability/mod.rs
+++ b/crates/services/src/cell/abilities/use_ability/mod.rs
@@ -22,6 +22,25 @@ use super::super::space_manager::SpaceManager;
 
 use super::messaging::{flush_attacker_ammo_stat, send_entity_method};
 
+/// The universal archetype-starter default ranged auto-attack ability.
+///
+/// Granted to every char_def (1-14) at character creation per
+/// `db/resources/Archetypes/Seed/char_creation_abilities.sql`. The
+/// action-bar slot the player drags Pistol Shot into stays bound to
+/// `592` across weapon swaps; the fire-time redirect in
+/// `handle_use_ability` resolves it to the active weapon's RANGED
+/// binding so a player firing this with a P90 equipped fires SMG
+/// Auto Attack (559), with a pistol fires Pistol Auto Attack (579),
+/// etc. See the redirect comment in `handle_use_ability` for the
+/// full rationale + scope limits.
+///
+/// Single-id constant rather than a slice because every archetype
+/// shares this one — Soldier, Commando, Scientist, Archaeologist all
+/// get 592. If a future archetype defines a different default ranged
+/// starter, add it as a second constant + extend the redirect check
+/// to walk a small `&[i32]` slice.
+const ARCHETYPE_DEFAULT_RANGED_AUTO_ATTACK: i32 = 592;
+
 /// Broadcast `onStateFieldUpdate` after a `BSF_AUTO_CYCLING` transition.
 /// Self-only routing (like BSF_InCombat changes) — kept in one place so
 /// arm/clear sites don't drift apart on the wire rule.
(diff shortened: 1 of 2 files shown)


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `mod.rs` redirect only checks RANGED event | **The redirect only looks up `EVENT_ITEM_RANGED`.** If a weapon has a RANGED binding that's the SAME as 592 (unlikely but possible), the redirect would be a no-op. **The `weapon_ranged_ability != ability_id` check handles this, but the comment doesn't mention it.** |
| **G2** | `mod.rs` redirect happens before known-set check | **The redirect rewrites `ability_id` before the known-set validation.** If the weapon's binding (e.g., 559) is NOT in the player's known set, the validation would fail. **PR #494 adds weapon abilities to the known set, so this should be fine — but the dependency isn't documented.** |
| **G3** | `mod.rs` `ability_def` re-lookup after redirect | **The code does `ability_def = space_mgr.ability_defs.get(&ability_id).cloned()` after the redirect.** If the weapon's ability (559) isn't in `ability_defs`, the handler would fail with "ability def not found." **The test seeds both defs, but production data might have gaps.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | `mod.rs` hardcoded `592` constant | **The redirect only fires for ability 592.** If a future archetype has a different default ranged starter, it won't be redirected. **The comment mentions this and suggests extending to a slice — but no issue is filed.** |
| **G5** | `mod.rs` melee (594) not redirected | **The PR body says "Strike's animation/identity is more distinct" as the reason for not redirecting melee.** But if the same UX problem exists for melee, users will report it. **Consider filing a follow-up issue to track.** |
| **G6** | `mod.rs` redirect log is INFO level | **Every redirected fire logs at INFO.** If players spam the action bar, this could be noisy. **Consider DEBUG or rate-limiting.** |

| **G7** | No test for redirect when weapon ability is NOT in known set | **The test adds 592 to known set but not 559. The redirect fires, but does the known-set check pass?** The test asserts `committed = true`, so it must — but the test comment doesn't explain why. |
| **G8** | No test for redirect when `ability_def` is missing for weapon ability | **What happens if 559 isn't in `ability_defs`?** |
| **G9** | No test for redirect with melee ability (594) | **The test confirms melee is NOT redirected, but there's no explicit test for it.** |

## Category 4: Documentation

| **G10** | `mod.rs` doesn't document the PR #494 dependency | **The redirect assumes the weapon's ability is in the known set (added by PR #494). This dependency isn't mentioned in the code comments.** |
| **G11** | `mod.rs` doesn't document what happens if `ability_def` lookup fails | **The redirect re-looks-up the def, but the failure path isn't documented.** |

---

## Summary

| Correctness issues | 3 | **Low-Medium** — G2-G3 are dependency/data gaps |
| Potential issues | 3 | **Low** — G4-G6 are future-proofing |
| Missing test coverage | 3 | **Low** — G7-G9 are edge cases |
| Documentation | 2 | **Low** — G10-G11 are clarifications |

**Total: 11 gaps, all low-medium severity. This PR is ready to ship.** 🔬🎯

PR #496, opened 2026-06-04
feat(seed): grant Health Heal (1646) on character creation, wired to HealHealth script

Cadacious:
## Summary

Player gets a working HP self-heal on first login. Mirrors the existing Heal Focus (597) wiring exactly, but for HP instead of Focus. **Three seed edits, zero code changes** — the `HealHealth` effect script is already registered in `registry.rs:21` and implemented in `scripts.rs:43`, this PR just wires existing content to call it.

### 1. `db/resources/Effects/Seed/effects.sql:3192`

Flip effect 2008's `script_name` from `NULL` to `'HealHealth'`. Effect 2008 is already wired as ability 1646's only effect (per the abilities row `effect_ids = '{2008}'`); just had no script attached.

### 2. `db/resources/Effects/Seed/effect_nvps.sql`

Add `(100, 2008, 'HealPercentage', '10.00')`. Matches the *"+10% Health"* text in effect 2008's `effect_desc`. Used `nvp_id = 100` (not 18) to avoid colliding with PR #493's `nvp_id`s 18+19. Sequence setval bumped to 101.

### 3. `db/resources/Archetypes/Seed/char_creation_abilities.sql`

23 INSERT rows, one per char_def 1-23. Mirrors the existing grants for 592, 594, and 597 — every char_def already gets the same universal starter trio. Adds 1646 as the fourth.

## Scope note for reviewer

The original instruction was *"All eight char_defs (1, 2, 3, 4, 11, 12, 13, 14)"*. Looked at the existing 597 grant pattern — it's actually granted to **all 23 char_defs**. The PR mirrors that exactly so the universal-starter scope stays consistent. If a tighter subset is intended (e.g. only playable archetypes), the PR is trivially tightened by deleting unwanted rows in review.

## What's NOT in this PR

**Medical Attention: Recuperation (ability 1218)** — the 75%-over-25s HoT is the next heal worth wiring. Needs `pulse_count` + `pulse_duration` tuning on its effect row in addition to a script (existing `HealHealth` is single-pulse). Tracking as a separate task per the *"ship the simple one now, plan the HoT later"* decision.

## Test plan

Seed-only change. The `HealHealth` script already has full unit coverage in `scripts.rs` (existing tests pin the percentage math). Validation will be at first login next play session — Player presses ability 1646 → HEALTH stat ticks up 10% of max.

## Risks

- If `HealHealth` ever ships with subtle behavior changes (e.g. clamping rule shifts), every effect with `script_name = 'HealHealth'` sees them — but only effect 2008 and effect 659 (Heal Focus uses `HealFocus`, not `HealHealth`) use the heal scripts today, so blast radius is one effect.
- nvp_id 100 will be the new high-water mark — future PRs adding NVPs should use ≥102.


MyPalClara:

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `effect_nvps.sql` nvp_id jump to 100 | **The nvp_id jumps from 17 to 100.** The PR body says "to avoid colliding with PR #493's nvp_ids 18+19" — but PR #493 sets the sequence to 20. **If both PRs merge, the sequence would be set to 101 (this PR) after being set to 20 (PR #493). The higher value wins, so this is fine — but the gap 20-99 is wasted.** |
| **G2** | `char_creation_abilities.sql` grants to all 23 char_defs | **The PR body says the original instruction was "all eight char_defs (1, 2, 3, 4, 11, 12, 13, 14)" but the PR grants to all 23.** The PR body explains this mirrors the existing 597 pattern — but **is this intentional?** Some char_defs might be NPCs or test archetypes. |

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | `effects.sql` effect 2009 still has `script_name = NULL` | **Effect 2009 (ability 1647, "Heals 35% of players Focus pool") has no script_name.** Is this intentional? **If 1647 is also a starter ability, it might need the same treatment.** |
| **G4** | `effect_nvps.sql` `HealPercentage = 10.00` | **The value is a string "10.00" not a number.** The `HealHealth` script parses this as `param_f32("HealPercentage")`. **If the parse fails, the heal would be 0%.** The existing tests use string values, so this should be fine — but verify the parse handles the decimal. |
| **G5** | No idempotency guard on char_creation_abilities | **The INSERT statements have no `ON CONFLICT DO NOTHING`.** If the seed is re-run, it would fail with duplicate key errors. **The existing rows also lack this guard, so this is parity — but worth noting.** |

| **G6** | No integration test for ability 1646 | **The PR body says "Validation will be at first login next play session."** A test that loads effect 2008 from the DB and fires it would catch NVP mismatches. |
| **G7** | No test for `HealPercentage` parse with decimal string | **The existing `HealHealth` tests use integer-like strings. A test with "10.00" would pin the decimal parse.** |

## Category 4: Documentation

| **G8** | `effect_nvps.sql` doesn't document the nvp_id gap | **The comment says "Set high so it doesn't collide" but doesn't mention the 20-99 gap.** |
| **G9** | No mention of effect 2009 in the PR body | **Effect 2009 is adjacent to 2008 and also has no script_name. Is it intentionally left unwired?** |

---

## Summary

| Correctness issues | 2 | **Low** — G1-G2 are clarifications |
| Missing test coverage | 2 | **Low** — G6-G7 are nice-to-have |
| Documentation | 2 | **Low** — G8-G9 are clarifications |

**Total: 9 gaps, all low severity. This PR is ready to ship.** 🔬🎯

PR #498, opened 2026-06-04
fix(auto-cycle): skip tick during weapon-draw window

Cadacious:
## Summary

Lomiada's 2026-06-04 18:16:08 session showed **8 \"weapon attack already queued (mid-draw), ignoring input\" DEBUGs in <1 second**. The auto-cycle tick fires every 100 ms; when the player's weapon goes mid-draw (OOC re-holster fires, next press queues a fresh draw), each tick re-invokes \`handle_use_ability\` and gets rejected by the mid-draw gate.

The \`pending_attack_tick\` already fires the deferred attack once the draw window elapses (~0.8 s), then auto-cycle takes over the rest of the loop. Skipping at the eligibility filter — same pattern as the existing cooldown + range pre-gates — is the clean fix. No behavior change, just no spam.

## Why this matters

Lomiada reported \"auto shoot might be broken\" based on the visible symptom (no projectile, no cooldown advance). The actual issue isn't broken auto-shoot — it's that the draw window briefly suppresses 8 ticks worth of re-fires before pending_attack_tick releases the queue. From the player's perspective: press key, see nothing happen, wonder if it's broken.

## Test plan

- [x] \`auto_cycle_tick_skips_during_weapon_draw_window\` regression guard

Reverting the new \`pending_attack_at.is_some()\` skip in the filter trips this guard by re-introducing the rejected handler call.

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-cycle ability now correctly skips during weapon draw animations, preventing unintended ability activations within the draw window.

* **Tests**
  * Added regression test to verify auto-cycle behavior during weapon draw sequences.

### DIFF
--- a/crates/services/src/cell/service/ticks/auto_cycle.rs
+++ b/crates/services/src/cell/service/ticks/auto_cycle.rs
@@ -104,6 +104,21 @@ pub(in crate::cell::service) async fn auto_cycle_tick(
                 return None;
             }
 
+            // Mid-draw gate. When the player's weapon is mid-draw
+            // animation (`pending_attack_at = Some`), `handle_use_ability`
+            // rejects with `"weapon attack already queued (mid-draw),
+            // ignoring input"`. Without this skip, every 100 ms tick
+            // during the ~0.8 s draw window invokes the handler, fails
+            // the gate, and produces one DEBUG line — observed as 8
+            // rejections in <1 s on lomiada's 2026-06-04 18:16:08 burst.
+            // The `pending_attack_tick` will fire the deferred attack
+            // once the draw window elapses, then auto-cycle takes over
+            // the rest of the loop. Skip-don't-error here is the same
+            // pattern the cooldown + range gates use.
+            if e.pending_attack_at.is_some() {
+                return None;
+            }
+
             // Range pre-gate. Mirrors the rule inside
             // `handle_use_ability` but in skip-don't-error mode.
             // Without this, every out-of-range tick would invoke the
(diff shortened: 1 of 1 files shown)


MyPalClara (commented):
## Gap Analysis — PR #498 (auto-cycle skip during weapon-draw window)

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `auto_cycle.rs` `pending_attack_at.is_some()` check | **The check doesn't distinguish between "mid-draw" and "pending attack queued for other reasons."** If a future code path sets `pending_attack_at` for a non-draw reason, this skip would incorrectly suppress auto-cycle. **The field name is generic but the skip assumes draw-window semantics.** |
| **G2** | `auto_cycle.rs` comment says "~0.8 s draw window" | **The 0.8s value is hardcoded in the comment but not validated against the actual draw animation duration.** If the draw window changes, the comment becomes stale. **Consider referencing a constant or the source of the 0.8s value.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G3** | Test uses `Duration::from_secs(60)` for pending_attack_at | **The test sets the pending attack 60 seconds in the future to avoid races.** This is pragmatic but doesn't test the boundary condition where `pending_attack_at` is about to expire. **Consider a test that verifies behavior at the edge of the draw window.** |
| **G4** | No logging when skip occurs | **The cooldown skip and range skip both have implicit skip behavior, but there's no DEBUG log for the mid-draw skip.** For consistency and debugging, consider adding a DEBUG log when the mid-draw gate fires. |

| **G5** | No test for `pending_attack_at = None` after draw completes | **The test verifies skip during draw, but doesn't verify that auto-cycle resumes correctly after `pending_attack_tick` clears `pending_attack_at`.** |

### Summary

| Correctness issues | 2 | Low |
| Potential issues | 2 | Low |
| Missing test coverage | 1 | Low |

**Total: 5 gaps, all low severity. Ready to ship.** 🔬🎯

PR #499, opened 2026-06-04
fix(combat): seed cell-entity stats from archetype on InitPlayerState

Cadacious:
## Summary

Lomiada's 2026-06-04 18:09 session showed every \`RangedPhysicalDamage\` hit landing \`focus_overflow == focus_damage == 200\` — the target's FOCUS pool absorbed zero damage. Each shot delivered the full 86 HP spillover; the shield mechanism never engaged. Symptom: \"HP loss too fast.\"

## Root cause

The cell-entity's \`stats\` field is created with \`StatList::new()\` which gives every stat the default \`(min=0, cur=0, max=0)\` tuple. The base service builds a LOCAL \`StatList\`, applies the archetype to it (via \`mercury::world_data::map_loaded\`), and serializes that for \`onStatUpdate\` — so **the CLIENT sees the right numbers from world entry while the SERVER's cell entity has zeroes.**

Consequences:
- \`RangedPhysicalDamage\` reads \`target.stats.get(FOCUS).cur\` as 0 → applied = 0, overflow = full focus_damage → spillover at max every shot.
- HEALTH cur/max both 0 → respawn's \`f.set_current(f.max)\` restores 0 HP, player spawns dead.

## Fix

In \`handle_init_player_state\`, after setting \`archetype_id\`, call \`entity.stats.apply_archetype(...)\` with the archetype-base values from \`mercury::archetype_stats\`. Same source-of-truth as the wire-side seed in \`map_loaded\`, so both sides stay aligned.

For the Soldier archetype this seeds:
- HEALTH: (min=0, cur=760, max=760)
- FOCUS:  (min=0, cur=1570, max=1570)

## Test plan

- [x] \`init_player_state_seeds_stats_from_archetype\` — pins HEALTH + FOCUS at Soldier archetype values; reverting the \`apply_archetype\` call trips the assertions.

## What this PR does NOT do

- **No regen tick.** Focus burns down during combat under the existing damage math; refill currently only happens on respawn or via Heal Focus. A separate PR is appropriate if the combat designer wants passive recovery.
- **No HEALTH regen either**, for the same reason.
- **Doesn't touch the wire path.** Client behavior is unchanged — it was already getting correct values from \`map_loaded\`.

### DIFF
--- a/crates/services/src/cell/service/base_messages/player_init.rs
+++ b/crates/services/src/cell/service/base_messages/player_init.rs
@@ -53,6 +53,56 @@ pub(in crate::cell::service) async fn handle_init_player_state(
         entity.player_id = Some(player_id);
         entity.archetype_id = Some(archetype_id);
 
+        // Seed core stats from the archetype's base values. Without
+        // this seed, the cell-side `CellEntity::stats` stays at the
+        // default `Stat { min: 0, cur: 0, max: 0 }` tuple it gets
+        // from `StatList::new()`, which means:
+        //   - FOCUS is at 0/0/0 → `RangedPhysicalDamage` reads
+        //     `target.stats.get(FOCUS).cur` as 0 and absorbs 0 — the
+        //     shield mechanism never engages and the player takes
+        //     full overflow damage every shot.
+        //   - HEALTH is at 0/0/0 → `set_current(max)` on respawn
+        //     restores 0, leaving the player with 0 HP at spawn.
+        //
+        // The base service's `world_data::map_loaded` builds a LOCAL
+        // `StatList`, applies the archetype to it, and serializes
+        // that for `onStatUpdate` — so the CLIENT sees the right
+        // numbers from world entry while the SERVER's cell entity
+        // has zeroes. Symptom on lomiada's 2026-06-04 18:09 session:
+        // every NPC shot landed `focus_overflow = focus_damage` (no
+        // absorb), spillover full strength, 86 HP per hit.
+        //
+        // The archetype lookup mirrors the map_loaded path so both
+        // sides use the same source-of-truth (the values hardcoded
+        // from `db/resources/Archetypes/Seed/archetypes.sql` in
+        // `mercury::world_data::stats::archetype_stats`).
+        {
+            let arch = crate::mercury::archetype_stats(archetype_id);
+            entity
+                .stats
+                .apply_archetype(&cimmeria_entity::stats::ArchetypeStatValues {
+                    coordination: arch.coordination,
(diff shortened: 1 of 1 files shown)


MyPalClara (commented):
## Gap Analysis — PR #499 (seed cell-entity stats from archetype on InitPlayerState)

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `player_init.rs` `apply_archetype` uses hardcoded `mercury::archetype_stats` | **The archetype stats are hardcoded in `mercury::world_data::stats::archetype_stats`.** If the DB seed values change, the hardcoded values become stale. **Consider loading from DB or adding a compile-time check that the hardcoded values match the seed SQL.** |
| **G2** | `player_init.rs` only seeds HEALTH and FOCUS | **The `apply_archetype` call seeds all stats (coordination, engagement, fortitude, etc.) but the test only validates HEALTH and FOCUS.** If `apply_archetype` has a bug in one of the other stats, the test wouldn't catch it. |
| **G3** | `player_init.rs` INFO log on every world entry | **The `archetype_stats_seeded` log fires on every `InitPlayerState`.** For high-traffic servers, this could be noisy. **Consider DEBUG level or rate-limiting.** |

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | No validation that archetype_id is valid | **The code calls `mercury::archetype_stats(archetype_id)` without validating that the archetype_id exists.** If an invalid archetype_id is passed, what happens? **Does `archetype_stats` panic or return defaults?** |
| **G5** | PR body says "No regen tick" | **The PR explicitly defers FOCUS/HEALTH regen.** This is documented, but there's no follow-up issue filed. **Consider creating an issue to track the regen implementation.** |

| **G6** | No test for non-Soldier archetypes | **The test only validates Soldier (archetype_id=1).** Other archetypes have different base stats — a test per archetype would catch archetype-specific bugs. |
| **G7** | No test for invalid archetype_id | **What happens if `archetype_id = 999`?** The behavior should be pinned. |

### Category 4: Documentation

| **G8** | `apply_archetype` call doesn't document the stat-set order | **The code builds an `ArchetypeStatValues` struct manually.** If the struct fields change order, the mapping could silently break. **Consider a comment noting the field-to-stat mapping.** |

### Summary

| Correctness issues | 3 | Low-Medium |
| Potential issues | 2 | Low |
| Missing test coverage | 2 | Low |
| Documentation | 1 | Low |

**Total: 8 gaps, all low-medium severity. Ready to ship.** 🔬🎯

Cadacious:
@coderabbitai review

PR #502, opened 2026-06-04
fix(bandolier): re-send onActiveSlotUpdate post-onClientReady (P90 swap)

Cadacious:
## Summary

Server-side mitigation for the client wire-emit suppression documented in [docs/reverse-engineering/findings/client-wire-emit-suppression.md](https://github.com/SandboxServers/Cimmeria/blob/main/docs/reverse-engineering/findings/client-wire-emit-suppression.md).

**Bug shape:** Lomiada 2026-06-04 18:09 — pressing F2 to swap to the P90 did nothing. SigNoz showed ZERO \`requestActiveSlotChange\` events in 3 h of session logs. The wire packet never got emitted.

## Ghidra-recovered cause

The client's Lua gate inside \`BandolierMod.ActivateBandolierSlotN\` guards on \`getActiveSlotForContainer(containerId) ~= N\`. If the cached active-slot value at the bag-list node's \`+0xc\` field already equals the pressed key's slot index, the keypress is a Lua no-op and \`requestActiveSlotChange\` is never called.

The cached value is written by the NetIn handler \`FUN_00da9ce0\` @ \`0x00da9ce0\` which walks \`SGWPlayer.bagList\` at \`+0x8c → +0x24\`. **Critical:** if the bag-list map is uninitialized at the time the packet is processed, the walk no-ops and the cached value stays at the default. The login burst at \`mercury::world_data::map_loaded:354\` DOES send \`onActiveSlotUpdate\`, but it's bundled with the bag-init packets in the same Mercury body — NetIn processing order can leave the bag-list map empty when the active-slot update is read.

## Fix

Re-send \`onActiveSlotUpdate\` from \`handle_init_player_state\` AFTER the client has confirmed \`onClientReady\`. This handler runs strictly after the entire \`mapLoaded\` bundle has been processed client-side, so the bag-list map is guaranteed initialized by then.

Wire format mirrors \`bandolier.rs:449-451\` and \`map_loaded.rs:354\`: bag_id=3, wire_slot=(server_slot + 1).

## Test plan

- [x] \`init_player_state_resends_active_slot_update_for_resync\` — walks the channel for an \`ON_ACTIVE_SLOT_UPDATE\` emission with the expected 8-byte payload and decodes the bag_id + wire_slot. Reverting the resend block trips this guard.

## What this PR does NOT do

- **Doesn't touch the login-burst send** in \`map_loaded.rs\`. That packet is still useful as the first-line value carrier; the resend is purely defensive.
- **Doesn't address Heal Focus** — that's a separate client-side gate (Lua arg validation + in-flight ability queue). Followup after a SigNoz playtest validates this one.

### DIFF
--- a/crates/services/src/cell/service/base_messages/player_init.rs
+++ b/crates/services/src/cell/service/base_messages/player_init.rs
@@ -228,6 +228,59 @@ pub(in crate::cell::service) async fn handle_init_player_state(
     // 3+ starter abilities couldn't see or click any of them.
     send_known_abilities_update(entity_id, tx, space_mgr).await;
 
+    // Re-send `onActiveSlotUpdate` for the bandolier — defensive resync
+    // against a client-side initialization race documented in
+    // [docs/reverse-engineering/findings/client-wire-emit-suppression.md].
+    //
+    // The login burst at `mercury::world_data::map_loaded` already sends
+    // this packet (see `map_loaded.rs:354`), but Ghidra analysis of the
+    // client's NetIn handler `FUN_00da9ce0` shows it walks
+    // `SGWPlayer.bagList` at `+0x8c → +0x24` and silently no-ops when
+    // the bag-list map is uninitialized. If the burst's
+    // `onActiveSlotUpdate` arrives before the bag-init packets in the
+    // same bundle are processed, the cached active-slot value at
+    // `slot+0xc` never gets written and stays at the default (0). The
+    // Lua gate inside `BandolierMod.ActivateBandolierSlotN` then reads
+    // `getActiveSlotForContainer(3) ~= N` as false for any keypress
+    // matching the stale-cached slot, and `requestActiveSlotChange` is
+    // never emitted — symptom: "F2 doesn't swap to the P90."
+    //
+    // `InitPlayerState` runs AFTER the client sends `onClientReady`,
+    // which is itself sent only AFTER the client has processed the
+    // entire `mapLoaded` bundle (per the handler's docstring above).
+    // So the bag-list map is guaranteed initialized by the time this
+    // resend lands — the cached value gets the correct write and the
+    // Lua gate stops misfiring.
+    //
+    // Wire format mirrors `bandolier.rs:449-451` and `map_loaded.rs:354`:
+    // bag_id (i32 LE) + (slot_id + 1) (i32 LE, 1-indexed wire) = 8 bytes.
+    {
+        const CONTAINER_BANDOLIER: i32 = 3;
+        let active_slot = space_mgr
+            .get_entity(entity_id)
+            .map(|e| e.active_bandolier_slot)
+            .unwrap_or(0);
+        let mut args = Vec::with_capacity(8);
+        args.extend_from_slice(&CONTAINER_BANDOLIER.to_le_bytes());
+        args.extend_from_slice(&(active_slot + 1).to_le_bytes());
+        crate::cell::abilities::send_entity_method(
+            entity_id,
+            crate::cell::client_methods::inventory::ON_ACTIVE_SLOT_UPDATE,
+            args,
+            tx,
+            space_mgr,
+        )
+        .await;
+        tracing::info!(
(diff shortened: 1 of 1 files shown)


MyPalClara (commented):
## Adversarial Review — PR #502

**This is a surgical fix for a real bug with Ghidra-confirmed root cause. The PR body is exceptional.**

---

### THE COMPLETE GAP LIST

#### **Category 1: Correctness Issues**

| Gap | Location | Concern |
|-----|----------|---------|
| **G1** | `player_init.rs` hardcoded `CONTAINER_BANDOLIER: i32 = 3` | **The constant is defined inline in the resend block.** There's likely a canonical constant elsewhere (e.g., `inventory::CONTAINER_BANDOLIER`). If the value ever changes, this inline definition would drift. **Use the canonical constant or add a static assertion.** |
| **G2** | `player_init.rs` `unwrap_or(0)` for missing entity | **If `get_entity(entity_id)` returns `None`, the code sends `active_slot = 0`.** This is probably fine (entity should always exist at this point), but **the failure mode is silent** — the client gets a valid-looking packet with slot 0 instead of an error. Consider `expect()` or a WARN log. |
| **G3** | `player_init.rs` resend happens unconditionally | **The resend fires for EVERY `InitPlayerState`, even if the login-burst packet was processed correctly.** This is documented as "defensive," but it means every login sends the same packet twice. **Is the bandwidth/processing cost acceptable?** (Probably yes — 8 bytes once per login — but worth noting.) |

#### **Category 2: Potential Issues**

| Gap | Location | Concern |
|-----|----------|---------|
| **G4** | Wire format duplication | **The wire format comment says "mirrors `bandolier.rs:449-451` and `map_loaded.rs:354`."** Three sites now encode the same 8-byte format. **Consider extracting a helper like `encode_active_slot_update(bag_id, slot) -> Vec<u8>` to avoid drift.** |
| **G5** | `tracing::info!` on every login | **The resend logs at INFO level for every player login.** In a busy server, this could be noisy. **Consider DEBUG level, or rate-limit to once per session.** |

#### **Category 3: Missing Test Coverage**

| **G6** | No test for `active_bandolier_slot = 0` | **The test uses `SERVER_SLOT = 1`, which is good, but doesn't test the boundary case of slot 0 (which would encode as wire_slot = 1).** |
| **G7** | No test for missing entity | **The `unwrap_or(0)` path (G2) isn't tested.** |

#### **Category 4: Documentation**

| **G8** | No cross-reference to the login-burst site | **The comment references `map_loaded.rs:354` but doesn't explain why we don't just FIX that site instead of adding a resend.** The answer is in the PR body (bundle processing order is non-deterministic), but a one-liner in the code comment would help future readers. |

---

### WHAT'S EXCELLENT

1. **The Ghidra analysis is gold.** `FUN_00da9ce0` @ `0x00da9ce0`, `SGWPlayer.bagList` at `+0x8c → +0x24`, `slot+0xc` — this is the level of detail that makes the fix trustworthy.

2. **The timing analysis is correct.** `onClientReady` is sent AFTER the client processes the `mapLoaded` bundle, so the bag-list map is guaranteed initialized. The resend is placed at exactly the right point in the lifecycle.

3. **The test pins the exact bug shape.** The test comment explains what reverting the fix would break, and the assertion checks both `bag_id` and the 1-indexed wire encoding.

4. **The PR body documents what this DOESN'T fix.** Heal Focus is explicitly called out as a separate issue. This prevents scope creep and sets expectations.

---

### SUMMARY

| Correctness issues | 3 | **Low** — G1-G3 are hardening |
| Potential issues | 2 | **Low** — G4-G5 are polish |
| Missing test coverage | 2 | **Low** — G6-G7 are edge cases |
| Documentation | 1 | **Low** — G8 is clarity |

**Total: 8 gaps, all low severity.**

---

**This PR is ready to ship.** The fix is surgical, the root cause is Ghidra-confirmed, and the test pins the exact bug shape. The 8 gaps are all low-severity hardening suggestions.

🔬🎯🔫

Cadacious:
@coderabbitai review

PR #511, opened 2026-06-10
fix(combat): persist BSF_AUTO_CYCLING across relog via sgw_player.state_field (#412)

### LINKED ISSUE #412: BSF_AUTO_CYCLING bit not persisted across relog
## Summary

`BSF_AUTO_CYCLING` (bit 1 of `state_field`) is not persisted across logins. On the user's first attack post-relog, the auto-cycle loop doesn't kick in until they manually fire once — same symptom as a fresh login where they've never pressed the auto-cycle button.

## Evidence

- `grep -rn "state_field.*save\|state_field.*persist\|sgw_player.*state_field" crates/services/src/base/world_entry/methods/player_load/` returns zero hits
- No SQL column for `state_field` in `db/sgw/` schema
- `state_field` is initialised in `CellEntity::new` to `0`, and on world entry never gets restored from any saved source

Other state-field bits are intentionally transient (`BSF_DEAD`, `BSF_MOVEMENT_LOCK`, `BSF_IN_COMBAT`), but `BSF_AUTO_CYCLING` is a player preference toggle — the original game persisted it.

## Fix shape

1. Add `state_field INT` column to `sgw_player` (migration in `db/scripts/`)
2. Read/write the bit in player save/load path
3. Only persist the user-preference bits — `BSF_AUTO_CYCLING` is the only one today, but the mask could grow
4. On load, restore the masked bits onto `entity.state_field` so the auto-cycle tick loop sees the flag set

## Acceptance

After implementation: enable auto-cycle, log out, log back in, first attack triggers the auto-cycle loop without needing a second click.

## Estimate

~3hr — schema migration + save/load wiring + 2 tests (one round-trip, one ensuring transient bits like BSF_DEAD aren't accidentally persisted).

## Related

- PR #410 — observability work that surfaced this. The user explicitly authorized "all cooldowns reset on world entry" which lines up with the relog-is-a-fresh-start philosophy — but explicit player-preference bits should still round-trip.

## Confidence

**High** on the diagnosis (zero code path persists `state_field`). **High** on the fix being mechanical.

BangRocket:
Closes #412.

> **Stacked on #510** (`fix/411-mission-offer-guard`) — the two changes share edits in `player_init.rs`. Merge #510 first; this PR's diff against its base branch is pure #412.

## What

`BSF_AUTO_CYCLING` (bit 1 of `state_field`) is a player preference toggle, but nothing persisted `state_field` — so after a relog the auto-cycle loop never armed until the player pressed the button again. This implements the fix shape from the issue:

1. **Schema**: `sgw_player.state_field INTEGER NOT NULL DEFAULT 0` (+ non-negative CHECK), with an idempotent migration in [db/scripts/add_player_state_field.sql](db/scripts/add_player_state_field.sql).
2. **Preference mask**: `PERSISTED_STATE_FIELD_MASK` in `cell/combat/state.rs` — `BSF_AUTO_CYCLING` alone today; growing the persisted set is a one-line change. Both the cell-side send and the base-side DB write mask with it, so transient combat bits (`BSF_Dead`, `BSF_InCombat`, `BSF_MovementLock`) can never reach durable state.
3. **Write path**: the explicit `setAutoCycle` toggle sends a new `CellToBaseMsg::StateFieldUpdate`; a new `cell_dispatch/state_field.rs` handler persists the masked value (same split as `SystemOptionsUpdate`). In-combat auto-clears (target death, manual fire of a different ability, `AF_DEACTIVATE_AUTO_CYCLE`) deliberately do **not** persist — the stored value tracks the player's deliberate button choice, not loop mechanics.
4. **Read path**: `InitPlayerState` carries `state_field` from the hydrate query; the handler masks again, ORs the bits onto `CellEntity::state_field` (raw `|=` per the single-source-flag rule in `auto_cycle.rs`), re-arms `abilities.auto_cycle`, and re-broadcasts `onStateFieldUpdate` so the client's gun-icon highlight survives the relog.

## Acceptance (issue #412)

Enable auto-cycle → relog → the restored `abilities.auto_cycle = true` means the first attack enters `arm_auto_cycle` and the server-driven re-fire loop starts — no second button press. The loop's ability stash stays empty until that first commit, by design (the wire payload carries no ability id).

## Tests

All revert-verified (each guard fails with its fix line neutered — 3 revert directions checked):

- **Toggle persist** (`world/tests.rs`): enable sends exactly one `StateFieldUpdate` with masked bits (a live `BSF_InCombat` riding the same value must be stripped); disable persists the cleared bit.
- **Login restore** (`player_init.rs`): persisted bit restores onto the entity + re-arms the loop flag + re-broadcasts `onStateFieldUpdate`; a corrupt row carrying `BSF_Dead | BSF_InCombat | BSF_MovementLock` restores **nothing** (a dead, movement-locked spawn is the failure shape the mask prevents); `state_field = 0` adds no packet to the login burst.
- **Live-DB** (`cell_dispatch/state_field.rs`, sentinel range `0x7000_1700`): fresh-row default 0 → enable → disable round-trip; unmasked mid-fight value stores only the auto-cycle bit; missing-row persist emits the documented `warn!` (LogCapture, per the negative-logging convention). These run in CI's live-DB job (no local Postgres on this dev machine).

## Docs

- [docs/architecture/state-field-bits.md](docs/architecture/state-field-bits.md) — new "Persistence across relogs" section (which bits persist, write/read contracts, why auto-clears don't persist).
- [docs/gameplay/combat-system.md](docs/gameplay/combat-system.md) — auto-cycle row notes the persistence.
- Schema comment on the new column + migration header explain the mask discipline in place.

### DIFF
--- a/.github/instructions/content-chains.instructions.md
+++ b/.github/instructions/content-chains.instructions.md
@@ -31,6 +31,17 @@ Every `op: "|"` (set) needs a matching `op: "~"` (clear) on the chain that compl
 
 For mission progression, also add a `player_loaded`-triggered chain that re-applies the bit for active steps. Interaction flags don't persist on the entity across server restart, so without restoration a relog mid-mission breaks interactivity. Worked example: chains 1045/1046 in `castle_cellblock_chains.sql` restore HackTheRings_Switch's bit based on which step is active.
 
+## Mission grants must gate on `not_active`
+
+Every chain whose actions include `accept_mission` must carry a
+`mission_status <id> eq not_active` condition (see chain 1001 for the
+canonical shape). Since #411 the server also refuses re-accepts
+authoritatively (`accept_mission`'s offer guard: already-active,
+failed-without-`can_repeat_on_fail`, or completed past `num_repeats`),
+so a missing gate no longer corrupts saved mission rows — but it still
+fires the chain's *other* actions (dialogs, highlights) spuriously, so
+the condition remains a review requirement.
+
 ## Inventory consumption
 
 `UseInventoryItem` fires `OnItemUse` as a pure event — the base no longer auto-consumes the stack. Chains that need to consume (consumable vials, mission objects) must include an explicit `remove_item` action. This is the correct pattern for `item_use`-triggered chains:
--- a/crates/services/src/base/world_entry/cell_dispatch/mod.rs
+++ b/crates/services/src/base/world_entry/cell_dispatch/mod.rs
@@ -37,6 +37,7 @@ use super::teleport::handle_teleport_player;
 mod aoi;
 mod bandolier;
 mod minigame;
+mod state_field;
 mod system_options;
 
 pub(crate) use aoi::flush_deferred_aoi;
@@ -783,6 +784,12 @@ pub(crate) async fn handle_cell_message(
                 db_pool,
             )
             .await;
+        }
+        CellToBaseMsg::StateFieldUpdate {
+            player_id,
+            state_field,
+        } => {
+            state_field::persist_state_field(player_id, state_field, db_pool).await;
         }
         CellToBaseMsg::RefreshAppearance {
             entity_id,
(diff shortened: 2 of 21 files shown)


BangRocket:
@coderabbitai review

BangRocket:
@coderabbitai full review

PR #518, opened 2026-06-14
feat(gm): SGWGmPlayer entity class — GMs enter world as 0x03 (#473, CAT-N-04)

BangRocket:
## In plain terms

Game Masters (admins/moderators) are supposed to log in as a special kind of player that the game client recognizes — one whose built-in GM console and debug commands actually light up. Until now, **everyone** logged in as the ordinary player type, so the client never enabled any of that, and the game's whole native GM toolbox sat dead. (This is the long-standing known issue **KI-14**.)

This change makes GM accounts enter the world as that special GM player type, and wires up the server so the GM commands the client sends are (a) only ever accepted from real GMs, and (b) safe even if a GM's client misbehaves. Regular players are completely unaffected — their login is byte-for-byte identical to before.

Three native GM commands are implemented as a working proof: **give yourself an item**, **teleport yourself to a coordinate**, and **kill a targeted NPC**. The rest of the GM command surface is recognized and locked down (only GMs can reach it) but not yet wired to actions — those are easy follow-ups.

> **Heads-up for review:** the server side is fully byte-verified by tests, but whether the real game client actually lights up its GM console when it sees the new player type can only be confirmed by logging in with an actual client. That manual check is the one thing tests here can't cover, and it's flagged in the docs.

---

## Technical summary

Resolves **KI-14 / #473 / CAT-N-04**. Depends on the cell-layer `access_level` plumbing + `gm_gate` from #512 (merged).

### The mechanism (why this is safe)

`SGWGmPlayer.def` declares `<Parent>SGWPlayer</Parent>` with an **empty** `<Implements>`. Under BigWorld's append-at-end inheritance, its own `<Exposed/>` CellMethods append at flattened indices **109–225** (117 methods) and the inherited SGWPlayer indices **0–108 do not renumber**. The wire `idbase` stays 61. So flipping a GM's `class_id` from `0x02` → `0x03`:

- changes nothing for the inherited method tables (the regular-player wire path is byte-identical), and
- makes the client expose the gm* cell surface at 109+.

The old `play_character.rs` comment claiming the flip "shifts ALL indices" was **wrong about the mechanism** — corrected here.

### What changed

[...]

Issue #475, opened 2026-06-01
[security-audit P0-1] Plumb `access_level` from base into cell-method dispatch — blocks every GM gate (CAT-N-03)


Cadacious:
Part of the server-authority audit — see umbrella #459 and category issue #473.

**Severity**: Critical — systemic. Blocks every GM-gating fix in CAT-N.

## Problem

The `access_level` (GM / admin / player) lives on `ConnectedClientState` in the base
layer. The cell-method dispatcher at `crates/services/src/cell/dispatch/router.rs` has
no access to it. Every future `gm*` handler added to the cell layer will be
unauthenticated-by-default. This means:

- The 40 `gm*` wire surfaces audited in CAT-N (#473) cannot be properly gated until
  this is plumbed — even if a handler does `if !access_level.is_gm() { return }`,
  the value isn't reachable from where it needs to live.
- The `Event_NetOut_SetHideGM` / `SetGodMode` / `SetHealth` / `SetSpeed` / `SetLevel` /
  `GiveItem` / `GiveNaqahdah` / `Kill` / `Spawn` / `Despawn` / `Goto*` / `Summon`
  surfaces (all in CAT-N) are all blocked on this.
- Related per-category gaps that depend on this: CAT-L-06 (`SendGMShout`),
  CAT-O-04 (`onWorldInstanceReset` on player entity).

## Evidence

- Ghidra: every `Event_NetOut_*` GM command exists on the client. Per the audit, the
  Rust dispatch fall-through at `crates/services/src/cell/dispatch/router.rs:101`
  logs a `warn!` and returns — no GM handler implemented today. **The gap is structural,
  not per-handler.**
- Wire surfaces inventoried in [`surface.md`](https://github.com/SandboxServers/Cimmeria/blob/worktree-server-authority-audit/docs/security-audit/2026-05-31-server-authority/surface.md#cat-n--gm--debug--cheat-critical--must-all-be-gm-gated).
- `access_level` source-of-truth lives at `crates/services/src/base/mod.rs:117`
  (`ConnectedClientState`).

## Suggested fix

Plumb `access_level` into the cell-method dispatch context — most likely as a field
on the `CellPlayerArgs` / equivalent passed into each `handle_*` handler. Consult
`network-security-auth` for the secure-by-default access_level lifecycle (default to
0 / player on session establish; only elevated when the auth ticket carries
`accesslevel > 0` from the `account` table).

[...]

PR #521, opened 2026-06-17
feat(gm): native gm* command suite — query / travel / missions / crafting / spawn

Cadacious:
## Scope (read this first)

This PR grew well beyond its original title. It lands **the developer-useful slice of the native `SGWGmPlayer` `gm*` command surface** — ~6 feature families plus a new base↔cell message round-trip — built on the native-console mechanism from #518 (no `.`-prefix chat commands; per the recorded design decision the client consumes `/`-commands and dispatches them as `SGWGmPlayer` cell methods). FanMMORPG (`doko972/FanMMORPG`) was the behavior reference.

It is **organized by commit, one feature family per commit**, so it can be reviewed (and reverted) family-by-family without splitting the branch:

| Commit | Family | Commands |
|---|---|---|
| `38297b70` | inspection | `gmShowPlayer` (131), `gmShowTargetLocation` (121), `gmShowRotation` (122) |
| `3d7f81a7` | query + travel + missions, **+ `tests.rs`→`tests/` split** | `listAbilities` (123), `gmShowFlag` (125), `gmGetMobAttribute` (127), `gmShowMobCount` (128), `gmDebugMobData` (180), `gmMissionList/Full/Details` (113/114/115), `gmGoto` (160), `gmSummon` (161), `gmMissionAssign` (109) |
| `012d617d` | **base-infra**: crafting + spawn | `gmGiveExpertise` (139), `gmGiveAppliedSciencePoints` (140), `gmSpawnByCmd` (185) |
| `1cf331b5`, `75610d24`, … | review fixes | CodeRabbit + Clara findings |

(Earlier groundwork — the inspection feedback channel, the REUSE set — landed in #518.)

## Highest-risk pieces (review these closely)

1. **`gmSpawnByCmd` — new cross-process handshake.** The cell can't build a `SpawnRecord` for an arbitrary `template_id` (no cell template cache), so this adds a round-trip mirroring `TrainAbility`→`AbilityGranted`: cell emits `CellToBaseMsg::GmSpawnNpc` → `base/gm_spawn.rs` queries `resources.entity_templates` → `BaseToCellMsg::GmSpawnNpcReady` → cell allocates an NPC id and spawns. New message **pair** + base DB loader + cell response handler. (`base/gm_spawn.rs`, `base_messages` arm.)
2. **Crafting grants — DB-writing.** `gmGiveExpertise`/`gmGiveAppliedSciencePoints` add two one-way `CellToBaseMsg` variants → `base/crafting/handlers.rs` (`load_crafting_state` → mutate → `save_crafting_state`), mirroring `handle_grant_cash`. Expertise pushes `onUpdateDiscipline`; applied-science persists (no outbound client method exists — refreshes on next crafting open/relog).

## Architecture / safety
- Every command is GM-gated by the whole-tail `gm_gate` (#475); handlers don't re-check access.
- Defense-in-depth input validation throughout (clamp, reject non-finite / non-positive / out-of-range / cross-space; `saturating_add` on expertise; `try_get` so a malformed `entity_templates` row drops the spawn rather than panicking the base task).
- GM spawns are one-shot (`spawn_id = -1`, `respawn_secs` forced `None`).

## Tests
- Cell-side guard tests across all families (happy + reject paths), the spawn round-trip covered both halves (base-side live-DB + cell-side handler), crafting handlers covered by live-DB guards, and a witness-visibility guard pinning that a summoned NPC broadcasts via the AoI tick.
- `tests.rs` was split into a `tests/` directory (no-behavior-change move, same commit as the query family).

## Status / dispatch table
Tally now **38 DONE · 0 REUSE · 35 ADAPT · 44 NEW** (`docs/protocol/cell-method-dispatch-table.md`); roadmap in `docs/architecture/gm-cell-method-adapt-plan.md`.

## Manual UAT still needed
Server side is test-covered; whether the real client's console actually fires each `gm*` index can only be confirmed by logging in as a GM. Gated to GM accounts only, so any client-side surprise is isolated.

## Summary by CodeRabbit

* **New Features**
  * Added GM mission tools (assign, list active/full, and mission details).
  * Added GM travel (goto and summon).
  * Added GM spawn-by-command to spawn NPCs from template id with X/Z offsets.
  * Added GM crafting grants (expertise and applied science points).
  * Expanded GM inspection overlays (target location/rotation, player view, abilities, flags, mob attributes/count, debug mob data).
* **Bug Fixes / Improvements**
  * Improved GM command validation and clearer GM feedback for success/failure across mission, travel, spawn, give/grants, stats, and world actions.
* **Documentation**
  * Updated GM command documentation and dispatch/roadmap status.
* **Tests**
  * Added/expanded async coverage for mission, travel, spawn-by-command, inspection, stats, world, and crafting-grant paths.


MyPalClara:
## Adversarial Review — Category 1: Critical / Functional

**The architecture is sound and the wiring is complete** (every constant → handler → message variant → base match arm checks out, base round-trips are tested both ways). But there are real functional gaps, not just doc nits.

| Gap | Location | Severity | Concern |
|-----|----------|----------|---------|
| **G1** | `travel.rs handle_summon` — NPC branch | **High** | When the summoned target is an **NPC**, the handler calls `update_entity_position(target_eid, caller_pos, …)` and then sends `TeleportPlayer` **only if `is_player`**. The comment asserts "an NPC via the spatial-grid update (witnesses pick it up on the next AoI tick)." That assumption needs proof. If NPC position deltas are normally emitted by the movement/AI tick (not by a raw grid mutation), a summoned **idle** NPC moves server-side but stays visually frozen on every client until something else dirties it. This is exactly the "works in the handler, does nothing in-game" class. **Does `update_entity_position` enqueue an AoI move-delta, or only mutate the grid cell?** If the latter, summon of a non-moving NPC is invisible until it next ticks. There is no test asserting a witness sees the NPC move — `tests/travel.rs` should pin it. |
| **G2** | `query.rs handle_show_rotation` — heading math | **Medium** | `heading = d.x.atan2(d.z).to_degrees()`. Every other system in this codebase that derives heading should agree on the convention. `atan2(x, z)` gives 0° at +Z (north) increasing toward +X. If the client/console or the existing `gmGotoLocation` path treats heading as `atan2(z, x)` (0° at +X) or uses a different winding, the reported number will be off by 90° or mirrored. **This is a read-only command so it can't corrupt state — but a GM trusting a wrong heading is a debugging trap.** Verify against the client's facing convention and cite it. |
| **G3** | `handle_summon` cross-space refusal is silent to the GM on the *NPC* path only via log | **Low-Med** | Player-target refusals and NPC refusals both `tracing::warn!` but send **no feedback** to the GM. `gmGoto`/`gmSummon` have no `send_gm_feedback` on the refusal/early-return paths, unlike the `query` family. A GM types `/summon 123`, nothing happens, no console reply — looks broken. Mirror the query handlers: emit a one-line feedback on every early return. |

MyPalClara:
## Category 2: Scope & PR Hygiene — *this is the real highest-severity finding*

| Gap | Severity | Concern |
|-----|----------|---------|
| **G4** | **High (process)** | **The PR description lies about its own scope.** The title and "What" section advertise *the inspection trio* (121/122/131 — `gmShowPlayer`/`gmShowTargetLocation`/`gmShowRotation`). The actual diff ships **far more**: crafting grants (`gmGiveExpertise` 139, `gmGiveAppliedSciencePoints` 140 + a whole `base/crafting/handlers.rs`), travel (`gmGoto` 160, `gmSummon` 161), missions (assign/list/listFull/details 109/113/114/115), spawn-by-cmd (185 + new `base/gm_spawn.rs` + a **new cross-process `GmSpawnNpc`/`GmSpawnNpcReady` handshake**), and more query handlers (123/125/127/128/180). That's ~6 distinct feature families and a new base↔cell message round-trip in one +3317/−873, 30-file PR. The CodeRabbit summary describes the real PR; the human-written body describes a different, smaller one. **A reviewer who trusts the body under-reviews the spawn handshake and the DB-writing crafting grants — the two highest-risk pieces.** Either split this (inspection / query / travel / missions / crafting / spawn are all independently shippable and independently revertible) or rewrite the body to match what actually changed. |
| **G5** | Medium | **The 823-line `tests.rs` → split `tests/` tree is bundled into the feature PR.** Deleting 823 lines of existing tests and re-adding them across 8 new files in the same commit as 6 new features makes it nearly impossible to verify *no existing test was silently dropped or weakened* in the move. The `−823 / +cumulative` split should be its own no-behavior-change refactor PR landed first, so this PR's diff shows only genuinely new test coverage. |
| **G6** | Low-Med | **`gmDebugMobData` ignores its first arg by design but the parse is fragile.** It reads `target` at offset 4 (`read_i32(args, 4)`) and discards `aSpaceID` at offset 0. If the def's arg order is ever `(target, spaceId)` instead of `(spaceId, target)`, this silently inspects the wrong entity with no error. There's a def-line citation for the *index* but none pinning the *arg order*. Add a wire-shape test from a real pcap or the def, like the trade PRs did for their traps. |

MyPalClara:
## Category 3: Correctness Edge Cases

| Gap | Location | Concern |
|-----|----------|---------|
| **G7** | `gm_spawn.rs load_spawn_record_for_template` | The query selects `template_name`, `class`, `body_set` and the record build uses `row.get(...)` (panicking accessor) on them, while the **test** deliberately filters to rows where those are `NOT NULL`. So the handler *assumes* those columns are non-null in production. If any `entity_templates` row has a NULL there (the seed audit you ran showed lots of half-wired content), `row.get` **panics the base task** instead of dropping the spawn gracefully like the not-found path does. Use `try_get` + a clean "template malformed" warn, matching the defensive posture everywhere else in this file. |
| **G8** | `gm_spawn.rs` GM spawn record | `spawn_id: -1` marks a non-DB spawn and `respawn_secs` is carried from the template. If a GM-spawned mob with a non-null `respawn_secs` is later killed, does the respawner try to respawn a `spawn_id = -1` instance? Either GM spawns should force `respawn_secs = None` (one-shot) or the respawner must explicitly skip negative spawn_ids. Worth a one-line assertion of intent. |
| **G9** | `query.rs subject_or_self` | `current_target_id` is filtered to the caller's space, good. But there's no staleness check — if the target died/despawned this tick, `get_entity(id)` returns None and you fall back to **self** silently. A GM inspecting a just-killed mob gets their own stats with no indication the target was lost. Minor, but a "(target gone, showing self)" note would prevent confusion. |
| **G10** | `handle_show_flag` | `1u32 << flag_id` with `flag_id` gated to `0..32` — correct, no UB. But `state_field` is printed as `0x{:08x}` (32-bit). If `state_field` is wider than u32 anywhere, the display truncates. Confirm `state_field` is u32. |

## Category 4: Tests & Docs

| Gap | Concern |
|-----|---------|
| **G11** | No witness-visibility test for `gmSummon` of an NPC (the G1 bug would pass all current tests, which only assert the message emission / grid value — not that a client sees the move). |
| **G12** | No test for `load_spawn_record_for_template` with a NULL `template_name`/`class`/`body_set` (the G7 panic path). The existing test *avoids* it by construction. |
| **G13** | `gmGiveAppliedSciencePoints` has "no client push" documented — but nothing tells the GM the grant landed. Since it's persist-only with no UI refresh until relog, the GM has no feedback the command worked. Consider a `send_gm_feedback` confirmation on the cell side. |

---

## Verdict

**This is genuinely good code** — the wiring is complete, the crafting clamp (`saturating_add` then `clamp` on the result, persist-before-push) is correct, the spawn round-trip is symmetric and tested both halves, validation rejects non-positive/non-finite/out-of-range inputs consistently. That's not rubber-stamp praise; I traced every handler to its dispatch arm and its message match.

**But it should not merge as-is**, and the reason is **G4 (scope) + G1 (NPC summon visibility)**, not the nits:

1. **G1** is a probable in-game no-op: summoned idle NPCs likely don't visually move. That's the "still doesn't work" failure mode. **Prove or fix the AoI delta before merge.**
2. **G4** — split the PR or fix the body. Six feature families + a new cross-process protocol hiding behind a "three inspection commands" title is how the spawn handshake and DB-writing crafting grants get under-reviewed.
3. **G7** — `row.get` on nullable template columns will panic the base task on the exact half-wired content your own audit found.

Fix G1, G7; split per G4. The rest are follow-ups.

Cadacious:
Thanks @MyPalClara — thorough pass. Resolutions (commit `c426d3d6` unless noted):

**Fixed (code):**
- **G7** — `load_spawn_record_for_template` now uses `try_get(...)?` for *every* column, so a NULL in a non-nullable `entity_templates` field drops the spawn with a warn instead of panicking the base task. Live-DB guard added (G12).
- **G8** — GM spawns are one-shot: `respawn_secs` forced `None` so a `spawn_id = -1` instance never reaches the respawner. (Removed the now-unused `normalize_respawn_secs` re-export.)

**Investigated → not a bug, now pinned (G1/G11):** `compute_aoi_changes` emits `EntityMoved` for *every* in-AoI entity on *every* ~100ms tick (position-snapshot, not dirty-flag), so a summoned idle NPC *does* broadcast. Added a witness-visibility test: summon NPC → AoI tick → `EntityMoved{witness, npc}` at the caller's position. It'll fail if AoI ever goes dirty-flag-based.

**Tests added (G11, G12 + the coverage gap):** the bigger problem was `missions.rs` at 18.9% (its happy paths were untested after the handler move). +15 behavior tests — full mission lifecycle, all `gmGetMobAttribute` arms, `gmDebugMobData`, `gmShowFlag` SET, summon-player/refusals, the 3 new `cell_dispatch` arms via the real router, and the G12 malformed-template drop.
- *Transparency on G12:* the live-DB guard ran green in a real pass, but my revert-verify was masked when the bundled Postgres dropped mid-session (`require_db_or_skip!` self-skips and still prints `ok` — a gotcha I've noted). The bug-shape proof rests on that executed pass + the mechanical certainty that `try_get::<String,_>` + `?` on a SQL NULL yields the `Err` arm.

**G4 (scope/body):** rewrote the PR title + description to honestly describe all ~6 families + the spawn handshake + the test split, with the two highest-risk pieces called out. Per maintainer decision we're keeping it one PR (it's one-family-per-commit, so it's reviewable/revertible per family) rather than splitting — which addresses your "either/or."

**Verified fine:**
- **G10** — `CellEntity.state_field` is `u32`; the `0x{:08x}` display can't truncate.

**Deferred as follow-ups (with reasons):**
- **G2** (heading convention) — read-only, can't corrupt state; the `atan2(x,z)` convention vs the client's facing should be confirmed before a GM trusts the number. Tracked, not blocking.
- **G3 / G13** (feedback on `gmGoto`/`gmSummon` refusals; ASP grant confirmation) — UX polish; the refusal paths warn server-side today. Worth a follow-up to mirror the query family's `send_gm_feedback`.
- **G5** (test split bundled) — acknowledged; it's a no-behavior-change move isolated to the query-family commit, kept here per the no-split decision.
- **G6** (gmDebugMobData arg-order wire test) — the index is def-cited; a pcap-pinned arg-order test is a good follow-up.
- **G9** (`subject_or_self` staleness note) — minor; a "(target gone, showing self)" hint would help, follow-up.

Net: the two you'd block on (G7 panic, G1 visibility) plus G8/G12 are resolved and tested; G4 addressed.

Cadacious:
@coderabbitai review

PR #526, opened 2026-06-19
feat(content): split mission 622 loot across both stasis-room corpses

Cadacious:
## What
"Arm Yourself" (mission 622) gave the pistol **and** the letter from Cpl Frost's body in one chain. This splits the loot across the **two real corpse NPCs** in the stasis room so each conveys something:

- **Cpl Frost** (`ArmYourself_FrostBody`, dialog 3995) → the **letter**.
- **Guard corpse** (`ArmYourself_GuardBody`, template 21) → the **pistol**, and it now owns the `2113 → 80622` ("equip the pistol") advance.

(The two distant near-door bodies the player sees are static map geometry, not server entities — so the two real corpses are the ones wired.)

## Wiring
- New **dialog 3996** ("Search the Guard's Corpse") + **dialog_set_map 5230**, bound to template 21 via a second `add_dialog_set` in **chain 1001** — the same mechanism (interaction_flags `0x40000000`) that makes Frost searchable from a 0-base interaction.
- **Chain 1003** (Frost): letter only — no pistol, no advance. Gated on `mission_status = active` (not step 2113) so the letter stays reachable regardless of loot order. Re-loot guard is the body's search-bit clear.
- **Chain 1005** (Guard): grants the pistol → backpack + advances `2113 → 80622`. Re-loot guard is the step-gate flip.
- **Chain 1004** (equip → door → complete) unchanged; **no new PAK step** (80622 already ships).
- Frost's dialog text updated (drops "a pistol and"); the guard's text conveys the pistol.

## Correctness notes (verified by the mission-systems advisor against the code)
- **`once` is dead code** — the `content_triggers.once` column is loaded but never enforced (`chain.rs` has no fired-set/dedup). So the only durable re-loot guards are a condition that flips false (chain 1005's step gate) or the client-side interaction-bit clear (chain 1003). This is why Frost's letter chain uses the bit-clear rather than `once`.
- **Loot order independence:** moving the advance onto the *pistol* chain and gating the *letter* chain on `mission_status = active` means neither grant becomes unreachable whether the player searches Frost or the Guard first.
- Honest caveat: the letter chain has no pure server-side re-grant guard (only the client bit-clear) — same exposure the original chain 1003 had. A hard guard would need a throwaway sub-step/counter; out of scope here (matches shipped behavior).

## Tests
`chain_replay_tests/mission_622.rs` (live-DB profile): chain 1003 = letter-only / no-advance; chain 1003 letter **still fires** after the guard advanced the step (order-independence); chain 1005 = pistol + advance; chain 1005 re-loot guard. Existing chain 1004 tests unchanged.

`cargo fmt` ✓ · `cargo clippy -p cimmeria-services --all-targets -D warnings` ✓ · `cargo check --all-targets` ✓. The chain-replay tests run in the live-DB CI profile (seeds fresh from `db/database.sql`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Restructured mission 622 loot distribution: Frost's corpse now grants only a letter, while the Guard's corpse grants a pistol and advances mission progression.
  * Updated mission dialogs to accurately reflect revised loot locations and item sources.

* **Tests**
  * Expanded test coverage to verify correct loot allocation and progression flow for mission 622.

### DIFF
--- a/crates/services/src/cell/content/chain_replay_tests/mission_622.rs
+++ b/crates/services/src/cell/content/chain_replay_tests/mission_622.rs
@@ -1,17 +1,21 @@
-//! Mission 622 — Arm Yourself! Pin the equip-from-inventory flow:
-//! chain 1003 grants the pistol to the backpack (container 1) and advances
-//! mission 622 to step 80622 ("Equip the pistol"); chain 1004 fires on
-//! `item_equipped(55)` and is the only path that plays kismet sequence
-//! 10000 (which opens the stasis-room door) and completes the mission.
+//! Mission 622 — Arm Yourself! Pin the **loot-split** + equip-from-inventory
+//! flow:
+//! - chain 1003 (Frost body, dialog 3995) grants ONLY the letter (3730) to the
+//!   mission inventory; it does not grant the pistol and does not advance.
+//! - chain 1005 (Guard corpse, dialog 3996) grants the pistol (55) to the
+//!   backpack (container 1) and owns the 2113 → 80622 advance.
+//! - chain 1004 fires on `item_equipped(55)` and is the only path that plays
+//!   kismet sequence 10000 (opens the stasis-room door) + completes the mission.
 //!
-//! The bug shape this guards: previously the pistol was added directly to
-//! the bandolier (container 3) and the mission auto-completed, which
-//! bypassed the bandolier ammo / appearance / fire-animation paths. Step
-//! 80622 is a Cimmeria-introduced step whose matching `<Steps>` row is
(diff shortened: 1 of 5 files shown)


Cadacious:
Superseding this PR: its loot-split commit (58d739aa) is already merged into `feat/dev-console-commands` (PR #524), which also carries the follow-on work (guard-corpse dialog PAK override + relog-safe Frost→Guard sequencing via the intermediate step 80623). Closing as redundant to avoid double-up on the mission 622 files; the loot split ships via #524.

PR #554, opened 2026-06-20
feat(discord): account names, new gameplay events, drop player IPs

Cadacious:
## What

Extends the Discord notification crate with the changes requested from live dev-session feedback.

- **Account names** — auth (`login`/`logout`/`disconnect`) and world (`entry`/`exit`) embeds now show the account name instead of a bare numeric id (`name (#id)`, falling back to `#id` then `?`). Threaded from the login ticket (`SessionRecord` → `PendingLogin.account_name` → `ConnectedClientState.account_name`), so **no extra DB lookup** at the Mercury login seam.
- **Clearer character name** — the character name is now on its own labelled `Character` field on world-entry / mission / item-used / loot / reward embeds, instead of a bare unlabelled description line.
- **Player IPs dropped everywhere** — no embed renders an IP anymore (login, disconnect, auth-failed, mercury-timeout, wire-format-error, high-latency). The `SocketAddr` is still carried internally for correlation; it just never reaches Discord.
- **New gameplay events** (44 `EventKind`s total, all on the `gameplay` channel):
  - `CharacterCreated` — default **on** (low volume / high signal)
  - `MinigameResult` — default **on** (entity-id only; the minigame server has no character-name handle)
  - `NpcDeath` — default **off** (high volume during combat)
  - `Dialog` (open + choice) — default **off** (high volume during questing)
- **Movement-validation noise suppressed** — `DiscordLayer` now drops `target="movement.validation"` warns (`speed_warning` / `validation_reject`). They're warn-only calibration telemetry for SigNoz that was flooding the errors channel during normal play (sub-tick deltas → huge/`inf` implied-speed ratios). Still flows to logs/SigNoz; just never posts to Discord.

## Emit sites wired

`base` (login/logout/disconnect/world-entry/exit, character_create), `cell` (NPC death in `death.rs`, dialog in `event_dispatch/dialog.rs`), and the minigame server.

## Notes

- Built on top of the split-discord module layout (#540), so it lands in the new `event/` `config/` `embed/` `sender/` homes.
- `MinigameResult` shows `entity:<id>` for the character — resolving the display name would need a cross-service round-trip not worth the coupling.

## Tests / gates

- `cimmeria-discord`: 62 unit tests pass (incl. new `movement_validation_target_filtered`, updated emit-count + variant-build).
- `cimmeria-services`: touched modules (login/world-entry/gate-travel/dialog) — 70 tests pass.
- `cargo fmt --check`, `cargo clippy -D warnings` (both crates), `cargo build` — all clean.
- No new Discord channel, so `compose.discord.yml` / release workflow untouched.


Cadacious:
Added a second commit: **graceful SIGTERM shutdown** (`crates/server/src/main.rs`).

The server only awaited `ctrl_c()` (SIGINT). Watchtower / `docker stop` / systemd send **SIGTERM**, which was never caught → the process was SIGKILLed after the grace period, so `stop_all()` never ran and the `ServerShutdown` Discord event never fired. The lingering hard-killed process could also race the new container's port binds — the likely reason the post-update `ServerStartup` looked missing.

Now selects over SIGTERM+SIGINT (Unix) / Ctrl-C (Windows) and threads the signal name into `emit_server_shutdown(reason, uptime)`, so the embed reports *why* (`SIGTERM (container stop / update)` vs `SIGINT (Ctrl-C)`). `ServerStartup` already fires on every boot; the clean old-container exit lets the replacement bind without racing.

Note: the Linux/SIGTERM branch (`#[cfg(unix)]`) is validated by the release-container CI build — this Windows dev host only type-checks the `not(unix)` branch.

PR #620, opened 2026-07-26
feat(client-telemetry): observe silently-dropped inbound entity methods

Cadacious:
Adds the client-side half of the round-trip oracle. The server can prove it *sent* an entity method; nothing until now could prove the real client **understood** it.

This is the observation primitive the automated Castle Cellblock loop needs, and it is the hook that would have caught the vendor-index bug in #609 on its first run.

## The failure mode being made visible

`Client_NetIn_EntityMethodDispatch` (`0x00c6f8f0`) searches the entity description's red-black handler map, keyed on `(componentKey, methodIndex)`. On a hit it fires the CME event. On a miss it calls `FUN_01590f30` at `0x00c6fa95` and returns — **the inbound method is discarded with no log, no error, and no wire response.**

That silence is the problem. A server-side bug that sends a well-formed method the client cannot route is invisible from both ends today.

## Why this hook point

Hooked at the not-found *callee* rather than the dispatcher entry or the drop site itself, for three reasons:

1. **The method index is computed inside the dispatcher** (`FUN_01590bb0`), so a hook at the entry cannot report *which* method was dropped — only that dispatch happened.
2. `FUN_01590f30` is a **function entry**, so the ordinary inline-hook path works instead of a mid-function splice.
3. It has **exactly one xref** — the drop site — so every call is a genuine drop, with no false positives. The address-pin test records that the single-xref property is load-bearing: if that ever stops being true, this hook starts reporting phantom drops.

Signature is `__thiscall(void* handler_map, uint method_index)`, so the index arrives as an argument rather than needing recovery from the stream.

## Emission policy

`client.dispatch.method_dropped` at **`warn`**, **unsampled**.

A drop is a correctness failure, not telemetry chatter. On a healthy session this is silent, so any volume here is itself the finding. If a future client build turns out to drop routinely, the right response is to add a sampler — not to lower the level.

**Known-expected drops today:** BlackMarket `onBM*` (methods 90–95) are parsed and flagged Exposed but were never bound into the handler map, so they always land here until the runtime patch in `black-market-client-window-patch.md` is applied.

## Thread safety

This runs on a **Mercury network thread**, not the main game thread — the same as `ADDR_HANDLE_MESSAGE`, which sets the precedent. `try_emit` is the lock-free bounded path and drops on a full queue rather than blocking the network thread. The detour reads no game state and takes no locks. That constraint is stated in the doc comment so a future edit does not casually add a game-state read.

## Verification, and a caveat that matters

Verified on the **real target** — `i686-pc-windows-msvc`, not the x86_64 host. This is worth calling out: an initial `cargo check` on the host passed while proving nothing, because the `target_arch = "x86"` cfg silently excluded every line of the change. `cargo check`, `clippy -D warnings`, and 50 tests pass on the 32-bit target.

**Not runtime-verified.** Confirming this hook actually fires needs an injected session against a live client, which is Stage 0/1 territory (#617). Adding a new address to the injected DLL should stay a human-supervised activity — a bad hook crashes the client, and the Black Market work produced four crashes from four distinct causes. Adding new *scenarios* on an existing hook does not carry that risk; adding a new *address* does.

## Unrelated observation

While running the suite repeatedly on this branch, I saw 2 intermittent failures across ~15 runs (49/50), followed by 8 consecutive clean runs. I could not capture the failing test name. It is **not** attributable to this change — which touches only a constant, a detour, and a deterministic address assertion — and the likely candidate is the uploader test that binds a local HTTP server, a classic port-race shape. Flagging it as an observation rather than a diagnosis; worth a look if CI shows the same pattern.

https://claude.ai/code/session_01GhH7BcbHyhFZHdWxKKc6gu

## Summary by CodeRabbit

* **Bug Fixes**
  * Added telemetry for entity methods that are silently dropped, including the dropped method index.
  * Improved handling when the underlying method hook is unavailable.


Cadacious:
@coderabbitai review

Cadacious:
Pushed 9a153a5e addressing all four review findings:

- **Copilot — ABI mismatch**: The v5-campaign RE checkpoint (`worker-4b1.checkpoint.json`) identifies `0x01590f30` as `EntityDescription_GetExposedClientMethodByIndex`, returning a `MethodDescription*` in EAX. The detour/trampoline now type the return as an opaque `*mut c_void` (same width, correct meaning), the doc states the return explicitly, and the no-trampoline fallback returns null — the callee's own "not found" answer.
- **Copilot — "lock-free"/"no locks" wording**: Reworded both spots. The documented constraint is now "non-blocking end to end": the builder heap-allocates but takes no game-state locks, and `try_emit` is a bounded-channel `try_send` that drops on full.
- **CodeRabbit — file-size cap**: `inline_hooks.rs` split into an `inline_hooks/` directory module (`mercury_dispatch` / `engine_frame` / `engine_loading` / `state_flags` / `anim_notify` / `console_command`, orchestration + aggregated anchor-pin test in `mod.rs`). Largest file is now ~420 lines. The split also repaired a misplaced insertion that had orphaned `state_field_update_detour`'s doc comment. `read_utf16_bounded` is re-exported at its pre-split path for `iat_hooks.rs`.
- **CodeRabbit — event documentation**: Added a "Dispatch drop oracle" section to `client-instrumentation-hookpoints.md` cataloguing `client.dispatch.method_dropped` (warn, unsampled, `method_index` field, known-expected BM 90–95 drops), and corrected the stale 11-hook count in `hooks.rs`.

Validated locally on the i686-pc-windows-msvc CI target: clippy `-D warnings` clean, `cargo nextest` 52/52, fmt clean.

